Issue #2, opened 2015-07-22
Convert to Xcodeproj + Unit Tests


AliSoftware:
Now that the API is stable and it's less experimenting in a playground and more a real lib, it's time to create a real Xcodeproj and add Unit Tests
(even if we keep a playground around so that people can play around and test the code, because I like the didactic aspect of playgrounds)

AliSoftware:
Started in 2d9a7c5a528172a3fa43da15b50de396fa0a7cef and cf190992c3922ad37497422efc80742079f65ac7 (but need more Unit Tests)

AliSoftware:
Added some more unit tests. Considered closed now, even if more tests can still be added in the future (especially when SwiftGen-L10n will be more robust)

PR #8, opened 2015-07-24
Segues support

### LINKED ISSUE #3: [SwiftGen-Storyboard] Add support for Segues
The `SwiftGenStoryboardEnumBuilder` only build enums for Storyboards and Scenes identifiers, but not yet for Segues.

_Note: Before starting implementing it, we need to brainstorm how we want to represent that in the generated code (to structure it properly and avoid mixing in the same `enum` both some `case` for the Scene identifiers and the `cases` for the Segue identifiers)_

esttorhe:
# Fix #3 

## Done:
- [x] Detects the segues from the `Storyboard` file
- [x] Extracting scenes and segues into different arrays to support 1 `Scenes` extension for `UIStoryboard` and 1 `Segues` extension
- [x] Only generating segue enums for `Storyboards` that have at least 1 (others are ignored)
- [x] `Scene` enum and `Segue` enum are now used instead of `Name`

## Generated code:
```swift

import Foundation
import UIKit

protocol StoryboardScene : RawRepresentable {
    static var storyboardName : String { get }
    static func storyboard() -> UIStoryboard
    static func initialViewController() -> UIViewController
    func viewController() -> UIViewController
    static func viewController(identifier: Self) -> UIViewController
}

extension StoryboardScene where Self.RawValue == String {
    static func storyboard() -> UIStoryboard {
        return UIStoryboard(name: self.storyboardName, bundle: nil)
    }

    static func initialViewController() -> UIViewController {
        return storyboard().instantiateInitialViewController()!
    }

    func viewController() -> UIViewController {
        return Self.storyboard().instantiateViewControllerWithIdentifier(self.rawValue)
    }
    static func viewController(identifier: Self) -> UIViewController {
        return identifier.viewController()
    }
}

extension UIStoryboard {
    enum Scene {
        enum Wizzard : String, StoryboardScene {
            static let storyboardName = "Wizzard"

            case CreateAccount = "CreateAccount"
            static var createAccountViewController : CreateAccViewController {
                return Wizzard.CreateAccount.viewController() as! CreateAccViewController
            }

            case AcceptCGU = "Accept-CGU"
            static var acceptCGUViewController : UIViewController {
                return Wizzard.AcceptCGU.viewController()
            }

            case ValidatePassword = "Validate_Password"
            static var validatePasswordViewController : UIViewController {
                return Wizzard.ValidatePassword.viewController()
            }

            case Preferences = "Preferences"
            static var preferencesViewController : UIViewController {
                return Wizzard.Preferences.viewController()
            }
        }
        enum Message : String, StoryboardScene {
            static let storyboardName = "Message"

            case Composer = "Composer"
            static var composerViewController : UIViewController {
                return Message.Composer.viewController()
            }

            case URLChooser = "URLChooser"
            static var urlChooserViewController : XXPickerViewController {
                return Message.URLChooser.viewController() as! XXPickerViewController
            }
        }
    }

    enum Segue {
        enum Message : String {
            /// Custom
            case CustomSegue = "Custom"

            /// NonCustom
            case NonCustomSegue = "NonCustom"
        }
    }
}
```

---
Giphy


SwiftGen.playground/Pages/Storyboard-Demo.xcplaygroundpage/Contents.swift
```diff
@@ -86,4 +101,12 @@ validateVC.title
    not known by the storyboard. But it should work correctly in a real project. */
 // let cgu = UIStoryboard.Name.Wizzard.createAccountViewController
 
+let segue = UIStoryboard.Segue.Message(rawValue: "CustomSegue")!
```
AliSoftware: Given that your string associated with the `.CustomSegue` case above is `"Custom"` not `"CustomSegue"` I guess this doesn't actually work? Probably forgot to update your example then? (reading that from GitHub so not actually tested here)

esttorhe: You are :100:% correct

AliSoftware: Could be a good idea to add an example with real-world code BTW, like how you would use that enum to typically implement `prepareForSegue` (even in case you need to comment that code if it doesn't work in a Playground context, that'd still be useful for the reader that opens the playground to understand typical usages)

SwiftGen.playground/Sources/SwiftGenStoryboardEnumBuilder.swift
```diff
@@ -14,30 +15,39 @@ public final class SwiftGenStoryboardEnumBuilder {
...
+                readyForFirstObject = true
+              case _ where readyForFirstObject:
+                if let storyboardID = attributeDict["storyboardIdentifier"] {
+                  let customClass = attributeDict["customClass"]
+                  scenes.append(Scene(storyboardID, customClass))
                 }
+                readyForFirstObject = false
+              case "connections":
+                readyForConnections = true
+              case _ where readyForConnections:
```
AliSoftware: why not `case "segue" where readyForConnections`, just to be sure to only grab `segue` nodes here?

Note that I used `_` for the `objects` subsides because they can actually be of various type, like `viewController` or `tableViewController` etc, but I don't think that could actually be the case for `segue`… And I even wonder if that `<connection>` node doesn't also contain `IBOutlets` if any, may be worth adding some in the demo storyboard to check that out.

SwiftGen.playground/Sources/SwiftGenStoryboardEnumBuilder.swift
```diff
@@ -102,14 +117,38 @@ public final class SwiftGenStoryboardEnumBuilder {
         if enumName != nil {
             text += "\(t)}\n"
         }
-        text += "}\n\n"
+      
+        /// Segues
+        if (storyboardsSegues.count > 0) {
+          text += "\n"
+          
+          let enumName = "Segue"
```
AliSoftware: That should probably be configurable in the builder as a parameter for the `build()` function (replace that single `enumName` parameter with `sceneEnumName`+`segueEnumName`)?

SwiftGen.playground/Pages/Storyboard-Demo.xcplaygroundpage/Contents.swift
esttorhe: I added this to the end of that `Storyboard`

```swift
/*******************************************************************************
This is a «real world» example of how you can benefit from the generated enum;
you can easily switch or directly compare the passed int `segue` with the corresponding
segues for a specific storyboard.
*******************************************************************************/
//override func prepareForSegue(_ segue: UIStoryboardSegue, sender sender: AnyObject?) {
//  switch UIStoryboard.Segue.Message(rawValue: segue.identifier)! {
//  case .CustomSegue:
//    // Prepare for your custom segue transition
//  case .NonCustomSegue:
//    // Pass in information to the destination View Controller
//  }
//}
```

What do you think?

AliSoftware: Except for the typo ("the passed int `segue`"), seems :ok_hand:

SwiftGen.playground/Sources/SwiftGenStoryboardEnumBuilder.swift
esttorhe: We would have to add then a custom view controller class to connect the `IBOutlet`s there which in turn would «bloat» the `Storyboard`.

But is worth checking that the `<connections>` won't ever contain stuff other than `segue`s and perhaps clean up that piece of the code.

AliSoftware: Yeah, good point. Probably don't do it in the storyboard actually shipped with the playground & code, but still worth checking in a storyboard created aside :wink: 

(That's why we need #2 to add tests :smile: )

SwiftGen.playground/Sources/SwiftGenStoryboardEnumBuilder.swift
esttorhe: Just checked in a commit to support this :ok_hand:

SwiftGen.playground/Sources/SwiftGenStoryboardEnumBuilder.swift
```diff
@@ -4,7 +4,8 @@ import Foundation
 
 public final class SwiftGenStoryboardEnumBuilder {
     private typealias Scene = (storyboardID: String, customClass: String?)
```
AliSoftware: We should probably rename that tuple to something else for consistency, to give it a name that can both fit for Scenes and Segues (and then use `identifier` for the first entry name, instead of `storyboardID`)

SwiftGen.playground/Sources/SwiftGenStoryboardEnumBuilder.swift
esttorhe: Just checked on another project of mine and here's a `Storyboard` scene with outlets
```xml
<connections>
    <outlet property="hairlineView" destination="VRg-OY-4IC" id="PuB-XN-okm"/>
    <outlet property="portTextField" destination="6b3-TB-J53" id="Ks7-gi-ZD7"/>
    <outlet property="serverTextField" destination="Lfy-mJ-Msa" id="apH-fb-2cT"/>
    <segue destination="Z5I-uz-FR0" kind="show" identifier="ShowClientsSegue" id="cml-0d-ha0"/>
</connections>
```

SwiftGen.playground/Sources/SwiftGenStoryboardEnumBuilder.swift
AliSoftware: :tada:

SwiftGen.playground/Sources/SwiftGenStoryboardEnumBuilder.swift
esttorhe: Argh… seems like I discarded that change…

I had it like `StoryboardElement` or `UIElement` can't remember and the first parameter was actually `identifier` :/

Which name would you rather use? or none of the above?

AliSoftware: `StoryboardElement` seems ok. `Entry` could do it too I guess.
I see the idea behind `UIElement` but think it's took generic (and somehow makes me think about `AccessibilityElement`, not sure why)

Or… we could also add another `typealias`, even if it's equal to the same tuple type, just to clarify things (and to be more flexible if some day we need to add another parameter to that tuple for scenes but not segues or the other way around)?

esttorhe: I explored adding an extra `typealias` but thought it would be weird.

Honestly you suggesting it kinds of validates my first thought. 

Let me add it and commit that.

esttorhe: This d773bd856afe56a29e8fbd00ff2b84a90ce407ca should take care of this

SwiftGen.playground/Sources/SwiftGenStoryboardEnumBuilder.swift
```diff
@@ -70,17 +83,19 @@ public final class SwiftGenStoryboardEnumBuilder {
         }
     }
     
-    public func build(enumName enumName: String? = "Name", indentation indent : SwiftGenIndentation = .Spaces(4)) -> String {
+    public func build(scenesEnumName enumName: String? = "Scene", seguesEnumName: String? = "Segue", indentation indent : SwiftGenIndentation = .Spaces(4)) -> String {
```
AliSoftware: Now that we have two different `enums`, it doesn't even make sense anymore to let the option to the user to _not_ provide an `enumName` (which so far was possible and generated all the `enums` for the `Scenes` at the root of the `UIStoryboard` extension). So we better migrate those two parameters to non-optional `String`s

SwiftGen.playground/Sources/SwiftGenStoryboardEnumBuilder.swift
AliSoftware: :dancer:

SwiftGen.playground/Sources/SwiftGenStoryboardEnumBuilder.swift
esttorhe: Changed here 3983edc6d87ae403b0ce866f6e5ecaa4990bb651

I left default values though just in case

SwiftGen.playground/Sources/SwiftGenStoryboardEnumBuilder.swift
AliSoftware: Ah I knew I've seen those outlets here some times back! Better add that `"segue"` in the `case` then!

SwiftGen.playground/Sources/SwiftGenStoryboardEnumBuilder.swift
AliSoftware: Yup, those default values are still good to keep :smiley:

AliSoftware:
Seems good from here now, will merge tomorrow after some good sleep.
007 can deserve his :cocktail: :wink:

SwiftGen.playground/Sources/SwiftGenStoryboardEnumBuilder.swift
```diff
@@ -70,17 +84,17 @@ public final class SwiftGenStoryboardEnumBuilder {
...
 
         text += "extension UIStoryboard {\n"
-        let s = enumName == nil ? "" : t
-        if let enumName = enumName {
-            text += "\(t)enum \(enumName.asSwiftIdentifier()) {\n"
-        }
-        for (name, scenes) in storyboards {
+      
+        /// Scenes
+        let s = t
```
AliSoftware: Now that those params are not optional anymore and those intermediate enums can't be omitted, this `s` can go away and be replaced by `t` everywhere

SwiftGen.playground/Sources/SwiftGenStoryboardEnumBuilder.swift
```diff
@@ -99,17 +113,38 @@ public final class SwiftGenStoryboardEnumBuilder {
...
+          for (name, segues) in storyboardsSegues
+            where segues.count > 0 {
+              
+            let enumName = name.asSwiftIdentifier(forbiddenChars: "_")
+            text += "\(s)\(t)enum \(enumName) : String {"
+            
+            for segue in segues {
+              let caseName = segue.segueID.asSwiftIdentifier(forbiddenChars: "_")
+              text += "\n\(s)\(t)\(t)/// \(caseName)\n"
+              text += "\(s)\(t)\(t)case \(caseName)Segue = \"\(caseName)\"\n"
```
AliSoftware: Wondering if that `Segue` suffix in the caseName is really necessary (given that this enum is already namespaced anyway)

esttorhe: it is fact __*not*__ necessary.

It actually reads wrong. I'm going to take it out as well as the `= "\(caseName)\"` assignment since 2.0 doesn't require that for `String` cases whose `rawValue` is not provided and is inferred to be the same as the `case` which IMHO reads better.

AliSoftware: **Don't** take out the assignment but rather fix it instead: missed that earlier but in fact it should `= "\(segue.segueID)"` instead (in case the `asSwiftIdentifier(…)` method did remove some special chars invalid for an identifier but you still need to associate the real identifier string to that enum)

esttorhe: Bitten for being «trigger-happy» :sweat: 

Give me a sec

esttorhe: :sweat_smile:  5c5e2b01c3d4752742833bafd7e521e40022b9bd

PR #12, opened 2015-08-21
Added import Foundation at top of L10N code.

Nick11:
Added "import Foundation" at top of L10N code.
Makes it more usable as build-script, since you don't need to add it manually (or y some ugly pre-pending script) each time.


AliSoftware:
:+1: well spotted thx!

PR #20, opened 2015-09-29
Avoid duplications in segue cases

ldiqual:
@AliSoftware 

Partial fix for https://github.com/AliSoftware/SwiftGen/issues/7

Since multiple segues can have the same identifier, the previous version of the code would create duplicate cases. A simple solution is to keep a reference on previously added case names so we don't add them more then once.

A cleaner solution would have been to use a `[caseName: segueID]` dictionary and then iterate over it to generate the code. However it would break the existing tests since iterating over a dictionary is not stable (doesn't necessarily keep the order of which elements were added).

Modified `Message.storyboard` with a duplicate segue ID, all tests pass.


AliSoftware:
This could be a workaround, but what about the case when two segue identifiers generate the same `caseName` BUT have different `segueID`?

Like `Account-Edit` vs. `Account_Edit` vs. `Account.Edit` which all generate the same `caseName` because `asSwiftIdentifier` generate the same output for all of those inputs…

AliSoftware:
One idea we could probably use instead:

*  keep track both of the `existingSegueID` AND the `existingCaseNames`, but maybe not in a `Dictionary` as you said, but in two separate `Arrays`  instead
* don't generate a case at all if a `segueID` already exists in the `existingSegueIDs` array
* otherwise (the `segueID` of the exact same name was never encountered before), compute the `caseName`, and if the generated `caseName` already exists in the `existingCaseNames`, generate an incremental variation of it

This way if we have the following `segueIDs`: `Account-Edit`, `Account_Edit`, `Account-Edit` again, and `Account.Edit`
Then we could generate stuff like this:

```
case Account_Edit = "Account-Edit"
case Account_Edit2 = "Account_Edit"
case Account_Edit3 = "Account.Edit"
```

ldiqual:
@AliSoftware Great idea. I just updated my PR with the required changes. I wonder if we could find something more elegant but at least this does the job. Also updated tests to make sure everything works alright.

AliSoftware:
Great work, thanks! :+1:

There are still problems I want to address before merging this, in particular:

* If you have `segueIDs` named `Account_Edit`, `Account-Edit` and `Account_Edit2`, then the first 2 `case` generated will be `Account_Edit` and `Account_Edit2`, but the third won't have an entry in your `existingCases` dictionary (because the `originalCaseName` will be `Account_Edit2`, but the only key it has will be `Account_Edit`) so it will generate another `case Account_Edit2`!
* I really want all that logic extracted. Because it will be useful for Storyboard, but also for a lot of other stuff. So I was thinking about a dedicated class `CaseGenerator` or something to handle all that logic separately from the rest of the `EnumBuilders`.


That dedicated class `CaseGenerator` could work something like this:
* You create one at the very beginning
* Each time you need a case, you give it the input value (in `StoryboardEnumBuilder` that would be the `segueID`, in `L10nEnumBuilder` that would be the localization key, in `AssetsEnumBuilder` this would be the image name, etc)
* The `CaseGenerator` will be in charge of calling `asSwiftIdentifier()` on it to transform it into a valid identifier
* Then it will check for duplicates, and increment until it finds a non-existing identifier
* It will add this identifier to its internal `Set` of already generated identifier, and return it to the caller

Basically all the logic you have implemented it is mainly good, but could greatly benefit being extracted into that dedicated class. Once it's extracted, we can convert the other `EnumBuilders` so they all use this `CaseGenerator`, then we can continue to improve it separately so it benefits all the builders. And have dedicated tests for that `CaseGenerator` that cover even edge cases like the one I'm talking about at the beginning of that comment :wink:

Sorry to make you change the code again, but I believe that would really be beneficial to improve the architecture and security of that enum generation :wink:

AliSoftware:
This has finally be fixed by #43

PR #28, opened 2015-10-05
Fix incorrect color parsing

Igor-Palaguta:
Fix parsing value, that starts with 0x and alpha is absent

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ---
 
+## 0.4.4
+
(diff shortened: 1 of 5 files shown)


AliSoftware:
LGTM :+1: 

Could you please:

* add such a case in the test fixtures (having a `0x123456` color in the `colors.txt` for example), to ensure this use case is covered and there won't be any regression in the future about this
* add an entry in the CHANGELOG to credit yourself

Thx!

Igor-Palaguta:
Please check commit

AliSoftware:
Perfect :ok_hand: thx

Issue #19, opened 2015-09-25
Support UITableViewCell identifiers


AliSoftware:
Generate something like this:

```swift
extension UITableViewCell {
  enum Identifiers : String {
    case Home
    case Menu
    case Product
    ...

    var nib: UINib { return UINib(nibName: self.rawValue, bundle: nil) }
  }
}

extension UITableView {
  func registerCell(identifier: UITableViewCell.Identifier) {
    self.registerNib(identifier.nib, forCellReuseIdentifier: identifier.rawValue)
  }
  func dequeueReusableCellWithIdentifier(_ identifier: UITableViewCell.Identifier,
      forIndexPath indexPath: NSIndexPath) {
    self.dequeueReusableCellWithIdentifier(identifier.rawValue,
                          forIndexPath: indexPath)
  }
}
```

Usage:

```swift
tableView.registerCell(.Home)
tableView.registerCell(.Menu)
…
var cell = tableView.dequeueReusableCellWithIdentifier(.Home, forIndexPath: indexPath)
```

AliSoftware:
Probably better wait for #24 before starting to work on this issue

yoiang:
Hey @AliSoftware ! In developing a tool with similar features to SwiftGen I built a library that makes extracting (eventually) everything, including UITableViewCell and UICollectionViewCell identifiers in Storyboards much easier. It may be overkill for your use but the library is [StoryboardKit](https://github.com/Adorkable/StoryboardKit).

To see UITableView Identifier extraction check seguecode's usage here: [ViewControllerClassInfo+seguecode.swift Line 35-66](https://github.com/Adorkable/seguecode/blob/master/seguecodeKit/StoryboardKit/ViewControllerClassInfo%2Bseguecode.swift#L35-L66)

AliSoftware:
Hey  @ioyang

Thanks for the input, will take a look later in the week!

phatblat:
+1 on this. I suggest adding `UICollectionViewCell` to the issue title (they are easy to forget about :wink:).

jmacmullin:
+1

AliSoftware:
Feel free to make a PR to make this be implemented and released faster, as I'm pretty busy those days and won't have much time implementing it that soon.

jmacmullin:
No worries :-)
https://github.com/AliSoftware/SwiftGen/pull/66

krodak:
My comment is not related to `SwiftGen`, but as it user I'll comment anyway.
[...]

PR #29, opened 2015-10-10
Support UITableViewCell identifiers #19

samirGuerdah:
This fix the #19 issue.
Let me know whether this work is the one you were expecting. Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # SwiftGen
 
+[![CI Status](http://img.shields.io/travis/AliSoftware/SwiftGen.svg?style=flat)](https://travis-ci.org/AliSoftware/SwiftGen)
+
 This is a suite of tools written in Swift to auto-generate Swift code for various assets of your project:
 
 * [`enums` for your Assets Catalogs](#assets-catalogs)


AliSoftware:
Thx for the PR!

The idea with SwiftGen though is to parse your code to auto-generate the constants, not to let the user give the constants themselves as command line arguments (otherwise it loose its benefits as the user could simply add a `case` in its code itself manually)

What I had in mind with `swiftgen-tableview` would be that it takes a path as argument, scan for every `*.xib` and `*.storyboard` files, parse those XML files, detect automatically the `UITableViewCell` elements in those XMLs and their `reuseIdentifier` attribute and generate the `enum` from those. Quite like `swiftgen-storyboard` does actually with the Scenes and Segues (maybe the code in there could help inspire the code for tableviewcells)

samirGuerdah:
Ok it's clear, thanks.
Will submit a new one this week.

AliSoftware:
Before you do too much work, maybe wait until I merge my refactoring from #30 

You can still work on your `TableViewEnumBuilder` class (that would probably be modeled after the `StoryboardEnumBuilder` class) and write the unit tests in the meantime, but maybe keep them in dedicated files on your side, and wait until I merge #30 to integrate them to my project — otherwise the `git merge` could be quite bumpy :wink:

PR #34, opened 2015-10-15
Excludechars

Igor-Palaguta:
Added excludechars argument to image, string, storyboard command

### DIFF
--- a/GenumKit/EnumBuilders/ImageEnumBuilder.swift
+++ b/GenumKit/EnumBuilders/ImageEnumBuilder.swift
@@ -32,7 +32,7 @@ public final class ImageEnumBuilder {
         }
     }
     
-    public func build(enumName enumName : String = "Asset", indentation indent : Indentation = .Spaces(4)) -> String {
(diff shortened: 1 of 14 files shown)


AliSoftware:
LGTM from the GitHub diff, will have to test in Xcode once I'm home.
Any idea why the tests fails on Travis? Did they pass locally in your Xcode?

Igor-Palaguta:
On my machine all tests passed, do you have expected and result values for failed tests?

AliSoftware:
Well I only see the same as what you can see in Travis CI, sot sure why it fails yet.

AliSoftware:
> On my machine all tests passed

I can't see how that was possible: on my machine they don't pass (in fact they crash), because it seem you forgot to add the new expected files (`*.swift.out`) to the Xodeproj (so it fails to load the fixture to compare the result with the expected fixture).

Are you sure you did run the test using the `swiftgen-cli` scheme? Maybe there was another scheme selected by default and you didn't actually run the test you thought?

Anyway, I'm adding them myself right now, so don't bother, I'll do it when merging ;-)

Issue #7, opened 2015-07-22
Avoid duplication of cases


AliSoftware:
Sometimes the scripts can generate multiple `cases` in an `enum` with the same name, because when converting a string to `asSwiftIdentifier()` may lead to the same result for different inputs.

For example if we have a storyboard Scene named "some$Identifier" and another named "some#identifier", this will generate the same `case someIdentifier` for both.

We should instead keep the already-generated enum cases in a `Set`, and make it add an auto-incrementing suffix when we encounter duplicate, to generate `someIdentifier` for the first one, but use `case someIdentifier2` the second time `asSwiftIdentifier()` return `someIdentifier`, and `case someIdentifier3` the third time, etc (then `someIdentifier5` the fourth time if we imagine `someIdentifier4` already exist because we have a scene named `some$identifier#4` that generated it for example, etc)

AliSoftware:
This issue could be the occasion to double-check that we use the same `forbiddenChars` everywhere (`_`) when calling `asSwiftIdentifier` in the context of our scripts

> (We could also make that `_` the default value for `forbiddenChars` in  `SwiftIdentifier.swift` but I don't like the idea because `SwiftIdentifier` is thought to be generic, possibly used by others, and we can't impose that `forbiddenChars` default value, as the expected result when calling `str.asSwiftIdentifier()` should be that it respects the official chars and don't add unexpected exceptions unless explicitly asked)

ldiqual:
:+1: Additionally, segue identifiers don't have to be unique in a same storyboard so there's definitely a high collision chance.

AliSoftware:
To implement this I see some new `IdentifierGenerator` class to which we can feed a string, it computes the associated identifier, check if it's already used and if so suffix it with some index, and then store it in the list of used identifiers and return it.

But it's probably better to wait for #24 (or do it alongside with #24) before implementing this, as using Stencil to have templates may change architecture choices on the API of this `IdentifierGenerator` so it can be easily used in a template context.

AliSoftware:
Some considerations to take into account when we'll be implementing this:

* Do the de-duplication at the template level, not directly in the Swift code of the Parsers, because anyone can generate the `case` names however they like. Some could use `segue.identifier|swiftIdentifier`, others could `segue.identifier|uppercase|swiftIdentifier` or `segue.identifier|swiftIdentifier|snakeToCamelCase` etc
[...]

PR #43, opened 2015-10-27
Do not generate duplicate enum cases for segue IDs

filwag:


AliSoftware:
What about Segues that have different names but that could result in the same `swiftIdentifier`? Like `Foo-Bar` and `Foo_Bar` both generating the case name `Foo_Bar`…

AliSoftware:
Note that this is also discussed in #7 where we discuss the various possible cases. Like:

* the fact that sometimes we need to generate the case anyway (and thus we need to append an index increment to it to make it distinct) because it actually doesn't represent the same Segue with the same identifier (like `case Foo_Bar = "Foo-Bar"` vs. `case Foo_Bar2 = "Foo_Bar"`)
* The idea that this shouldn't be managed at the code level, but rather at the template level, because it depends on how the user chose to generate the swift identifier for that segue (maybe some people will use `segue.identifier|swiftIdentifier` but some other will use `segue.identifier|swiftIdentifier|uppercase` because they prefer having their cases all in uppercase of whatnot… and the important thing is that the generated final identifier is not a duplicate, so that `FooBar` and `FOObar` don't both generate `case FOOBAR`
* The fact that this logic should be applied to quite everything, not just segues but also images, colors, strings, …

filwag:
Thanks a lot for pointing this out!

What if we use nested enums as associated values to disambiguate? Something similar to this:
```swift
/* storyboard segues:
   my_orange
   myorange
   my_Pineapple
*/

// generated enum:
enum Fruit {
  enum MyOranges: String {
    case my_orange = "my_orange"
    case myorange = "myorange"
  }
  
  case MyOrange(MyOranges)
  case MyPineapple
  
  func string() -> String {
    switch self {
    case .MyPineapple: return "my_Pineapple"
    case .MyOrange(let orange): return orange.rawValue
    }
  }
}

let pineapple: Fruit = .MyPineapple
let orange: Fruit = .MyOrange(.my_orange)
```

Please note that this PR solves a smaller issue that is a bit different from `swfitIdentifier` transformation ambiguity issue (#7). I'm simply filtering out exact duplications on an early stage, this is pretty similar to what happens for images (please see `AssetsCatalogParser.swift:14`).

filwag:
My example above would not work if segue string includes dots.

AliSoftware:
I don't like the nested enums approach, not having all the enums of the same storyboard at the same level (some at the root and some nested) seems inconsistent, and it doesn't feel intuitive at the call site.
Additionally, it would make the `Stencil.Context` quite complex to grasp and the stencil templates harder to write.

GenumKit/Parsers/StoryboardParser.swift
```diff
@@ -66,6 +71,11 @@ public final class StoryboardParser {
           break;
         }
       }
+      
+      private func appendSegue(segue: Segue) {
```
AliSoftware: What about using a `Set<Segue>` instead of a `[Segue]` here, so we get that for free and won't need the `appendSegue` private method then, letting that work to the `Set`?

filwag: Good point, changed it to `Set<Segue>`.

GenumKit/Parsers/StoryboardParser.swift
```diff
@@ -87,3 +97,8 @@ public final class StoryboardParser {
     }
   }
 }
+
+extension StoryboardParser.Segue: Equatable { }
+func ==(lhs: StoryboardParser.Segue, rhs: StoryboardParser.Segue) -> Bool {
+  return lhs.segueID == rhs.segueID
```
AliSoftware: Feels strange to only compare `segueID`.

What if the user wants to create a template that use both `segueUD` and `customClass` to generate the `case` name?

Like some might prefer to have their template generating this: `case {{segue.identifier|swiftIdentifier}}_{{segue.customClass|swiftIdentifier}} = {{segue.identifier}}` and thus if two segues have the same identifier but different customClass that wouldn't generate a conflict, and the user won't have the code he expect to be generated.

filwag: Yep, good catch. There were no mentions of customClass in default template, so I've just ignored it which is not right. I'll address this.

AliSoftware:
I think for the sake of code consistency, migrating the `Scene` to a `struct` and the list of `Scene` into a `Set` could be nice?

filwag:
Sure, I'll update this too.

filwag:
Updated the PR according to the comments. My simple de-duplication now also takes `customClass` into account. It will still end up with duplicates for two segues with the same id but different custom classes when `customClass` is not used in the template. But there won't be duplicates when both params are equal. Other cases mentioned in #7 remain unresolved.

AliSoftware:
Let me review that properly once I'm on my Mac before merging, but from there in GitHub LGTM ;)

GenumKit/Parsers/StoryboardParser.swift
```diff
@@ -63,7 +72,7 @@ public final class StoryboardParser {
         case "connections":
           readyForConnections = false
         default:
-          break;
```
AliSoftware: :joy: dammit Objective-C habits ^^

AliSoftware:
Hey @filwag 

So I reviewed that code rn and I just miss a couple of things:

* Now that we adopted `Sets`, I think it'd be more logical to [expose the `storyboardScenes` and `storyboardSegues` dictionaries](https://github.com/AliSoftware/SwiftGen/pull/43/files#diff-78a5fbbd777e77426abc79d34854beb8L12) as `[String: Set<Scene>]` and `[String: Set<Segue>]` directly, [instead of converting them back to `Array`](https://github.com/AliSoftware/SwiftGen/pull/43/files#diff-78a5fbbd777e77426abc79d34854beb8R85), and do the conversion `Array(scene)` and `Array(segue)` [here instead](https://github.com/AliSoftware/SwiftGen/blob/0804b488d5c1812ba9e9d55e2492a77099578162/GenumKit/Stencil/Contexts.swift#L64), for more consistency.
* Please also add an entry in the `CHANGELOG.md` to credit yourself :wink:

filwag:
Hi @AliSoftware, thanks for the review, I'll work on that :)

filwag:
Updated. @AliSoftware, explicit `Array(scenes)` conversion [here](https://github.com/AliSoftware/SwiftGen/blob/0804b488d5c1812ba9e9d55e2492a77099578162/GenumKit/Stencil/Contexts.swift#L64) is not required, because `sort()` returns an `Array`.

AliSoftware:
Merged via e85f7c184a07efee34a452949ea1c7f87ff51a47

Thanks again for your work on this!

Issue #42, opened 2015-10-26
Store custom templates in ApplicationSupport


AliSoftware:
The idea of this feature would be to:

* Create a `swiftgen template` subcommand that could:
  * with no parameter: list all the installed templates in `~/Library/Application Support/SwiftGen/templates`
  * `--add FILE` copy `FILE` into `~/Library/Application Support/SwiftGen/templates` to install it as template
  * `--open`: Opens `~/Library/Application Support/SwiftGen/templates` in the Finder
* Add an option parameter to all `swiftgen` subcommands to specify template by name
  * The option flag will be something like `--templateName NAME` / `-n NAME`
  * It will search for a template named `NAME.stencil` in `~/Library/Application Support/SwiftGen/templates` first
  * If not found, It will search for a template named `NAME.stencil` in the bundled templates (installed in `TEMPLATES_RELATIVE_PATH`)

This will allow people to store/manage/use their SwiftGen templates in an easier way, instead of having to use the full path with `--template` each time. Note that the `--template` option will still work, and using both options should lead to an error.

See also https://github.com/AliSoftware/SwiftGen/issues/31#issuecomment-151088648

AliSoftware:
Also note that `--template` will now default to `nil` and the value having a default will now be `--templateName` instead.
For example, for `swiftgen colors`, specifying neither `--template` nor `--templateName` would default to the same as using `--templateName colors`

The positive side effect that this would have is that people would be able to simply store their preferred default template for `swiftgen colors` simply in `~/Library/Application Support/SwiftGen/templates/colors.stencil` and it will be picked by default instead of the one bundled with SwiftGen. This way, no need to specify `--templateName` each time if you have your own custom default templates!

PR #44, opened 2015-10-28
Feature/adding all values in images enum

GUL-:
Add an array in the images enum so we can iterate over them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,8 @@ extension UIImage {
     case Banana = "Banana"
     case BigPear = "Big_Pear"
     case StopButtonEnabled = "stop.button.enabled"
+
+    static let allValues = [GreenApple, RedApple, Banana, BigPear, StopButtonEnabled, ]
     
     var image: UIImage {
       return UIImage(named: self.rawValue)!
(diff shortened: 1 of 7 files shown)


AliSoftware:
I think that feature is very rarely useful (I never had to iterate over all the images in my Assets Catalog, and don't think most of people do), so I think that'll make the generated file quite long (like when people have 100+ images in their assets catalog)

So here's what I suggest:

1. Keep the existing template unchanged (and revert the examples in README & Playground)
2. Create **a separate template** that adds what you did so we can have this `allValues` entry, but as an opt-in option. People wanting to have this could simply use your template instead of the default one.
3. I'll then include your template alongside mine when SwiftGen is installed. Once #42 is merged, which would allow people to use `--templateName` to specify a template by name instead of by full path, this would make it easier for people to choose the default template without `allValues` or your template with it

AliSoftware:
Merged via d430f568a7cf98abfa58c94b3bf029db2853362d

GUL-:
Thank you very much to accept this MR :-) :+1:

AliSoftware:
Now that #50 is merged, you'll be able to use `swiftgen images -t allvalues PATH` in order to use this specific template which will be bundled with SwiftGen in the next release.

PR #53, opened 2015-11-04
Added font support via plist file in swiftgen.

joseprl89:
I'm afraid I don't have enough time today to finish the whole font support, however most of it is already done.

I think I have some issue with the XCTDiffStrings method and the encoding I use in the files I created, I assume it should be easy to sort out if you know that method well. (Also maybe worth taking a look at it?).

Hope you can pick this up, since I won't be able to follow up much in the following days.

Cheers

### DIFF
--- a/GenumKit/Parsers/FontsFileParser.swift
+++ b/GenumKit/Parsers/FontsFileParser.swift
@@ -1,0 +1,58 @@
+//
+// GenumKit
+// Copyright (c) 2015 Josep Rodriguez
+// MIT Licence
+//
+
+import Foundation
+
+public typealias Font = (fontName:String,fontSize:Int)
+
+public enum FontFileParserError: ErrorType {
+	
+	/// Thrown when the path specified contains no data.
+	case NoSuchFile
+
+	/// A required parameter was not specified
+	case MissingParameter(String)
+}
+
+public final class FontFileParser {
+	
+	var fonts = [String:Font]()
+	
+	public init() {}
+	
+	public func addFontWithName(name: String, value: Font) {
+		fonts[name] = value
+	}
+	
+	/// Text file expected to be in plist format, containg an array of dictionaries
+	/// with the following keys:
+	///  - Key
+	///  - FontName
+	///  - FontSize
+	public func parseTextFile(path: String, separator: String = ":") throws {
+		guard let fonts = NSArray(contentsOfFile: path) as? [[String:AnyObject]] else {
+			throw FontFileParserError.NoSuchFile
+		}
+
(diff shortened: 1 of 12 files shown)


AliSoftware:
Thanks so much for this PR, looks quite promising!
Will take a look when I have a minute to see what's wrong with the tests :-)

---

`XCTDiffStrings` is a function of mine that typically works like `XCTAssertEqual` but when it detects that the two strings to compare (actual result vs. expected result) differs, it prints them properly in the console with line numbers in front of the code to compare, and telling you at which line the error was.
That helped me a lot figuring out what was the difference when comparing the actual generated code vs. the expected code, because `XCTAssertEqual` would only print a condensed version of the string, with `\n` all over instead of actual CRLFs, which made it very hard to figure out where the mismatch was in the generated code.

So if your test fails, you can open the Console output in Xcode (Cmd-Shift-Y) and check in the console log what that `XCTDiffStrings` printed there, and it will tell you at which line it found a mismatch.

joseprl89:
Reviewed a bit my code this morning, and noticed a few issues with it (too many hours yesterday, my bad :P)

* I committed an output.swift file which was just a test output I created.
* The expected output of the unit tests has strings instead of floats in the size.

AliSoftware:
Here's some points this PR still needs addressing:

* [ ] Fix tests
* [ ] Playground Page to demo this new subcommand
* [ ] README paragraph to list and explain this new command
* [ ] CHANGELOG entry

Any help welcome to provide those.

AliSoftware:
@joseprl89 Now that #102 has been merged, this idea is back in the game and can be a nice addition to #102 ! But it's now a bit dated and the codebase has evolved (especially with #102 now merged) so your'll need to update your PR a bit now if you are still interested in making it happen :wink:

@AFcgi I saw in #102 that you were interested in a feature such as the one implemented here. I'm not sure if it would be easier to start from scratch now that #102 laid all the ground work in the `struct` & all necessary for that or you it's easier to get some inspiration from this PR #53 to rebase and adapt its code to the latest master, but feel free to contribute in any way to make this feature happen 👍

AliSoftware:
I think I'm gonna close this PR now, because now that Font support has been added in SwiftGen via #102, then:
1. this PR might not be as necessary as it was before, and
2. even if you still find it useful to define your fonts in a plist rather than let SwiftGen parse the fonts folder, I'm sure it'll be better to open a new PR starting from the code for the Fonts that is now in place rather than adapt this very PR.

Feel free to open a new PR for this feature (but starting with latest commits) if this is still a feature you need.

PR #60, opened 2015-11-11
Added doc comments for default translation of strings, fixes #58

### LINKED ISSUE #58: Documenting String key enum cases w/ default translation
Something that xcres has been doing for a while is to document each generated string key with the default translation for that key. That way, you can see the output string when alt-clicking the enum or using auto-complete which is really nice. Currently, the string value isn't exposed as a variable in the strings template engine.

Example input:
``` C
"confirmation_alert_title" = "Do you really want to do this action?";
"confirmation_alert_cancel_title" = "Abort";
```

Desired output:
``` Swift
enum L10n {
  /// Do you really want to do this action?
  case ConfirmationAlertTitle

  /// Abort
  case ConfirmationAlertCancelTitle
}
```

MrAlek:
See #58.


AliSoftware:
Perfect :ok_hand: thx!

PR #66, opened 2015-11-26
Added support for UITableViewCell & UICollectionViewCell identifiers …

### LINKED ISSUE #19: Support UITableViewCell identifiers

jmacmullin:
…in Storyboards.

Fix #19

### DIFF
--- a/GenumKit/Parsers/StoryboardParser.swift
+++ b/GenumKit/Parsers/StoryboardParser.swift
@@ -17,9 +17,15 @@ public final class StoryboardParser {
     let segueID: String
     let customClass: String?
   }
+    
+  struct Cell {
+    let reuseID: String
+    let customClass: String?
+  }
   
   var storyboardsScenes = [String: Set<Scene>]()
   var storyboardsSegues = [String: Set<Segue>]()
+  var storyboardsCells = [String: Set<Cell>]()
   
   public init() {}
   
@@ -29,8 +35,10 @@ public final class StoryboardParser {
     class ParserDelegate : NSObject, NSXMLParserDelegate {
       var scenes = Set<Scene>()
       var segues = Set<Segue>()
+      var cells = Set<Cell>()
       var inScene = false
       var readyForFirstObject = false
+      var readyForPrototypes = false
       var readyForConnections = false
       
       @objc func parser(parser: NSXMLParser, didStartElement elementName: String,
@@ -49,6 +57,13 @@ public final class StoryboardParser {
             scenes.insert(Scene(storyboardID: storyboardID, tag: tag, customClass: customClass))
           }
           readyForFirstObject = false
+        case "prototypes":
+            readyForPrototypes = true
+        case "tableViewCell" where readyForPrototypes:
+            if let reuseID = attributeDict["reuseIdentifier"] {
+                let customClass = attributeDict["customClass"]
+                cells.insert(Cell(reuseID: reuseID, customClass: customClass))
+            }
         case "connections":
           readyForConnections = true
         case "segue" where readyForConnections:
@@ -56,6 +71,11 @@ public final class StoryboardParser {
             let customClass = attributeDict["customClass"]
             segues.insert(Segue(segueID: segueID, customClass: customClass))
           }
+        case "collectionViewCell":
+            if let reuseID = attributeDict["reuseIdentifier"] {
+                let customClass = attributeDict["reuseIdentifier"]
+                cells.insert(Cell(reuseID: reuseID, customClass: customClass))
+            }
         default:
           break
         }
@@ -69,6 +89,8 @@ public final class StoryboardParser {
           inScene = false
         case "objects" where inScene:
           readyForFirstObject = false
+        case "prototypes":
+            readyForPrototypes = false
         case "connections":
           readyForConnections = false
         default:
@@ -84,6 +106,7 @@ public final class StoryboardParser {
     let storyboardName = ((path as NSString).lastPathComponent as NSString).stringByDeletingPathExtension
     storyboardsScenes[storyboardName] = delegate.scenes
     storyboardsSegues[storyboardName] = delegate.segues
+    storyboardsCells[storyboardName] = delegate.cells
   }
   
   public func parseDirectory(path: String) {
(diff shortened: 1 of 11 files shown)


AliSoftware:
Thx for the PR!

A few remarks:

* Don't forget about XIBs. Especially when the same template of UITableViewCells is used across multiple UITableViews in my project, I like to still design those cells in a dedicated XIB so it can be registered with multiple tableviews instead of being duplicated in each storyboard.
  * That means the tool should be able to parse the XML of XIBs as well as Storyboards, but I think those will probably be pretty similar
  * But also take into account that a `tableViewCell` XML node might not be attached to a particular cell, so the structure of the context might vary. We might need to keep both (e.g. the cells attached to a storyboard scene like you did + an special entry at the root of the context for  cells not attached to a scene, or whatnot)
* It seems that your templates only provide an extension for `UICollectionView` but not for `UITableView`?
* It would be pretty cool if we could find a way to make the `dequeueReusableCellWithReuseIdentifier` directly return the right subclass of cell instead of the generic `UICollectionViewCell` type. Might need some reflection on how to do that properly both in Swift in particular (what syntax) and the necessary changes in this code to allow SwiftGen to do that, but one idea would be to get inspiration from what I did for the Scenes methods returning the proper subclass of `UIViewController` for each Scene.


Especially in a lot of my projects, a cell class is often bound to a given cell identifier (I use the same cell identifier every time I use that cell class in my storyboard or XIB), so one idea would be to have some code like [this one], where the cell class has a method returning it's cellIdentifier, and we call `dequeue` on the cell class, passing the tableView as parameter, instead of the opposite like done in UIKit, allowing us to `return Self`.  
What do you think about that approach? To be able to make a template that support this kind of code, one might need to improve your code to at least provide the custom class associated to each cell identifier.

jmacmullin:
All good points :-)

I'll update my PR to address them, but probably not till some time next week now.

ilyapuchka:
It would be also cool not to forget about header/footer/supplementary views as they all work pretty the same way as cells.

Btw just one more variant of type-safe register/dequeue method:

```swift
protocol ReusableView: class {
  static var nib: UINib? { get }
  static var reuseIdentifier: String { get }
}

extension ReusableView {
  static var reuseIdentifier: String = { return "\(self)" }
  static var nib: UINib? { return UINib(nibName: "\(self)", bundle: nil)}
}

extension UITableView {
  func dequeReusableCell<T: ReusableView>(cell: ReusableView.Type, forIndexPath indexPath: NSIndexPath) -> T {
    return dequeueReusableCellWithIdentifier(cell.identifier, forIndexPath: indexPath) as! T //maybe it's even better to throw error here? though it's most likely unrecoverable
  }
  func registerReusableCell<T: ReusableView>(cell: ReusableView.Type) {
    if let nib = cell.nib {
       registerNib(nib, forCellReuseIdentifier: cell.identifier)
    }
    else {
      registerClass(cell, forCellReuseIdentifier: cell.identifier)
    }
  }
}
```

Igor-Palaguta:
@ilyapuchka it is not even required to pass cell class as param in dequeReusableCell
here is an example:
https://gist.github.com/Igor-Palaguta/8051da7c6854eb7f6d24
And how it looks:
let assetCell: AssetCell = collectionView.dequeueCellForIndexPath(indexPath)

ilyapuchka:
@Igor-Palaguta sure, there are different options

AliSoftware:
Ah yeah I actually prefer the code from Igor, more logical (dequeue still called on tableView/collectionView, one less parameter…) and it's actually a lot like what we use in Dip (using a generic function that returns the generic parameter and leverage Swift type inference to detect that) so Ilya you should like it :wink:

AliSoftware:
Given the existence of [Reusable](https://github.com/AliSoftware/Reusable) now, not sure if this feature is still worth it?

aabdellah:
It's good for cells where they don't have a custom class, Reusable requires one. could it be an optional feature?

phatblat:
Personally, I'd like to see SwiftGen support this. While Reusable is fabulous for enabling very clean table an collection view cell handling code, it is still vulnerable to string mismatches at runtime. With [the right template](https://github.com/AliSoftware/SwiftGen/issues/19#issuecomment-223788486) SwiftGen can compliment Reusable to possibly eliminate that sort of crash, and at the very lease move issues to compile time.

AliSoftware:
@phatblat Your idea of having a template that generate the `extension FooCell { let reuseIdentifier = "<GeneratdString>" }` sounds good to me. Let's make that template possible!

My remarks in the first comment are still true though, we need to address those issues while parsing the IB files, especially cells in XIBs..

AliSoftware:
Hi all,

First of all, thanks a lot again for taking the time to create this PR @jmacmullin , and @Igor-Palaguta and @ilyapuchka to come up with additional ideas.

I'd like to apologise for letting that feature stale for so long. SwiftGen goals and priorities, including a big refactoring and splitting the project into multiple repositories, took precedence at some point, and we let that PR a bit forgotten.

Given that this PR is quite old and has gotten out of sync with how SwiftGen evolved, I'm gonna close this in favor of https://github.com/SwiftGen/SwiftGenKit/pull/16. We can continue working on this feature in SwiftGenKit from this PR from now on.

To be honest, the next main milestone for SwiftGen and what we're focusing on right now is migrating to a new major version 5.0, removing deprecated template variables and make some cleanup after all that refactoring we did in 4.2. But once it's released (which I hope we won't take too long doing), it'll be time again for adding new feature in this newly cleaned codebase :wink:

---

To thank you for your work and for the time you took to improve the tool, I've invited you to join SwiftGen's GitHub organization's "CoreContributors" team. No pressure to accept nor any obligation in that, don't worry :wink: It will just give you the ability to help triage issues or push directly branches to the repo (instead of your fork)… in case you want to contribute again in the future or help finally make that feature happen 😉 

For more information about contributing to SwiftGen in this new GitHub organisation and context, you can read the [COMMUNITY.md](https://github.com/SwiftGen/SwiftGen/blob/master/COMMUNITY.md) and [CONTRIBUTING.md](https://github.com/SwiftGen/SwiftGen/blob/master/CONTRIBUTING.md) documents, and if you have any questions don't hesitate to reach out!

PR #73, opened 2015-12-17
Add == operator for StoryboardSegueType and String?

ikesyo:
This would be useful for use in `prepareForSegue`.

### DIFF
--- a/Pods/Pods.xcodeproj/xcshareddata/xcschemes/GenumKit.xcscheme
+++ b/Pods/Pods.xcodeproj/xcshareddata/xcschemes/GenumKit.xcscheme
@@ -7,30 +7,33 @@
       buildImplicitDependencies = "YES">
       <BuildActionEntries>
          <BuildActionEntry
-            buildForAnalyzing = "YES"
             buildForTesting = "YES"
             buildForRunning = "YES"
             buildForProfiling = "YES"
-            buildForArchiving = "YES">
+            buildForArchiving = "YES"
+            buildForAnalyzing = "YES">
             <BuildableReference
-               BuildableIdentifier = 'primary'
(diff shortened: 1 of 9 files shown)


AliSoftware:
I'm ok with this addition, but just for curiosity, why do you like this better than using a `switch` on the enum?

If feel using a `switch` ensures that you cover all the cases in your `prepareForSegue`, as demonstrated at [the very bottom of this playground page](https://github.com/AliSoftware/SwiftGen/blob/master/SwiftGen.playground/Pages/Storyboards-Demo.xcplaygroundpage/Contents.swift#L113-L120)

```
override func prepareForSegue(_ segue: UIStoryboardSegue, sender sender: AnyObject?) {
  switch UIStoryboard.Segue.Message(rawValue: segue.identifier)! {
  case .Custom:
    // Prepare for your custom segue transition
  case .NonCustom:
    // Pass in information to the destination View Controller
  }
}
```

What's the rationale for using `==` instead here?

ikesyo:
Thanks for the tip! I didn't realize the example and it looks good. Though we need some tweaks to let the code:

```swift
override func prepareForSegue(_ segue: UIStoryboardSegue, sender sender: AnyObject?) {
  // `segue.identifier` returns `String?`, so we can't use the value as is for `rawValue` since the type is plain `String`.
  switch StoryboardSegue.Message(rawValue: segue.identifier ?? "") {
  // Optional pattern can be used here which is from Swift 2: https://developer.apple.com/library/ios/documentation/Swift/Conceptual/Swift_Programming_Language/Patterns.html#//apple_ref/swift/grammar/optional-pattern
  case .Custom?:
    // Prepare for your custom segue transition
  case .NonCustom?:
    // Pass in information to the destination View Controller
  case nil:
    break
  }
}
```

I'm closing this in favor of the style above.

PR #77, opened 2015-12-31
Updated ReadME to use StoryboardScene and StoryboardSegue

AJ9:
When following the README Xcode complained of the following problem:

Code:

`UIStoryboard.Scene.Login.Start` 

Error:

`Type 'UIStoryboard' has no member 'Scene'`

This is due to UIStoryboard not being extended, a solution is provided where `UIStoryboard.Scene` is replaced with `StoryboardScene`.

The same is true for `UIStoryboard.Segue` to `StoryboardSegue`

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,14 +290,14 @@ struct StoryboardSegue {
 
 ```swift
 // Initial VC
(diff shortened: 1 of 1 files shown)


AliSoftware:
Well spotted! :+1:

PR #79, opened 2016-01-03
Update colons to pass SwiftLint

AJ9:
To use SwiftGen in tandem with [SwiftLint](https://github.com/realm/SwiftLint) the colons on `strings-default.stencil`, `storyboards-default.stencil` and `images-default.stencil` need to be updated to remove the prefixing space.

I realise this is a trivial improvement, but will reduce compiler warnings when used with SwiftLint!

storyboard stencil
image stencil

Pull request recreated after Master was updated. [Old Pull Request](https://github.com/AliSoftware/SwiftGen/pull/78)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # SwiftGen CHANGELOG
 
 ---
+## 0.7.5
+
+### Enhancements
+
+*  Updated stencils and unit tests to pass [SwiftLint](https://github.com/realm/SwiftLint).  
+  [Adam Gask](https://github.com/AJ9), [#79](https://github.com/AliSoftware/SwiftGen/pull/79)
+
 ## 0.7.4
 
 #### Enhancements
--- a/UnitTests/expected/Colors-List-RawValue.swift.out
+++ b/UnitTests/expected/Colors-List-RawValue.swift.out
@@ -14,7 +14,7 @@ extension UIColor {
 }
 
 extension UIColor {
-  enum Name : UInt32 {
+  enum Name: UInt32 {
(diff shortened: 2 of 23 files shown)


AliSoftware:
:ok_hand:
Coule you just add a commit to log this and credit yourself in the CHANGELOG? Thx!

AJ9:
Squashed the commits and updated the CHANGELOG, hoping that's what you meant? :grin:

AliSoftware:
Yep! :+1:
 (There was no need to squash the commits but it's ok, most important thing is to credit your changes)

AliSoftware:
@AJ9 Any idea how we could integrate a proper SwiftLint check in `.travis.yml` to avoid any future regression (like if we add more templates in the future, make sure they are also passing SwiftLint)?

I was thinking making `.travis.yml` install `swiftlint` then run it on the `expected` files in the Unit Tests, but it would be even better to actually make it _fail_ (and not just warn) on any warning to make the Travis-CI build fail if there were errors.
Also, not sure if the best move would be to add the `swiftlint` command to be run by `.travis.yml`, or to configure the Xcode project to have it run as a Script Build Phase so it also warns when using the SwiftGen xcodeproj and Cmd-U etc (but in such case it shouldn't maybe fail but just warn then…?)

AJ9:
Why not have both? I think adding a build script to the Xcode project will improve code quality and is a great idea. If it's implemented as suggested [here](https://github.com/realm/SwiftLint#xcode) - it will not require other contributors to necessarily have the tool installed (It'll quietly fail during builds/tests and doesn't make SwiftLint a hard dependancy for SwiftGen), but will be flagged during the Travis build - and those errors could be fixed without SwiftLint.

I'll make a PR for the Build Script - I've not worked with Travis before... something new to learn!

PR #80, opened 2016-01-03
Add SwiftLint to build script

AJ9:
As discussed in #79 and [here](https://github.com/AliSoftware/SwiftGen/pull/79#issuecomment-168542621), adding SwiftLint as a build script will reduce regression SwiftLint errors. This build script does check the entire SwiftGen project and I've added the following exclusions (to the `.swiftlint.yml` file) with the following reasons:

```
disabled_rules: 
  - force_try
  - force_cast 
  - variable_name 
```

* force_try - Used in some unit tests to create `result` objects
* force_cast - Used [here](https://github.com/AliSoftware/SwiftGen/blob/master/GenumKit/Stencil/SwiftIdentifier.swift#L71) and [here](https://github.com/AliSoftware/SwiftGen/blob/master/SwiftGen.playground/Pages/Storyboards-Demo.xcplaygroundpage/Contents.swift#L56)
* variable_name - Used sensibly [here](https://github.com/AliSoftware/SwiftGen/blob/master/swiftgen-cli/main.swift#L13)

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,0 +1,5 @@
+disabled_rules: # rule identifiers to exclude from running
+  - force_try
+  - force_cast
+excluded: # paths to ignore during linting. Takes precedence over `included`.
+  - Pods
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,5 +1,7 @@
 disabled_rules: # rule identifiers to exclude from running
   - force_try
   - force_cast
+  - variable_name 
+
 excluded: # paths to ignore during linting. Takes precedence over `included`.
   - Pods
--- a/SwiftGen.xcodeproj/project.pbxproj
+++ b/SwiftGen.xcodeproj/project.pbxproj
@@ -357,6 +357,7 @@
 				09A87B3E1BCC9B8000D9B9F5 /* Frameworks */,
 				09A87B3F1BCC9B8000D9B9F5 /* CopyFiles */,
 				1B796448B0837DA367438799 /* Copy Pods Resources */,
+				59962C4A1C39D38A00F68B7A /* SwiftLint */,
 			);
 			buildRules = (
 			);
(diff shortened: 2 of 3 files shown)


AJ9:
This does result in 242 issues, maybe we can find a way of just checking the `expected` folder?

AliSoftware:
great work thx!
I think we might get rid of the `try!` calls by replacing them with `guard try? … else { fatalError("some explicit message") }` instead, which would both satisfy SwiftLint and make better error messages, as [I hate killing ponies](http://alisoftware.github.io/swift/2015/09/06/thinking-in-swift-1/) :wink:

AliSoftware:
> This does result in 242 issues, maybe we can find a way of just checking the expected folder?

You can use `swiftlint lint --path …/expected` for that.

The main concern is indeed to ensure the code _generated_ by SwiftGen match SwiftLint styling rules, so the `expected` folder is indeed the most important one to check. But it would be cool to make the codebase of SwiftLint itself cleaner by addressing those 242 warnings too.

I think we could use `swiftlint lint --strict --path …/path/to/expected` in the `.travis.yml` file, so that Continuous Integration only check for `expected` folder, but fail even on warning (`--strict`).

And then use `swiftlint` in the Script Build Phase to check the whole codebase (and not just expected) but only warn and not fail, and (later?) address those lint warnings in the SwiftGen codebase itself.

AliSoftware:
I just realized a problem: `swiftlint` won't be able to lint the files in `UnitTests/expected` because those files have the `.swift.out` extension, not just `.swift`, so they are discarded by SwiftLint and not analyzed. SwiftLint only analyze files having the `.swift` extension.

So I'm not sure there is a proper way to validate that the templates or their output does not violate SwiftLint rules. Maybe there is hope with the `swiftlint lint --use-stdin` somehow, but not with `swiftlint` directly analyzing the files.

AliSoftware:
Hi @AJ9 

I finally took the time to merge this. In fact I reintegrated SwiftLint the way you started to do it (but in my own branch to start anew since your branch started to be a bit behind the latest merged PRs), and in addition I added linting for both the templates and the SwiftGen code itself, and fixed all the warnings & errors generated by the listing.

Thanks again for taking a stab at it!

PR #81, opened 2016-01-04
Added clr files parser

ilyapuchka:
This PR adds support for clr files using the same templates used for text files.


GenumKit/Parsers/ColorsFileParser.swift
```diff
@@ -66,3 +73,47 @@ public final class ColorsFileParser {
...
+
+public final class CLRFileParser: ColorsFileParser {
+  public private(set) var colors = [String: UInt32]()
+  
+  public init() {}
+  
+  public func parseFile(path: String) {
+    if let colorsList = NSColorList(name: "UserColors", fromFile: path) {
+      for keyColorPair in colorsList.allKeys.map({ ($0, colorsList.colorWithKey($0)) }) {
+        let colorName = keyColorPair.0.stringByReplacingOccurrencesOfString(" ", withString: "")
```
AliSoftware: Why is the removal of spaces necessary here?

GenumKit/Parsers/ColorsFileParser.swift
```diff
@@ -66,3 +73,47 @@ public final class ColorsFileParser {
...
+
+//MARK: - CLR File Parser
+
+public final class CLRFileParser: ColorsFileParser {
+  public private(set) var colors = [String: UInt32]()
+  
+  public init() {}
+  
+  public func parseFile(path: String) {
+    if let colorsList = NSColorList(name: "UserColors", fromFile: path) {
```
AliSoftware: Wait… wow… that's all it takes to parse a `.clr` file? :astonished: #TIL

GenumKit/Parsers/ColorsFileParser.swift
```diff
@@ -66,3 +73,47 @@ public final class ColorsFileParser {
...
+      for keyColorPair in colorsList.allKeys.map({ ($0, colorsList.colorWithKey($0)) }) {
+        let colorName = keyColorPair.0.stringByReplacingOccurrencesOfString(" ", withString: "")
+        if let colorValue = keyColorPair.1?.rgbColor?.hexValue {
```
AliSoftware: Mouahaha I love Swift. `keyColorPair.1?.rgbColor`… seriously, how magic this `.1?` notation is :heart_eyes:

GenumKit/Parsers/ColorsFileParser.swift
```diff
@@ -66,3 +73,47 @@ public final class ColorsFileParser {
...
+          addColorWithName(colorName, value: colorValue)
+        }
+      }
+    }
+  }
+  
+  public func addColorWithName(name: String, value: String) {
+    addColorWithName(name, value: ColorsTextFileParser.parse(value))
```
AliSoftware: Not sure this method belongs to `CLRFileParser`, probably a copy/pasted bit of code you forgot to remove?

GenumKit/Parsers/ColorsFileParser.swift
```diff
@@ -66,3 +73,47 @@ public final class ColorsFileParser {
...
+  private var rgbColor: NSColor? {
+    return colorUsingColorSpaceName(NSCalibratedRGBColorSpace)
+  }
+  
+  private var hexValue: UInt32 {
+    let hexRed = ((UInt32(redComponent * 255.0) << 8) << 8) << 8
+    let hexGreen = (UInt32(greenComponent * 255.0) << 8) << 8
+    let hexBlue = UInt32(blueComponent * 255.0) << 8
+    let hexAlpha = UInt32(alphaComponent * 255.0)
+    return hexRed + hexGreen + hexBlue + hexAlpha
```
AliSoftware: You should use `|` here (bitwise-OR) and not `+` (arithmetic addition) to combine the different bytes together.

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 
 *  Updated stencils and unit tests to pass [SwiftLint](https://github.com/realm/SwiftLint).  
   [Adam Gask](https://github.com/AJ9), [#79](https://github.com/AliSoftware/SwiftGen/pull/79)
+* Added support for `*.clr` files.
```
AliSoftware: Maybe explain a bit what those files are here. I wasn't aware of this files extension before, and just called them "system color palettes" so that "`.*.clr` files" didn't ring a bell to me first.
Just adding "(System Color Palettes)" should help — maybe a link to the the README too?

GenumKit/Parsers/ColorsFileParser.swift
ilyapuchka: In palettes who knows what users can have. At least we provide some sanity checks. Of course if user have Color#1 like names it will not work anyway.

README.md
```diff
@@ -321,6 +321,14 @@ This will generate a `enum Name` in an extension of `UIColor`, with one `case` p
 
 The text file is expected to have one line per color to register, each line being composed by the Name to give to the color, followed by ":", followed by the Hex representation of the color (like `rrggbb` or `rrggbbaa`, optionally prefixed by `#` or `0x`). Whitespaces are ignored.
 
+SwiftGen can also read `*.clr` files. For example you can use this comand to generate colors from one of your system palletes:
```
AliSoftware: `palletes` -> `palettes`

Would also be interesting to link to Natasha's blog post which explain how to create those color palettes and where they are located in ~/Library`, to encourage people to use them.

GenumKit/Parsers/ColorsFileParser.swift
ilyapuchka: That's used by tests, they are exactly the same as for text file. to make sure that they both generate the same result using shared method. Do you think we don't need them?

GenumKit/Parsers/ColorsFileParser.swift
AliSoftware: But you don't need that. It's not the work of the parser to sanitize the color name. It's the work of the template filters, like [here](https://github.com/AliSoftware/SwiftGen/blob/master/templates/colors-default.stencil#L23).

That's because people could want to customize their template and use the real name (with spaces and all) for purposes other than just use them as the swift identifier, but to use them to generate comments for each color, or to associate a color name as a String or implement a nice CustomStringConvertible or whatnot. But if they use it as a swift identifier, that's their responsibility to call the `swiftIdentifier` Stencil filter on the name in their template in the appropriate place(s).

GenumKit/Parsers/ColorsFileParser.swift
AliSoftware: No I think we should remove them, that does not make sense that the `CLRFileParser` has a method to add a color by passing its value as an hex `String` (whereas it makes sense for the `ColorsTextFileParser` because that's the format which is expected in the text file).

We shouldn't clobber the API of the class _just_ because it's easier to test.

GenumKit/Parsers/ColorsFileParser.swift
ilyapuchka: Maybe this will make it more readable?

```swift
for nameColorPair in colorsList.allKeys.map(nameColorPair) {
  addColorWithName(nameColorPair.name, value: nameColorPair.value)
}

func nameColorPair(key: String) -> (name: String, value: UInt32)? {
  return colorsList.colorWithKey(key).rgbColor.map({ (key: key.stringByReplacingOccurrencesOfString(" ", withString: ""), color: $0.hexValue) })
}
```

AliSoftware: No I'm fine with your version, I was just loving how we could make amazing stuff like this with Swift  :wink: 

But if you had to change, I'm not sure using `map` is worth it there. Why not simply:
```
for colorName in colorsList.allKeys {
  if let colorValue = colorsList.colorWithKey(colorName)?.rgbColor?.hexValue {
    ...
  }
}
```

AliSoftware:
Aside from the 1-2 comments I did on your last commit, looks good to merge, nice work :+1:

PR #83, opened 2016-01-15
Fixed typos: 'instanciate' -> 'instantiate'

pankolega:
'nuff said :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,12 +11,12 @@
   [Olivier Halligon](https://github.com/AliSoftware)
    * Now `cases` names are suffixed with `…Scene` and `static func` are prefixed with `instantiate…` to lower the risks of a name conflict with your ViewController classes.
    * The old template is still available but has been renamed `storyboards-uppercase.stencil`
-* Added support for `*.clr` files (files to store `NSColorList`'s presented in Color Picker on "Color Palettes" tab)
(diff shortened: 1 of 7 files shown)


darecki:
Great idea! That typo was bothering me very much.

AliSoftware:
Damn, that's one of those words I'll misspell every time due to the different spelling of the same word in French :laughing: Well spotted, sir, thx!

AliSoftware:
Could you just maybe add an entry in the `CHANGELOG.md` to credit yourself and make note that the `storyboards-default` template changed?

It may seem like a very little change, but every change deserve attribution, and as a lot of people probably use the default template this will change their generated code and will have to adapt their codebase, so it's worth mentioning anyway imho.

pankolega:
Sure thing. It is definitely a breaking change.

AliSoftware:
Sorry I should have been more explicit: version `0.7.5` is already released to Homebrew, without that change. So the entry in the changelog should be in a new `## Master` entry, ready for the next release 
(which will probably be 0.7.6, but we don't know yet, as I might wait for some other issues to be fixed before releasing a new version and those new issues might deserve a `0.8.0`, idk yet)

pankolega:
That's right. I didn't want to do that without your explicit instruction.

AliSoftware:
Perfect :ok_hand:  thx for your contribution @pankolega !

phatblat:
Yay! Was just about to create a PR for this. English is weird. :confused:

Issue #39, opened 2015-10-23
Support other names than `Localizable.strings` as input (`tableName` parameter)


iGranDav:
Currently the generated `tr` function is using `NSLocalizedString(key, comment: "")` while it should be using `NSLocalizedString(key, tableName: <tableName>, comment: "")` if the user provide a different `tableName`

AliSoftware:
#24 is now merged: it should now be quite easy to make a PR to support this.

Ideas for the first steps to this:
* modify [`strings.swift` here](https://github.com/AliSoftware/SwiftGen/blob/master/swiftgen-cli/strings.swift#L23) to inject the input file basename (without the `.strings` extension) [in the `stencilContext` here](https://github.com/AliSoftware/SwiftGen/blob/master/GenumKit/Stencil/Contexts.swift#L90-L100).
* Then [alter the default template here](https://github.com/AliSoftware/SwiftGen/blob/master/templates/strings.stencil#L29) to use this new variable from the context (maybe make an exception if that variable is equal to the default `Localizable` table name?)
* Don't forget Unit Tests & CHANGELOG

Then next steps would be to fix #32 and #33 so that we can generate multiple enums at once in the same file, instead of having to call `swiftgen` once for each `.strings` to generate.

joshavant:
So, I thought about this considerably today, and ran into a few tricky issues.

In Swift, `enum`s (and, so it follows, all of their `case`s) must be declared at compile-time. Said another way, we can't take advantage of composition to add `case`s to a previously-declared `enum`.

What this restriction means for SwiftGen users who are trying to generate a Swift `enum` type derived from a collection of separate `.strings` files is: all `.strings` files would necessarily have to be specified at SwiftGen execution-time. (This assumes we want to stick with generating a single, parent `L10n` `enum` type for strings, of course.)

I envisioned that would lead to a generated data structure interface that could look roughly like:
```
enum L10n {
    enum BaseTable { // i.e. Localizable.strings (special case)
        case AlertTitle
        case AlertMessage
        case Greetings(String, Int)
    }
    
    enum LoginTable { // i.e. Login.strings
        case SignUp
        case SignIn
        case ResetPassword
    }
    
    case Base(BaseTable)
    case Login(LoginTable)
}
```
(While we're at it, input welcomed on that interface, too!)

This would create a less flexible SwiftGen interface, but, on the flip side, it would provide a solution for something like #32.

[...]

Issue #32, opened 2015-10-14
Allow `swiftgen strings` to use a directory as input


AliSoftware:
`assets strings` should be allowed to take a directory as parameter and scan for every _.strings files in it.
This would allow people using `Base.lproj` and one `_.strings`file for each`.xib`or`.storyboard` to work too

iGranDav:
:+1: would be awesome!

AliSoftware:
This could also use #33 to parse multiple `Localizable.strings` files passed as argument instead of a dir

iGranDav:
Issue #39 could be related also

AliSoftware:
Ping'd in https://github.com/SwiftGen/SwiftGen/issues/497#issuecomment-422438917 with thoughts on cases where folders have multiple `.lproj` files

djbe:
This has been implemented by #570, released in SwiftGen 6.1.0 🎉

PR #86, opened 2016-02-02
Issue #39: NSLocalizedString table support

joshavant:
#39

Opening this PR for discussion.

I implemented the `tableName` param sent to  `StringsFileParser.stencilContext(enumName:tableName:)` in the `strings-default.stencil` template to generate a parent `L10n` struct w/ a single, nested `enum` which enumerates that table's strings.

This opens the way to refactoring the `tableName` argument (and subcommand input) into a collection of table names, which would allow support for multiple tables (#32).

**This PR will lead to a breaking change in the code generated by the `strings` subcommand!**

When everyone thinks it looks good, I'll wrap up this PR by pushing a final CHANGELOG diff.


README.md
```diff
@@ -188,19 +199,19 @@ func tr(key: L10n) -> String {
 Once the code has been generated by the script, you can use it this way in your Swift code:
 
 ```swift
-let title = L10n.AlertTitle.string
+let title = L10n.LocalizableTable.AlertTitle.localized
 // -> "Title of the Alert"
 
 // Alternative syntax, shorter
-let msg = tr(.AlertMessage)
+let msg = tr(L10n.LocalizableTable.AlertMessage)
```
AliSoftware: If we provide all the overloads necessary for `tr`, is the compiler intelligent enough to find the right enum for `tr(.AlertMessage)` (assuming only one of all the tables have a `case AlertMessage`)? Would be very cool if so, but not sure it's not too much to ask to the compiler…

AliSoftware:
Just glanced at the code in Github quickly, but looks nice from here, thanks!

---

Didn't check yet, but as not many people use custom tables, we'd have to make sure that the default template — the one not using multiple tables — would use the same `protocol Localizable` and method names, so it's just be a matter of doing a find&remplace `L10n.` -> `L10n.LocalizedTable.` on existing code to switch from the default template and default table to new template and multiple tables.

The default template would still be used by most projects — the one only using the default `Localizable.strings` table — while your new template will only be useful to projects using multiple strings table. But it would be cool to make it easy to switch from the default template to your new template for people suddenly deciding to add a different table in their project one day, so the protocols and functions names should be consistent between the two templates, and not sure that's the case (definitely not for the newly introduced `protocol Localizable`) yet.

README.md
joshavant: Yes! Good spot! After adding additional overloads, the compiler (and autocomplete) identified the nested type's shorthand case notation.

joshavant:
That's all taken care of. There's still only one strings template (`strings-default.stencil`).

I refactored that in such a way so, whether you're SwiftGen'ing one `.strings` file or multiple `.strings` files, the code that's generated will always have the same overall interface, just different and different numbers of 'nested `enum`s' within the parent `struct` (i.e. `L10n.LocalizableTable` is default, but if you generate `L10n.LoginTable`, it will still be consumed in the same way).

Of course, even with this diff, SwiftGen will only generate code for one `.strings` file, at a time. But the way I refactored the stencil in this PR lends itself to being refactored for #32 more easily (and with no breaking API changes).

To that end, the Global Find & Replace you described would, from what I understand, be exactly how someone currently using the default settings would migrate (but, hopefully someone who uses this feature more extensively than I could fact check this?).

AliSoftware:
Mmh not sure I'm following. You mean that ou changed the default templates directly? So than one that PR will be merged, everybody will *have to* use `L10n.LocalizeTable.Foo` even people only using the default `Localizable.strings` but no other table?

Given that the majority of projects only use the default table, I'd prefer to keep `L10n.Foo` for those people (using one template) and only have `L10n.LocalizedTable.Foo` for people using multiple tables and thus using another template (your new one). So people could choose between the simple template with only one level of enum and only default table, or new template if they use multiple tables.

Having everyone to type `L10n.LocalizedTable.Foo` even if the project only use the default table and no other seems painful for those user not caring for alternate tables.

joshavant:
Right. So, in my mind, I was primarily concerned about: should SwiftGen-generated code cause a codebase-wide breaking API change, if a user scales from 1x `.strings` file to multiple `.strings` files?

I focused my design on avoiding that, which it does. If, as you're suggesting, we relied on two different stencils (and, therefore, generated interfaces), it would cause that breaking API change (which seems bad, IMHO).

However, with fresh eyes on this issue (and considering this new feedback), I've had a flash of an idea on a design that could make everyone happy! Take a look...

##### Usecase: User SwiftGen's a `Localizable.strings` and `Login.strings` file:

```

import Foundation

protocol Localizable: CustomStringConvertible {
    var localized: String { get }
    
    static func tr(key: String, _ args: CVarArgType...) -> String
}

extension Localizable {
    var description: String { get { return self.localized } }
    
    static func tr(key: String, _ args: CVarArgType...) -> String {
        let format = NSLocalizedString(key, comment: "")
        return String(format: format, arguments: args)
    }
}

protocol TableLocalizable: Localizable {
    static var tableName: String { get }
    
    static func tr(table: String, _ key: String, _ params: CVarArgType...) -> String
}

extension TableLocalizable {
    static func tr(table: String, _ key: String, _ params: CVarArgType...) -> String {
        let format = NSLocalizedString(key, tableName: table, comment: "")
        return params.count == 0 ? format : String(format: format, args: params)
    }
}


enum L10n: Localizable {
    // Login.strings
    enum LoginTable: TableLocalizable {
        static let tableName = "Login"
        
        case LoginComplete
        case LoginFailed
        
        var localized: String {
            switch self {
            case .LoginComplete:
                return L10n.LoginTable.tr(LoginTable.tableName, "login_complete")
            case .LoginFailed:
                return L10n.LoginTable.tr(LoginTable.tableName, "login_failed")
            }
        }
    }
    
    // Localizable.strings
    case WelcomeToTheApp
    case WhatsYourName
    
    var localized: String {
        switch self {
        case .WelcomeToTheApp:
            return L10n.tr("welcome_to_the_app")
        case .WhatsYourName:
            return L10n.tr("whats_your_name")
        }
    }
}

func tr(key: Localizable) -> String { return key.localized }

// Localizable.string Usage Example
let welcome1 = L10n.WelcomeToTheApp.localized
let welcome2 = tr(L10n.WelcomeToTheApp)

// Login.strings Usage Example
let complete1 = L10n.LoginTable.LoginComplete.localized
let complete2 = tr(L10n.LoginTable.LoginComplete)
```
(above sample is a tested, functional code example)

New `strings` subcommand usage, and example generated code, related to above sample:
`./swiftgen strings Localizable.strings`: Generates same `enum L10n`, with no nested `enum LoginTable`
`./swiftgen strings Login.strings`: Generates `enum L10n`, which adopts `TableLocalizable` instead of `Localizable`
`./swiftgen strings StringsFolder/`: Generates sample above

Curious about your thoughts on this approach! If that seems like a good idea, I'll tweak my PR accordingly. Otherwise, I can take the 'two stencils' approach, from your reply.

AliSoftware:
I really like the idea, it's like the best of both worlds :+1: 

I thought about that but was afraid it would be quite hard to represent using Stencil, but maybe not

----

:bulb:  I think maybe the [use of the `include` tag in Stencil](https://github.com/kylef/Stencil/#include) might help here, to avoid repeating the same code for both the root enum and the nested enums?
Or you can of course use my `set` node instead to keep it all embedded in a single file, but I think there is potential for factorization inside the template anyway, would be worth trying to at least.

AndrewSB:
Looking great @joshavant, were you able to tweak the PR to use the new strategy yet? Anything I can do to help out?

joshavant:
Thanks for the offer @AndrewSB. I'm running a little slower than I'd like with this PR, due to busyness at work.

Got a little progress made this weekend, hope to finish it up this week!

joshavant:
Blah. So I'm a little blocked on this by a big project at work. I'm hoping to wrap this up next week.

If anyone is itching to get at this sooner, I wouldn't be offended by someone running with it. This PR is about half done on my end (Swift output structure determined, stencil's done), and I'd be glad to contribute those commits (after cleaning them up really quick) if someone wants them.

joshavant:
Currently pending this PR, so I can use the new Stencil `include` feature for this PR:
https://github.com/kylef/Stencil/pull/63

(Usecase: `GenumNamespace` does not get passed down into `include`'d templates, so all of its custom tags + filters are not available to `include`'d templates.)

joshavant:
Current status is that we need to wait for this Stencil commit to make it out into a production build:
https://github.com/kylef/Stencil/commit/bdc14ab1e121fdae32aac82a156db327b3bc7529

Presumably, this would be the next build (since it was committed only 8 days after 0.5.3 went out).

Why is this needed? In 0.5.3, the `include` feature of Stencil does not pass custom Namespace's into `include`'d templates in (so any `include`'d templates do not receive `GenumNamespace` and its custom tags + filters). The above commit _should_ fix that, once it is released.

Waiting and watching...

mgorbach:
Interestingly, it looks like my work on #93 should wait for this to be completed. I've got StringsParser building an arbitrarily deep nested structure of groups and entries, and now I need to figure out how to recursively generate structs or enums based on this structure. This is almost the same problem as this one, except it's arbitrary nested. The solution to this should really be the same as the solution to my issue.

AliSoftware:
@kylef Any visibility on making `Stencil 0.6.0.beta.1` out of beta, and into a production build?

That would allow me to merge this PR and @mgorbach to continue their work on #93 too

kylef:
@AliSoftware I've been meaning to find some time to finish up https://github.com/kylef/Stencil/pull/44 and get this into Stencil. I'd like to ship these changes in Stencil 0.6.0. 

I'll try get to this soon, sorry it's taking so long.

AliSoftware:
Given that it's been ~2 month since I had time to work on SwiftGen, and that I too am sorry to have taken so much time to merge SwiftGen PRs… I'll certainly not be the one blaming you for your delay 😄 
Good luck and keep us posted :+1:

AFcgi:
I need this feature too.

AliSoftware:
@AFcgi still waiting for Stencil 0.6.0 to be out of beta to be able to merge this, that's why it's still pending.

kylef:
@AliSoftware Any reason against just linking against the beta version of Stencil?

AliSoftware:
@kylef probably just the fact that I don't like depending on beta 😄 you're right we might give it a try.

kylef:
@AliSoftware Well, Stencil hasn't reached 1.0 yet either 😉.

kylef:
Not sure if you realised, but Stencil 0.6.0 has been out for 2 months now (https://github.com/kylef/Stencil/releases/tag/0.6.0). Is there anything blocking this PR from Stencil anymore?

AliSoftware:
Probably nothing more than having free time to rebase and merge, indeed.

Sorry everyone for taking so much time reviewing and merging lately (but @djbe has done an awesome job helping on triaging and reviewing, so things should get better :wink:)

djbe:
Swift 3 was only merged a couple of days ago, and that was a blocker for using Stencil 0.6.0.

djbe:
Hi @joshavant, with the latest SwiftGen release we've split the whole project into many sub-parts. If you'd like I can help you get started on reworking this to match the project changes.

The best place to start is by taking a look at these PRs:
- https://github.com/SwiftGen/SwiftGenKit/pull/13: adds a tables variable to the context
- https://github.com/SwiftGen/templates/pull/28

Once I get those merged, you're welcome to create a PR that adds support for `tablename` in the [templates](https://github.com/SwiftGen/templates) repository. If you're feeling up to it, you could also rewrite the strings templates to support multiple tables (currently we only support one table in one file).

AliSoftware:
Just FYI this has is being implemented in https://github.com/SwiftGen/SwiftGenKit/pull/41

PR #87, opened 2016-02-08
Fixed an issue if install locations prefixed with `~`

jessearmand:
The surrounding double quotes need to be removed,
to read the target path correctly.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -74,15 +74,15 @@ task 'install:light', [:bindir, :fmkdir, :tpldir] => :build do |_, args|
   (bindir, fmkdir, tpldir) = defaults(args)
 
   print_info "== Installing binary in #{bindir} =="
-  sh %Q(mkdir -p "#{bindir}")
-  sh %Q(cp -f "#{BUILD_DIR}/#{BIN_NAME}" "#{bindir}")
+  sh %Q(mkdir -p #{bindir})
+  sh %Q(cp -f "#{BUILD_DIR}/#{BIN_NAME}" #{bindir})
 
   print_info "== Installing frameworks in #{fmkdir} =="
-  sh %Q(mkdir -p "#{fmkdir}")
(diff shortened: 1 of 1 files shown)


AliSoftware:
Thanks for the PR

That's strange, because usually the double-quotes in `sh` and `bash` do interpret the content (so they replace `$XX` variables with their value at least, I thought they did the same for `~` too)

Removing the double-quotes to solve the solution is problematic though. Because then the script won't work with paths containing spaces then.

Maybe we can use ruby's `shell_escape` method instead and other helpers (or even use ruby's `Pathname` class to manipulate path properly) to cover both the `~` case and paths with spaces?

jessearmand:
Alright, I'll take a look into that later, I'm not very familiar with ruby filesystem or shell API.

I tested this issue on both zsh and bash.

AliSoftware:
Ok I looked at it a little further and I saw/remembered that my call to `defaults(binder, fmkdir, tpldir)` return a tuple of `Pathname` objects (and not plain `String` objects).
`Pathname` is a class in ruby dedicated to handle paths and already have handy functions to solve such issues, including a method named `expand_path` which expands any `~` in the `Pathname`.

So in short, the only change necessary to fix this issue is probably to change [this line](https://github.com/AliSoftware/SwiftGen/blob/master/Rakefile#L31) in the implementation of the `defaults(…)` function, to apply `expand_path` on each of the 3 `Pathnames`.

Which can be done by calling `expand_path` on each of the array items… or in a more functional and ruby way, by calling `map` on the array, passing it a reference to the method symbol:

```diff
- [bindir, fmkdir, tpldir]
+ [bindir, fmkdir, tpldir].map(&:expand_path)
```

By doing this change in the `Rakefile` **but** keeping the double-quotes in the rest of the code, this should hopefully cover both paths with `~` and with spaces in it.

PR #88, opened 2016-02-08
Fixed an issue with install paths prefixed with ~

jessearmand:
Expand paths for binary, template, and framework.


AliSoftware:
:+1: cool thanks!!

Did you have the occasion to test it both with `~` and path with spaces then?

jessearmand:
It still have a minor issue with the template directory. I tried to surround it with double quotes, but it generates another directory inside the template dir. 

Which is the 1st word of that template directory with space.

jessearmand:
Ah nevermind, my local templates folder was messed up until I add the surrounding double quotes. After I cleaned up, it's okay.

You can test these changes again. Thanks.

AliSoftware:
Thanks again for the PR, seems :ok_hand: now; merged! :tada:

Issue #89, opened 2016-02-08
Build loops when using storyboards


mathiasnagler:
I am using `swiftgen storyboards` in a project that has multiple storyboards. In those storyboards I use multiple `IBDesignable`s and `IBInspectable`s. Unfortunately this seems to cause build loops for me.

There is a wiki page in the `SwiftGen` project, that describes how to prevent build loops by using Xcode's support for custom `Input Files` and `Output Files`. I do not see how this works in my case however.

Using a single `Input File` is not possible as I have multiple storyboards. Using multiple `Input Files` is also impossible currently there is seemingly no support for multiple parameters in `SwiftGen`. I also tried to put all of my Storyboard in an extra folder and specifying this as my `Input File` but it seems as this does not work because Xcode does not correctly detect changes. Thus, the build phase is never executed.

Is there any workaround for my issue besides of disabling `Automatically Refresh Views` in Interface Builder?

AliSoftware:
Hi Mathias,

The normal solution should be to specify as the `Input Files` the directory SwiftGen will scan for storyboards, and as the `Output Files` the file generated by SwiftGen (like `Storyboards.swift` if you used `--output Storyboards.swift`)

According to the documentation but also to [this article](http://indiestack.com/2014/12/speeding-up-custom-script-phases/) — which actually tested using a directory as input "files" — this should normally work. So that's strange that you seem to says that "Xcode does not correctly detect changes" and the build phase is never executed :unamused: 

What you could try if the "directory as input file" solution doesn't work is:

* Either create a Makefile, which have a single target depending on every `*.storyboard` file and run the `swiftgen` command — The Makefile will take care of looking for modifications and only run the command if there is no modification
* Or make a Pull Request to SwiftGen so that `--output` will only write the string to the file if the file content is different, and won't touch the file if it's already the same content. That could actually solve multiple problems at once I guess, so it could be both easy to implement and a good solution for your use case and other's.

mathiasnagler:
Your second suggestion seems like a very sensible approach to me. I will try to make a Pull Request in the next couple of days.

AliSoftware:
Closed via #90

PR #90, opened 2016-02-11
Don't write unchanged contents to file

mathiasnagler:
Currently, `swiftgen` will always write the generated content to the `--output` file. This also happens when the generated content is exactly equal to what is already contained in the output file. 

Xcode will pick up the "change" regardless of the content and this behaviour can lead to build loops, especially when using `IBInspectable` and `IBDesignable`.

I changed `write:` to take a second parameter `onlyIfChanged:` that controls wether the generated content will be written to a file if it is exactly the same as already in the file. I also changed commands `strings`, `storyboards`, `images` and `colors` to use the new behaviour.

This idea was also discussed in #89.

### DIFF
--- a/swiftgen-cli/ArgumentsUtils.swift
+++ b/swiftgen-cli/ArgumentsUtils.swift
@@ -47,6 +47,10 @@ enum OutputDestination: ArgumentConvertible {
     }
   }
 
+  func writeIfChanged(content: String) {
+    write(content, onlyIfChanged: true)
+  }
+  
   func write(content: String, onlyIfChanged: Bool = false) {
     switch self {
     case .Console:
--- a/swiftgen-cli/ArgumentsUtils.swift
+++ b/swiftgen-cli/ArgumentsUtils.swift
@@ -47,12 +47,16 @@ enum OutputDestination: ArgumentConvertible {
     }
   }
 
-  func write(content: String) {
+  func write(content: String, onlyIfChanged: Bool = false) {
     switch self {
     case .Console:
       print(content)
     case .File(let path):
       do {
+        if try path.isContentEqualTo(content) && onlyIfChanged {
+          print("Skipping write as content is unchanged")
+          return
+        }
         try path.write(content)
         print("File written: \(path)")
       } catch let e as NSError {
--- a/swiftgen-cli/ArgumentsUtils.swift
+++ b/swiftgen-cli/ArgumentsUtils.swift
@@ -46,10 +46,6 @@ enum OutputDestination: ArgumentConvertible {
     case .File(let path): return path.description
     }
   }
-
-  func writeIfChanged(content: String) {
-    write(content, onlyIfChanged: true)
-  }
   
   func write(content: String, onlyIfChanged: Bool = false) {
     switch self {
--- a/swiftgen-cli/ArgumentsUtils.swift
+++ b/swiftgen-cli/ArgumentsUtils.swift
@@ -53,9 +53,8 @@ enum OutputDestination: ArgumentConvertible {
       print(content)
     case .File(let path):
       do {
-        if try path.isContentEqualTo(content) && onlyIfChanged {
-          print("Skipping write as content is unchanged")
-          return
+        if try onlyIfChanged && path.read(NSUTF8StringEncoding) == content {
+          return print("Not writing the file as content is unchanged")
         }
         try path.write(content)
         print("File written: \(path)")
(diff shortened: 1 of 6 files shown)


swiftgen-cli/ArgumentsUtils.swift
```diff
@@ -47,12 +47,20 @@ enum OutputDestination: ArgumentConvertible {
     }
   }
 
-  func write(content: String) {
+  func writeIfChanged(content: String) {
```
AliSoftware: I think this helper function is not really needed. Directly call `write(…, onlyIfChanged: true)` everywhere we want this behavior is sufficient, having a dedicating function just for that is probably not worth it :wink:

swiftgen-cli/ArgumentsUtils.swift
```diff
@@ -47,12 +47,20 @@ enum OutputDestination: ArgumentConvertible {
...
+    write(content, onlyIfChanged: true)
+  }
+  
+  func write(content: String, onlyIfChanged: Bool = false) {
     switch self {
     case .Console:
       print(content)
     case .File(let path):
       do {
+        if try path.isContentEqualTo(content) && onlyIfChanged {
```
AliSoftware: I think inverting the order of the test (`onlyIfChanged && try path.isContentEqualTo(content)`) would be more efficient, so it doesn't read the file first then don't care about the result if `onlyIfChanged` is `false`, but rather check the `onlyIfChanged` flag first and read the file to compare contents only if needed.

AliSoftware:
Great, thanks a lot for the PR! :+1: 

Don't forget to add an entry in the `CHANGELOG.md` to credit yourself :wink:

swiftgen-cli/ArgumentsUtils.swift
mathiasnagler: I guess you're right. The helper function probably is of litte value.

I will remove it.

swiftgen-cli/ArgumentsUtils.swift
mathiasnagler: Great suggestion, thanks!

swiftgen-cli/ArgumentsUtils.swift
```diff
@@ -62,6 +70,15 @@ enum OutputDestination: ArgumentConvertible {
   }
 }
 
+// MARK: Change comparsion
+
+extension Path {
+  func isContentEqualTo(string: String) throws -> Bool {
```
AliSoftware: Just nitpicking here, but I have to admit I feel like the name of this method sounds a bit Objective-C-y to me?
Open to suggestions as to what would be a better Swift name for it though.

swiftgen-cli/ArgumentsUtils.swift
mathiasnagler: Actually, this does not seem to be possible. When the order is inverted `onlyIfChanged && try path.isContentEqualTo(content)` there is a build error  `'try' cannot appear to the right of a non-assignment operator'. 

Is there a way of achieving this behaviour other than to use nested `if` statements?

swiftgen-cli/ArgumentsUtils.swift
mathiasnagler: I actually think this is fine. Using Objective-C the method name would probably be named `isContentEqualToString:`. As the `String` part is obvious in Swift, i dropped that part. 

Maybe dropping `is` from the name would make it even Swift-ier: `contentEqualTo:`.

swiftgen-cli/ArgumentsUtils.swift
AliSoftware: It seems like putting the try in front of the whole expression might work.

What about this:

```
if try onlyIfChanged && path.read(NSUTF8StringEncoding) == "Foo" {
  return print("Not writing the file as content is unchanged")
}
```

Given that the method name `isContentEqualTo` doesn't sound very Swifty to me anyway, this would solve both problems :wink:

mathiasnagler: Concise. I like it.

AliSoftware: Btw did you check what happens if the output file doesn't exist? I mean, I hope it doesn't throw in that case — not being able to read a non-existing file — and make the whole command fail the first time you launch it :wink:

Maybe worth adding a `path.exists` test in that condition or sthg?

mathiasnagler: How in the world did I miss that? 😳

It throws. I will update the PR.

PR #92, opened 2016-02-18
fix misplacement in playground files

bvonheid:
The header text was misplaced and some examples didn't work therefore.


AliSoftware:
Huh?! :astonished:  Must have happened after a previous merge gone wrong, strange…

Anyway, thanks a lot for the PR! :+1:

Issue #96, opened 2016-02-26
Avoid force {whatever}ing


toshi0383:
I thought it would be just safer to avoid force {whatever}ing.
Could you tell me reasons why you using ! mark everywhere ?
I'm thinking about some cases like when people accidentally forget to commit the Storyboard file.

AliSoftware:
I'm totally for removing force-casting everywhere and replacing them with `guard let foo as? Bar else { fatalError("some descriptive message") }` instead :+1: 

PR welcome, it should be very easy to fix the templates about this :smiley:

toshi0383:
Right, I'm on it ! 😀

PR #98, opened 2016-02-27
Avoid force casting/unwrapping

toshi0383:
Addresses #96

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,12 +101,12 @@ extension UIImage {
     case BigPear = "Big_Pear"
     case StopButtonEnabled = "stop.button.enabled"
 
-    var image: UIImage {
-      return UIImage(named: self.rawValue)!
+    var image: UIImage? {
+      return UIImage(named: self.rawValue)
     }
   }
 
-  convenience init!(asset: Asset) {
+  convenience init?(asset: Asset) {
     self.init(named: asset.rawValue)
   }
 }
@@ -115,8 +115,8 @@ extension UIImage {
 ### Usage Example
 
 ```swift
-let image1 = UIImage(asset: .Banana)   // Prefered way
-let image2 = UIImage.Asset.Apple.image // Alternate way
+let image1 = UIImage(asset: .Banana)!   // Prefered way
+let image2 = UIImage.Asset.Apple.image! // Alternate way
 ```
 
 This way, no need to enter the `"Banana"` string in your code and risk any typo.
@@ -226,8 +226,8 @@ extension StoryboardSceneType {
         return UIStoryboard(name: self.storyboardName, bundle: nil)
     }
 
-    static func initialViewController() -> UIViewController {
-        return storyboard().instantiateInitialViewController()!
+    static func initialViewController() -> UIViewController? {
+        return storyboard().instantiateInitialViewController()
     }
 }
 
@@ -258,16 +258,16 @@ struct StoryboardScene {
     }
 
     case URLChooser = "URLChooser"
-    static func urlChooserViewController() -> XXPickerViewController {
-      return Message.URLChooser.viewController() as! XXPickerViewController
+    static func urlChooserViewController() -> XXPickerViewController? {
+      return Message.URLChooser.viewController() as? XXPickerViewController
     }
   }
   enum Wizard : String, StoryboardSceneType {
     static let storyboardName = "Wizard"
 
     case CreateAccount = "CreateAccount"
-    static func createAccountViewController() -> CreateAccViewController {
-        return Wizard.CreateAccount.viewController() as! CreateAccViewController
+    static func createAccountViewController() -> CreateAccViewController? {
+        return Wizard.CreateAccount.viewController() as? CreateAccViewController
     }
 
     case ValidatePassword = "Validate_Password"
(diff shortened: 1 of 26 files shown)


README.md
```diff
@@ -101,12 +101,12 @@ extension UIImage {
...
 
-    var image: UIImage {
-      return UIImage(named: self.rawValue)!
+    var image: UIImage? {
+      return UIImage(named: self.rawValue)
     }
   }
 
-  convenience init!(asset: Asset) {
+  convenience init?(asset: Asset) {
```
AliSoftware: Oh, that's not what I was thinking about at all, sorry…

Given that the enums are auto-generated by SwiftGen each time you build (as you generally setup SwiftGen in a Script Build Phase), that is **very unlikely** that the init would fail here (and same for the computed property `image: UIImage`). So making them plain optionals would be a pain to use then because you'll have to unwrap them at call site, whereas you know that they should never be `nil` (because SwiftGen will ensure they are in sync with the assets catalog) and if they are, that's a programmer error (which should generate a `fatalError`).

So making this optional here just deports the unwrapping at call site, which would then be a pain to unwrap each time for nothing.

What I was thinking about was more along the lines of this:

```swift
convenience init(asset: Asset) {
  self.init(named: asset.rawValue)
  guard self != nil else {
    fatalError(
      "You used an Asset enum which doesn't match an existing Image. \n"
      + "Be sure to run SwiftGen again to ensure the enum and the Assets Catalog are in sync"
    )
  }
}
```

You can find some inspiration [on how I do this in my Reusable pod on those lines](https://github.com/AliSoftware/Reusable/blob/master/Sources/Reusable.swift#L63-L70).

AliSoftware:
Thanks for the PR, but again, sorry if I wasn't clear in my comment on #96 but as SwiftGen generates the enum each time you compile your project and make sure everything is always in sync, then:

1. It's very unlikely that the images, viewControllers, etc would be `nil` at runtime
2. If they are, that means that you forgot to regenerate the enums and that would mean a programmer error that you need to fix (as opposed to a failable situation where `nil` would be an acceptable value)
3. Forcing the users to unwrap at call site each time they want to use an image, viewController, etc… would be a pain to use at call site and defeat the purpose of this tool entirely, whereas the security of having always-existing resources is supposed to already be guaranteed by running SwiftGen in a Build Phase to make them always in sync.

So my idea in #96 was more to replace calls to `whatever!` with calls to `guard let whatever = whatever else { fatalError("Some comprehensive message to suggest the developer to re-run SwiftGen") }` and replace calls to `let foo = bar as! Foo` with `guard let foo = bar as? Foo else { fatalError("Some comprehensive message") }`. Like I did in Reusable linked in my comment above.

Hope you see what I mean and the rationale behind this and you could adjust the PR accordingly :wink:

toshi0383:
Understood. That's good way to handle the resource issues I guess.
I will close this PR, since fatalError and force casting makes no difference for me.

BTW, I'm not sure about this.
> the security of having always-existing resources is supposed to already be guaranteed by running SwiftGen in a Build Phase to make them always in sync.

IMO, SwiftGen **could be** updated unexpectedly including some breaking changes. There is no guarantees that SwiftGen always generates the correct code for me. So I rather not run SwiftGen in the Run Script Phase, but I run it when needed, test it to make sure it's working okay, merge it to master. I don't want any unexpected changes after I merged to master.

AliSoftware:
If you are afraid that a version of SwiftGen changes the templates and generate a different code than its last version, you can still copy the template to use next to your project and use the `--templatePath` option. This way, even if the templates bundled in SwiftGen change in a future version, you'll still using the templates that you would have committed and pushed on your project's repo, so no risk of breakage.

AliSoftware:
Btw, if you're willing to use optionals even in such cases (where resources should always be present and a nil value would mean a programmer error), you can totally use your own template (like the ones you have created for this PR) in your projects.

I can even totally add those "optional" templates of yours as alternative templates vendored with SwiftGen (like that's the case for `colors-rawValue` or `storyboards-lowercase` templates already). Users interested in those templates could thus simply use them on demand with the `-t optionals` flag when invoking SwiftGen.

Simply create a new PR by adding those templates as separate files in the `template/` directory (and ideally add some test cases for them) and I'll accept them :wink:

toshi0383:
Hi, thanks for following up !
I will do it when I have time! 👍

README.md
lgauthier: Unfortunately, I'm not sure if there's actually a way to make use of the fatalError here. The example you have won't compile because a non-failable initializer can't delegate to a failable initializer. 

One option would be to make the asset initializer failable and use the fatalError in the image property on the Asset enum:

```
extension UIImage {
    enum Asset: String {
        case Logo = "logo"
        
        var image: UIImage {
            guard let image = UIImage(asset: self) else {
                fatalError(
                    "You used an Asset enum which doesn't match an existing Image. \n"
                        + "Be sure to run SwiftGen again to ensure the enum and the Assets Catalog are in sync"
                )
            }
            return image
        }
    }
    
    convenience init?(asset: Asset) {
        self.init(named: asset.rawValue)
    }
}
```

Of course, with this route, you'd still be stuck having to use an optional UIImage if you call this initializer directly.

Do you know of any other way you could make this work?

Issue #51, opened 2015-11-04
UIFont support


joseprl89:
It would be great if we could have something similar to the UIColor functionality, but for UIFont.

AliSoftware:
I think it would be even better if the command could parse a directory to look for `.ttf` and `.otf` fonts, then automatically detect their postscript name (probably using `CTFontManagerCreateFontDescriptorsFromURL` from `swiftgen-cli` to load the font and analyse it).

joseprl89:
@AliSoftware makes sense, but in some cases you don't provide the ttf files in the bundle since they are already supported by iOS (e.g. Helvetica Neue and so on). The use case I had in mind was setting something like:

body-text : Helvetica Neue ; 17 ; Bold

Or a similar format. Would that make sense?

AliSoftware:
Ah, good point, that actually makes total sense (and would be way easier to do, BTW, you could even make a PR for that, using the `ColorsParser` as inspiration :wink:). Will tag as "easy" then.

joseprl89:
Created a pull request for this. The tests at the moment are failing due to encoding (unless I'm missing something).

I'm afraid I won't be able to spend more time on it for a few days, so if someone can pick it up  it would be much appreciated.

I'll paste the pull request, because I assume travis CI will reject it? 

https://github.com/AliSoftware/SwiftGen/pull/53

bill350:
I think the font approach can be more simple.
Here how we construct font extension without specific sizes :
 
```
extension UIFont {
    enum AppFont : String {
        case Regular    = "MarkPro"
        case Heavy      = "MarkPro-Heavy"
        case Light      = "MarkPro-Light"
        case ExtraLight = "MarkPro-ExtraLight"
        case Medium     = "MarkPro-Medium"
        case Bold       = "MarkPro-Bold"
    }
    
    convenience init(named name: AppFont, size:CGFloat) {
        self.init(name:name.rawValue, size:size)!
    }
}
```

So, with a `Fonts.txt` file and a structure like `Colors.txt`, it could be simple to generate extension based on the example above. 

File structure :
```
Regular = MarkPro
Light = MarkPro-Light
```

AliSoftware:
I think it really depends on how people use their fonts.

* Some people use constants only for font names/families without size, like you do @bill350 , and so you'll provide the size at call site and use `UIFont(named: Regular, size: 16)` because at this specific point in code you know you want to use this font at this size
[...]

PR #102, opened 2016-03-11
Fonts

dostrander:
We can now generate fonts!

#51


CHANGELOG.md
```diff
@@ -3,7 +3,8 @@
 ---
 
 ## Master
-
+* Added support for Fonts using the `swiftgen fonts` command.  
+  [Derek Ostrander](https://github.com/dostrander)
```
AliSoftware: Could you add a link to this PR next to the link to your name please? Thx!

GenumKit/Parsers/FontsFileParser.swift
```diff
@@ -0,0 +1,108 @@
...
+            where CTFontManagerRegisterGraphicsFont(font, &error) else {
+                return nil
+        }
+        return CTFontCreateWithGraphicsFont(font, 0, nil, nil).font
+    }
+
+    private var font: Font? {
+        let family = CTFontCopyFamilyName(self).alphanumericString
+        let name = CTFontCopyFullName(self).alphanumericString.stringByReplacingOccurrencesOfString(family, withString: "")
+        if family.characters.count == 0 {
```
AliSoftware: prefer `family.isEmpty` over `family.characters.count`

GenumKit/Parsers/FontsFileParser.swift
```diff
@@ -0,0 +1,108 @@
...
+            print("There was a problem detecting the name of the font. \(CTFontCopyFamilyName(self)) -- \(CTFontCopyFullName(self)) ")
+            return nil
+        }
+        return (family, name.characters.count > 0 ? name : "Regular")
```
AliSoftware: same here

GenumKit/Parsers/FontsFileParser.swift
```diff
@@ -0,0 +1,108 @@
...
+            let fullPath = (path as NSString).stringByAppendingPathComponent(file)
+            if isDirectory(fullPath) {
+                parseFonts(fullPath)
+            } else if let font = CTFont.loadFont(fullPath) {
+                addVariation(font.name, variation: font.variant)
+            }
+        }
+    }
+
+    public func printFonts() {
```
AliSoftware: That functions was probably very useful for debug, but I'm not sure we need to keep it now that the code is working?

GenumKit/Parsers/FontsFileParser.swift
```diff
@@ -0,0 +1,108 @@
...
+//
+//
+
+import Foundation
+import AppKit.NSFont
+
+typealias Font = (name: String, variant: String)
+
+extension CFString {
+    var alphanumericString: String {
```
AliSoftware: I'm not sure why filtering the font families to only keep alphanumeric characters is necessary?

In Swift, a lot of stuff can be used as Swift identifies, even emojis or Japanese characters car be case identifiers.
Besides, that's the job of the template to filter and transform those strings (because people creating templates might want to use the full name of the font to generate a comment before the `case` for exemple), and that's what the Stencil filter `swiftIdentifier`will be used for in the Stencil template :wink:

GenumKit/Parsers/FontsFileParser.swift
```diff
@@ -0,0 +1,108 @@
...
+            print(str)
+        }
+    }
+}
+
+// MARK: Entry Managment
+// needs to be testable
+extension FontsFileParser {
+    func addVariation(name: String, variation: String) {
+        if var variations = entries[name] {
```
AliSoftware: In future versions of Swift I think `if var` will be removed. Maybe we could use `if let` there instead to be prepared (and use `entries[name] = variations + [variation]` below instead)?

GenumKit/Parsers/FontsFileParser.swift
```diff
@@ -0,0 +1,108 @@
...
+        print("There was a problem grabbing the contents of the directory -- \(err)")
+    }
+
+    return files
+}
+
+private func isDirectory(path: String) -> Bool {
+    var isDir: ObjCBool = false
+    return NSFileManager.defaultManager().fileExistsAtPath(path, isDirectory: &isDir) && isDir
+}
```
AliSoftware: add CRLF at end of file

GenumKit/Parsers/FontsFileParser.swift
```diff
@@ -0,0 +1,108 @@
...
+        else {
+            entries[name] = [variation]
+        }
+    }
+}
+
+// MARK: Helpers
+
+private func contentsOfDirectory(path: String) -> [String] {
+    var files: [String] = []
```
AliSoftware: `var` -> `let`

GenumKit/Parsers/FontsFileParser.swift
```diff
@@ -0,0 +1,108 @@
...
+
+// MARK: Interface
+public final class FontsFileParser {
+    public var entries: [String: Set<String>] = [:]
+
+    public init() {}
+
+    public func parseFonts(path: String) {
+        for file in contentsOfDirectory(path) {
+            let fullPath = (path as NSString).stringByAppendingPathComponent(file)
```
AliSoftware: I think you might be able to use `NSFileManager.defaultManager().enumeratorAtPath(path)` here instead to automatically iterate recursively over all the files  the directory, like I did in [AssetsCatalogParser.swift](https://github.com/AliSoftware/SwiftGen/blob/master/GenumKit/Parsers/AssetsCatalogParser.swift#L24)

AliSoftware:
Awesome! Thanks a lot for this Pull Request! :ok_hand: : 

I made some small feedbacks to improve the code, but looks very promising!

Here are some additional notes/suggestions I'd love your feedback on:

#### Add a `UIFont` initializer

What do you think about adding another way to create the fonts?
Tbh, even if `UIFont.Family.HelveticaNeue.Bold.font(12.0)` is OK, I find the code line kinda long, so maybe we could add an alternative to the template, so we could also create the same font using `UIFont(variant: .HelveticaNeue.Bold, size: 12)`?

This would also match the other templates like the one for images or colors, where images can be created either by `UIImage.Asset.Foo.image` or via `UIImage(asset: .Foo)`

#### 2-space indentation

I know it's a matter of preferences, but the whole code in SwiftGen uses two spaces indentation, so it would be nice to keep it consistent across the whole code. Would you be kind enough to set your files's indentation (setting per project or per file in the File Inspector on the right) to 2 then `Ctrl-I` on your code? Thx!

#### Unit Tests with OTF/TTF files

I think it would be a good idea to add some more unit tests, parsing OTF & TTF font files, to ensure that your code using `CTFont` works as expected.
Those unit tests might not even need to generate the template and compare to an expected template like the other tests, they can just checking that the parsed `Font` tuples match your expectations.

Using some tricky/unusual font files, like fonts which don't contain "Regular" in their names for the regular variant, or font files for which the file name does not match the postscript font name stored in the TTF/OTF file itself, or fonts with unusual names like containing kanjis or non-alphanumeric chars…


Thanks again for all the work :+1:

GenumKit/Parsers/FontsFileParser.swift
dostrander: This will recursively go over through the directory? I had no idea

AliSoftware: Yeah that's pretty powerful.

I just realized that, as you'll then need to filter the enumerated files to only process font files (and skip any file in the scanned directory that might be present but not be a font file, in case the user scan for example the whole `$PROJECT_DIR`), that you might even use this method of `NSFileManager` instead which is even more powerful:

```
let dirEnum = NSFileManager.defaultManager().enumeratorAtURL(
  NSURL(fileURLWithPath: path),
  includingPropertiesForKeys:[NSURLTypeIdentifierKey],
  options: [.SkipsPackageDescendants, .SkipsHiddenFiles]
  errorHandler: nil)
```

This will enumerate the files recursively and I guess you should then be able to check `dirEnum.fileAttributes[NSURLTypeIdentifierKey]` and compare it with `public.font` to only process font files?

AliSoftware: Here's what I came up to in my tests to parse a directory, get the Universal Type Identifier of each file to only process files with an UTI conforming to "public.font", then parse only those files:

```
import Foundation
import MobileCoreServices
…
guard let dirEnum = NSFileManager.defaultManager().enumeratorAtURL(
  NSURL(fileURLWithPath: path),
  includingPropertiesForKeys:[NSURLTypeIdentifierKey],
  options: [.SkipsPackageDescendants, .SkipsHiddenFiles],
  errorHandler: nil) else {
    fatalError("Failed to enumerate directory \(path)")
}

var utiObject: AnyObject? = nil
while let file = dirEnum.nextObject() as? NSURL {
  // Retrieve UTI of the file
  guard let _ = try? file.getResourceValue(&utiObject, forKey: NSURLTypeIdentifierKey),
    let uti = utiObject as? String else {
      print("Unable to determine the Universal Type Identifier for file \(file)")
      continue
  }
  // Skip non-font files
  guard UTTypeConformsTo(uti, "public.font") else { continue }
  // Parse font file
  guard let font = CTFont.loadFont(fullPath) else { continue }
  addVariation(font.name, variation: font.variant)
}
```

HTH
---

PS: Once you've migrated to that solution, you will even be able to remove your private `contentsOfDirectory` and `isDirectory` functions below :+1:

GenumKit/Parsers/FontsFileParser.swift
```diff
@@ -0,0 +1,108 @@
...
+         return (self as String).componentsSeparatedByCharactersInSet(NSCharacterSet.alphanumericCharacterSet().invertedSet).joinWithSeparator("")
+    }
+}
+
+extension CTFont {
+    static func loadFont(path: String) -> Font? {
+        let inData = NSData(contentsOfFile: path)
+        var error: Unmanaged<CFError>?
+        guard
+            let provider: CGDataProvider = CGDataProviderCreateWithCFData(inData),
```
AliSoftware: It's more efficient to use `CGDataProviderCreateWithURL` than read the whole `NSData(contentsOfFile:)` and use `CGDataPrividerCreateWithCFData`. Because `CGDataProviderCreateWithURL` will only reference the URL then only read the amount of data needed by further operations, avoiding reading the while file data if not needed.

Besides, I'm wondering if all that code is necessary. It seems quite CPU intensive to read the whole font data, create a CGFont from it, register it with the Font Manager (whereas we don't intent to use the font in a graphic context, only to extract its name and variations), then create a `CTFont` from there only to extract the metadata. I'm sure there are less intensive ways to do that?

GenumKit/Parsers/FontsFileParser.swift
dostrander: Ok moving over to this, curious if you know why "public.font" ? I can't find it documented anywhere but if I do a quick search i can find uses of it?

AliSoftware: TBH, I used `mdfind /Library/Fonts/Arial.ttf` to display the metadata of a random font file which gave me:
```
kMDItemContentTypeTree         = (
    "public.truetype-ttf-font",
    "public.truetype-font",
    "public.font",
    "public.data",
    "public.item"
)
```
So that's how I saw the UTI tree of the font file and guessed that the generic UTI for any font file is `public.font`

GenumKit/Parsers/FontsFileParser.swift
AliSoftware: Ok, I came up with a way better way to load fonts:

* It support more type of fonts (I tested by iterating over all the fonts in `/Library/Fonts` and for some of them, creating the intermediate `CGDataProvider` or the `CGFont` fail, whereas with this new code it works all the time)
* It extracts the font variant using the official name, instead of trying to guess it from the full name

Here's the code I used (tested in playground):

```swift
extension CTFont {
  static func loadFont(fileURL: NSURL) -> Font? {
    let descs = CTFontManagerCreateFontDescriptorsFromURL(fileURL) as NSArray?
    guard let desc = descs?.firstObject else { return nil }
    let font = CTFontCreateWithFontDescriptorAndOptions(desc as! CTFontDescriptor, 0.0, nil, [.PreventAutoActivation])
    guard let family = CTFontCopyAttribute(font, kCTFontFamilyNameAttribute) as? String else { return nil }
    guard let variant = CTFontCopyAttribute(font, kCTFontStyleNameAttribute) as? String else { return nil }
    return Font(name: family, variant: variant)
  }
}
```

Then no need to build a data provider, build an intermediate `CGFont` register the `CTFont` or whatnot, and you can now get rid of the `alphanumericString`, `private var font: Font?` helpers and stop guessing the font variant!

GenumKit/Stencil/Contexts.swift
```diff
@@ -131,3 +131,30 @@ extension StringsFileParser {
...
+            - `value`: `String`
+*/
+
+extension FontsFileParser {
+    public func stencilContext(enumName enumName: String = "Family") -> Context{
+        // turn into array of dictionaries
+        let fontFamilies = entries.map { (name: String, variants: Set<String>) -> [String:AnyObject] in
+            let fontVariants = variants.map { return [
+                "name" : $0,
+                "value" : ($0 == "Regular" ? name : "\(name)-\($0)")
```
AliSoftware: Be careful, to construct a `UIFont` you must pass the font **postscript name** (which is not always `Family-Variant`!)
For example for `/Library/Fonts/Hannotate.ttc`  the family name is `"Hannotate SC"`, the style/variant name is `"Regular"` but the PostScript name is `"HannotateSC-W5"`! (yeah, fonts are hard)

I suggest you add a `postscriptName` entry to your `Font` tuple, fill it with the value or `CTFontCopyPostScriptName(font) as String` when parsing the font file in the function `CTFont.parseFontInfo` I gave in the comment above, and use that to fill the StencilContext instead.

```
fonts: Array
 - family: String   // Like "Arial"
 - variants: Array
    - style: String  // Like "Bold Italic"
    - name: String // Postscript name to use to build UIFont, like "Arial-BoldItalicMT"
```

GenumKit/Parsers/FontsFileParser.swift
dostrander: :+1: SO AWESOME, thanks! This is a great improvement and set's us up more for doing non-custom fonts, great find.

GenumKit/Parsers/FontsFileParser.swift
```diff
@@ -0,0 +1,102 @@
...
+}
+
+public func ==(lhs: Font, rhs: Font) -> Bool {
+  return lhs.postScriptName == rhs.postScriptName
+}
+
+// MARK: CFString
+
+extension CFString {
+  var alphanumericString: String {
```
AliSoftware: This is not needed anymore!

GenumKit/Parsers/FontsFileParser.swift
```diff
@@ -0,0 +1,102 @@
...
+
+          addFont(font)
+        }
+    }
+  }
+}
+
+// MARK: FontsFileParser - Entry Managment
+
+private extension FontsFileParser {
```
AliSoftware: Not sure it's worth having that in a dedicated extension

I assume that's a coding convention of yours to group all private functions in an extension to clearly separate them from the public API? (if so that's a nice idea but might not b worth it there esp given that it's not enforced on the rest of the code)

GenumKit/Parsers/FontsFileParser.swift
```diff
@@ -0,0 +1,102 @@
...
+  func addFont(font: Font) {
+    let familyName = font.familyName
+    if let _ = entries[familyName]
+    {
+      entries[familyName]!.insert(font)
```
AliSoftware: Please [don't kill ponies](http://alisoftware.github.io/swift/2015/09/06/thinking-in-swift-1/). :hocho: :horse: :x:

dostrander: :sweat_smile:

AliSoftware: What about this?
```
func addFont(font: Font) {
  let familyName = font.familyName
  var variants = entries[familyName] ?? []
  variants.insert(font)
  entries[familyName] = variants
}
```

GenumKit/Parsers/FontsFileParser.swift
```diff
@@ -0,0 +1,89 @@
...
+
+          addFont(font)
+        }
+    }
+  }
+
+  private func addFont(font: Font) {
+    let familyName = font.familyName
+    if let styles = entries[familyName] {
+      entries[familyName] = styles.union([font])
```
AliSoftware: Ah, even better! Perfect

GenumKit/Stencil/Contexts.swift
```diff
@@ -131,3 +131,35 @@ extension StringsFileParser {
...
   }
 }
+
+/* MARK: - Stencil Context for Fonts
+- `enumName`: `String`
+- `families`: `Array`
+  - `name`: `String`
+  - `fonts`: `Array`
+    - `name`: `String`
+    - `value`: `String`
```
AliSoftware: Feels strange that the `families.fonts.name` is not the font name but just the style, while `families.fonts.value` is not very explicit.

I suggest instead:
```
 - `enumName`: `String`
 - `families`: `Array`
   - `name`: `String` // Like "Arial" 
   - `variants`: `Array`
     - `style`: `String` // Like "Bold Italic"
     - `fontName`: `String` // Postscript name to use with UIFont(name:size:), like "Arial-BoldItalicMT"
```
Thoughts?

AliSoftware: Another alternative suggestion:
```
 - `enumName`: `String`
 - `families`: `Array`
   - `name`: `String`        // Like "Arial" 
   - `fonts`: `Array`
     - `style`: `String`     // Like "Bold Italic"
     - `name`: `String`  // Postscript name to use with UIFont(name:size:), like "Arial-BoldItalicMT"
```

GenumKit/Parsers/FontsFileParser.swift
```diff
@@ -0,0 +1,89 @@
+//
+//  FontsFileParser.swift
+//  Pods
```
AliSoftware: Hey, it's the professional nitpicker again! :stuck_out_tongue_winking_eye: 

Please make the file headers for your new files conform to the other ones, especially mention the MIT License, so something like:

```
//
// GenumKit
// File created by Derek Ostrander on 3/7/16.
// Copyright (c) 2015 Olivier Halligon
// MIT Licence
//
```

README.md
```diff
@@ -54,6 +55,7 @@ The tool is provided as a unique `swiftgen` binary command-line, with the follow
 * `swiftgen strings [OPTIONS] FILE`
 * `swiftgen storyboards [OPTIONS] DIR`
 * `swiftgen colors [OPTIONS] FILE`
+* `swiftgen colors [OPTIONS] DIR`
```
AliSoftware:
```diff
- swiftgen colors [OPTIONS] DIR
+ swiftgen fonts [OPTIONS] DIR
```

templates/fonts-default.stencil
```diff
@@ -0,0 +1,29 @@
+// Generated using SwiftGen, by O.Halligon — https://github.com/AliSoftware/SwiftGen
+
+{% if families %}
+import UIKit.UIFont
+
+extension UIFont {
+{% for family in families %}
+    convinience init?(font: {{enumName}}.{{family.name|swiftIdentifier|snakeToCamelCaseNoPrefix}}, size: CGFloat) {
```
AliSoftware: `convinience` -> `convenience`

Also I think you might instead leverage `protocol` with default implementations (aka [Mixins](http://alisoftware.github.io/swift/protocol/2015/11/08/mixins-over-inheritance/)) here.
You can get some inspiration from what I did in the `storyboards-default.stencil` template. This would probably give something like this:

```swift
protocol FontConvertible {
  func font(size: CGFloat) -> UIFont!
}

extension FontConvertible where Self: RawRepresentable, Self.RawValue == String {
  func font(size: CGFLoat) -> UIFont! {
    return UIFont(name: self.rawValue, size: size)
  }
}

extension UIFont {
  convenience init!(font: FontConvertible, size: CGFloat) {
    let font = font.font(size)
    self.init(name: font.fontName, size: size)
  }
}
```

This way, you can avoid that loop to create as many convenience init as font families in `UIFont`, and also avoid redeclaring the `func font` for each enum: simply mark them to be conform to `FontConvertible` and they will get that implementation for free.

---

PS: Given that the enums are auto-generated by SwiftGen, I think that returning a `UIFont!` here instead of an `UIFont?` is OK there, as if the font can't be created from the enum, that can only be a developer error — probably because the developer forgot to re-run SwiftGen before compiling.

dostrander:
@AliSoftware Will get to these comments early this week

GenumKit/Parsers/FontsFileParser.swift
```diff
@@ -0,0 +1,85 @@
...
+}
+
+public func ==(lhs: Font, rhs: Font) -> Bool {
+  return lhs.postScriptName == rhs.postScriptName
+}
+
+// MARK: CTFont
+
+extension CTFont {
+  static func parseFontInfo(fileURL: NSURL) -> Font? {
```
AliSoftware: should be `private` maybe?

GenumKit/Parsers/FontsFileParser.swift
```diff
@@ -0,0 +1,85 @@
...
+  let familyName: String
+  let style: String
+  let postScriptName: String
+
+  init(familyName: String, style: String, postScriptName: String) {
+    self.familyName = familyName
+    self.style = style
+    self.postScriptName = postScriptName
+  }
+}
```
samirGuerdah: Structure types automatically receive a memberwise initializer like the one you defined) if you don't provide any custom initializer.  You can remove the init.

AFcgi:
Hey, very nice idea. It's possible to define and import the fonts from a file like colors? From my side it's enough to support only one file format (json or xml).

AliSoftware:
Hey @dostrander!

Finally had time to merge your Pull Request 🎉 
Very sorry it took so much time, it has been crazy last months.

---

Following the same principles as [Moya Contribution Guidelines](https://github.com/Moya/contributors) and to thank your for your great contribution, I just added you as contributor to SwiftGen.
This doesn't necessarily mean any obligation from your part (doesn't mean any special expectation), but at least if you're interested you can now contribute more directly: master is locked to enforce Travis to run tests on PRs before merging, but you can merge your own PRs if you want and close or reopen issues, etc!

Thanks again for your contribution!

AliSoftware:
@AFcgi this PR was aimed at parsing TTF/OTF/… font files from a directory so the import from a text file was not part of this PR.

But #53 started to address that (well way before this very PR) so now that this PR is merged it might be the perfect time to refactor and finish #53 to address this!

PR #108, opened 2016-03-29
Add Nib names generator command

ludov04:
Hello.

Your tool is great, I added a command to generate nib name automatically.
Would you mind to merge it ?

Thanks, 
Ludovic

### DIFF
--- a/UnitTests/TestSuites/NibsTests.swift
+++ b/UnitTests/TestSuites/NibsTests.swift
@@ -1,0 +1,24 @@
+//
+// SwiftGen
+// Copyright (c) 2015 Olivier Halligon
+// MIT Licence
+//
+
+import XCTest
+import GenumKit
+
+
+class NibssTests: XCTestCase {
+  
+  func testFileWithDefaults() {
+    let parser = XibParser()
+    parser.parseDirectory(fixturesDir)
+    
+    let template = GenumTemplate(templateString: fixtureString("nibs-default.stencil"))
+    let result = try! template.render(parser.stencilContext())
+    
+    let expected = self.fixtureString("Nibs.swift.out")
+    XCTDiffStrings(result, expected)
+  }
+  
+}
--- a/GenumKit/Parsers/XibParser.swift
+++ b/GenumKit/Parsers/XibParser.swift
@@ -1,0 +1,33 @@
+//
+// GenumKit
+// Copyright (c) 2015 Olivier Halligon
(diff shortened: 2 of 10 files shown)


UnitTests/TestSuites/NibsTests.swift
```diff
@@ -0,0 +1,24 @@
...
+// SwiftGen
+// Copyright (c) 2015 Olivier Halligon
+// MIT Licence
+//
+
+import XCTest
+import GenumKit
+
+
+class NibssTests: XCTestCase {
```
AliSoftware: Probably too much `s` there ^^

AliSoftware:
This looks nice, thanks! :+1: 

But given all your parser does is listing files with a given extension (here `.nib`) without any specific treatment of the file content itself (so it doesn't depend on the file format at all), maybe we could extend that feature to parse any file extension?

e.g. The idea would be to extension your idea to let people specify what extension they want to list. For example, let's say we call that swiftgen subcommand `filelist` then you could achieve the list of NIB files by invoking `swiftgen filelist nib "$PROJECT_DIR"` but one could also list all the `*.json` files they have in their bundle, or every `*.plist` file they might use as resources, e.g. using `swiftgen filelist plist "$PROJECT_DIR/Resources/DataSets/` etc.

What do you think about the idea? Do you think you can adapt your PR to generalize to this?

Thx!

AliSoftware:
@ludov04 any news on this? What do you think about making it generic to work on any file extension?

Note that since some other PRs have been merged since, you'll need to rebase your PR to fix the conflicts (mainly on the CHANGELOG file, so not too hard) so it can be merged

smadbe:
Ludov04 is not part of the company for which he has done this fork anymore, so I don't think you should count on him to continue the integration work.

smadbe:
For info, I've merged with your changes and adapted to Swift3 in https://github.com/Keemotion/SwiftGen/

djbe:
Hi @valerianb,

Another user started a newer PR #134 that superseded the functionality in this one. Due to the refactoring (great split) in 4.2.0, that PR is no longer useable. Another user has started a new PR which you can find here:
https://github.com/SwiftGen/SwiftGenKit/pull/15

That last PR is a start, but still needs quite a bit of work to match the functionality in #134. You're welcome helping out, but for now this isn't a priority (for us maintainers) while we're still cleaning up the code base and documenting stuff.

PR #114, opened 2016-05-09
Show translation as documentation

patoroco:
Hi,

this small patch changes documentation for strings, allowing to do `Opt + Click` in any translation and show there the current translation as you can see in the image below.

screen shot 2016-05-09 at 11 00 28

This is useful because sometimes keys aren't enough explanatory.

**PD:** I'm not sure if this should be translated to other templates (storyboards, images...), so I'll wait to your feeling about that to do too.

### DIFF
--- a/UnitTests/expected/Strings-Entries-Defaults.swift.out
+++ b/UnitTests/expected/Strings-Entries-Defaults.swift.out
@@ -3,9 +3,13 @@
 import Foundation
 
 enum L10n {
-  /// My awesome title
+  /**
+    My awesome title
+  */
   case Title
-  /// Hello, my name is %@ and I'm %d
+  /**
+    Hello, my name is %@ and I'm %d
+  */
   case Greetings(String, Int)
 }
 
--- a/UnitTests/expected/Strings-File-CustomName.swift.out
+++ b/UnitTests/expected/Strings-File-CustomName.swift.out
@@ -3,17 +3,29 @@
 import Foundation
 
 enum XCTLoc {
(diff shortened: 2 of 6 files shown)


AliSoftware:
Thanks for the PR !

Didn't the Option-clic documentation work already with comments marked as `///` in the current template? I also use this `///` syntax for the colors template to display the actual color in the opt-clic doc, so I'm little surprised if it doesn't work in your Xcode, because that's exactly why I added those type of comments (in the colors and strings templates) in the first place :wink:

Note that I preferred `///` over `/** … */` because most of the time localized strings fit in a single line, so it makes the generated source code smaller to use a one-line `///` than use 3 lines with the `/** … */` syntax. But if that means that the opt-clic documentation doesn't work anymore in latest Xcode or whatnot…?

Also, if we migrate to `/** … */` instead of `///`, then I think it might be worth checking the behavior in case the translation string itself contains some `*/` which would break the generated code (which is also why `///` is simpler to generate)

patoroco:
You're right, I don't understand what happened, but Xcode didn't autocomplete two hours again.

I just try it again, and it now autocompletes!

So, sorry for the stupid PR, and I'll close it ;)

AliSoftware:
No problem, thanks for taking the effort of submitting it anyway 😉

AliSoftware:
Also note that my current code might have a similar issue to what I pointed out in yours: if the translation contains `\n` in the string itself, I guess my implementation that might also break the generated code. But that's another issue :wink:

PR #122, opened 2016-06-01
Explicates semi-colon requirement

AmitaiB:
This drove me nuts for hours, until I got lucky. Paying it forward.


AliSoftware:
Thanks, for the addition, could indeed help others loosing hours on this kind of stuff :wink:

But to be fair, that is not a requirement specific of SwiftGen parser or whatnot. That's the syntax of the `.strings` format. So even without involving SwiftGen (even in projects where you don't use SwiftGen at all), if you forget the semi-colon at the end of your lines in `Localizable.strings`, then Xcode won't even compile your `Localizable.strings` file and would generate a build error. Because that's the syntax of that Xcode file format (and not a deliberate choice I made when I created SwiftGen or whatnot).

So I'm OK adding a note to remind people not to forget semicolons, but that should be explained differently:

* "Endline semi-colons in the source file" is misleading. We are not talking about "source files" here (I mean, not in the sense of files containing source _code_), but we are talking about `*.strings` Xcode resource files
* "… are **required** by the parser" is also misleading, as it's in fact primary required by Xcode itself, which imposes that as part of this file format. the parser is only matching the expectations Xcode does set for this.

So we should probably write that note differently, or suggesting people that they build their project (so that Xcode builds the `Localizable.strings` file) first to ensure their `Localizable.strings` file is valid if they see some lines not parsed by SwiftGen?

PS : I'm curious about how you managed forget a semi-colon and _not_ having Xcode generating an error to explain where you missed that semicolon? As Xcode should have helped you avoid loosing so much time in the first place already, so it's strange if the compiler didn't complain with an explicit message?

AliSoftware:
Note added after a little re-wording in 21ef0a643a3193e1d6edb1eae7108c4f55b46397

AmitaiB:
Firstly, thank you for responding in such detail! I am a career-changer, and in my haste to get on my feet, sometimes (ok, let's be honest -- quite often) that my growth is uneven, and I find that I am lacking in my fundamentals. The cure for ignorance is knowledge, hence my gratitude. 👍 

Specifically, I didn't realize that *.strings was a resource file, having never encountered them before. I had assumed that `Localizable.strings` was a self-documenting-ly named variation of `example.file`. To save Past Me from that error -- on the stipulation that the Universe revolves around me -- hyperlinking the extension in `Localizable.`[`strings`](https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/LoadingResources/Strings/Strings.html) would do it.

Xcode wasn't catching it at first, since I was testing it in Terminal, but I don't know why it never gave me an error. Just now, I created a new strings file, and added Swifty, semi-colonless lines and all of a sudden Xcode gives me helpful and descriptive errors.

Thanks again!

Issue #123, opened 2016-06-01
💡 Improve templates command


AliSoftware:
It would be useful to add some subcommands like these:

* `swiftgen templates cat xxx` to print the template to stdout
* `swiftgen templates cp xxx filepath` to copy the template to a new file
* `swiftgen templates which xxx` to print the path to the template matching this name

All those command would make creating new templates from existing ones easier, as one can use `cp` or `cat` to create a copy of an existing template in order to modify it.

We could even imagine a `swiftgen templates install filepath` to copy the templates pointed by `filepath` to the appropriate folder in Application Support to make it easy to install a custom template on a given machine.

AliSoftware:
Additional idea: add a `swiftgen templates desc xxx` command to print the description of template xxx.
That description could be extracted from the template itself, for example considering that if the template starts with a Stencil comment `{# … #}` then that comment is considered as the template description for example?

AliSoftware:
Closed by c953d75e17ef1f6993365755d7c8351fabfdb78a 😉

PR #117, opened 2016-05-18
Add Swift3 Templates 

AndrewSB:
related to #116

Issue #116, opened 2016-05-18
Swift 3 support


AndrewSB:
Whats the timeline for this library to support Swift 3?
Thoughts on adding a `--swift3` flag to the commands to generate Swift 3 compatible code?

AliSoftware:
There is no need for a `--swift3` support as templates already provides this ability and you can already customize the generated code and generate whatever code you want.

For example for colors create a `colors-swift3.stencil` template and use `swiftgen colors -t swift3` and it will work.

See the documentation about creating your custom templates in the README and in the doc/ folder.

I'm closing this issue as I don't plan to bundle Swift 3 templates with SwiftGen until Swift 3 itself is stable and the syntax for Swift 3 is frozen, but will do it late '16 at the same time Swift 3 is officially frozen and out.

But feel free to make a PR I  the meantime to share your `xxx-swift3.stencil` templates if you create ones for your own use case and I'll gladly include them with the other templates already provided with SwiftGen (see the templates/ directory in the repo)

AndrewSB:
in the process of making a PR, you're right there aren't many (if any) changes required for swiftgen, moreso changes in the dependencies that need to be integrated.

Thank you for explaining how to make the template, hoping to have that to you later today 😄

AliSoftware:
:+1:

AliSoftware:
@AndrewSB as @akaralar pointed out in https://github.com/AliSoftware/SwiftGen/pull/124#issuecomment-223342448 we might want to consider doing another PR to slightly adapt the Swift 3 templates you submitted, and make `enum cases` *lowercase*, as it will be the new Swift 3 conventions!

Sajjon:
Hi! Thanks @AliSoftware  for amazing SwiftGen! Best thing that happened Swift since... the release of Swift! 👍  

I have recently converted my whole project - which heavily makes use of SwiftGen - to Swift 3.

No I get compilation error 

> CVarArgType has been renamed CVarArg
For the method 

```
private static func tr(key: String, _ args: CVarArgType...) -> String {
    let format = NSLocalizedString(key, comment: "")
    return String(format: format, locale: NSLocale.currentLocale(), arguments: args)
  }
```

And I also think we should change `key: String` to `_ key: String` for both the method above and this below:
```
func tr(key: L10n) -> String {
  return key.string
}
```

@AndrewSB did you get it working with Swift 3?

AliSoftware:
[...]

PR #124, opened 2016-06-02
Add strings template for compatibility with genstrings workflow

akaralar:
Hey,

Thanks for this great tool and support for the fonts! I added a strings template to support a particular workflow regarding strings.

`genstrings` tool extracts all strings in `NSLocalizedString` calls into a `Localizable.strings` file.
With the current templates, the created strings file doesn't list `NSLocalizedString`s for each case, therefore we can not use `genstrings` for `Localizable.strings` creation, hence we need to update it manually ourselves each time we add a new string to the app.

With this template, the default template is modified only slightly so that each case makes a call to `NSLocalizedString`. 
Therefore we can work like this: 

- When coding, enter your new string as a call to `NSLocalizedString` ie. `self.titleLabel.text = NSLocalizedString("new string")`
- run `genstrings`
- run `swiftgen`
- replace the `NSLocalizedString("new string")` with the newly created string enum case.


AliSoftware:
Nice idea, but I'm just wondering, couldn't you just use `genstrings -s tr` for example?

> **-s** _routine_
>  Substitutes routine for NSLocalizedString.  For example, -s MyLocal-
>  String will catch calls to MyLocalString and MyLocalStringFromTable.

I haven't tested it as I don't use genstrings personally — especially because I feel it is not suited for adding new strings after the first generation as you point out — so maybe I'm wrong, but according to the man page at least this `-s` flag allows you to specify a different macro/function to parse for other than `NSLocalizedString`, especially for those use cases, so I think it would probably work here?

akaralar:
I didn't know about the `-s` flag, nice piece of info. However in the case of strings, the global `tr` function doesn't take a string as a parameter (hence `genstrings` complains that it's not a string literal.) the `tr` function inside the enum is private, as it should be, so I can not say `titleLabel.text = L10n.tr("new string")`

I find `genstrings` valuable as it automatically puts the strings into alphabetical order and removes the entries in `Localizable.strings` file if the string is no longer present in the codebase. The appealing part of the workflow I mentioned for me is that you don't need to navigate anywhere else, just type your string in using `NSLocalizedString` and run a script or build a scheme and your strings are ready to use, without leaving the same spot on the codebase :)

If you feel that this template doesn't fit in the main repo, that's fine as well.

BTW: There are swift3 templates but in swift3 enum cases will be `lowerCamelCase`, not `CamelCase`, [see here](https://github.com/apple/swift-evolution/blob/master/proposals/0006-apply-api-guidelines-to-the-standard-library.md). so swift templates need to change but should be trivial to do that I guess, just letting you know.

AliSoftware:
Ah right, got it!

Will look into it then and will consider merging 👍 

---

As for Swift3 templates we should indeed use lowerCamelCase you're right! Let's continue this discussion in the (closed) issue #116 (or the PR #117 that added that Swift 3 support)

AliSoftware:
Ok looks ok to me be could you please:

* add a simple unit test (with the expected generated code in `UnitTests/expected/….swift.out`) so that we ensure this template won't break in the future
* add an entry in the CHANGELOG to mention that new template and credit yourself!

Thx!

akaralar:
Hey, I'm a bit busy nowadays, will do as you said in a few days and let you know. Also is there a documentation explaining what different templates do?

AliSoftware:
Cool!

> Also is there a documentation explaining what different templates do?

Ah, sadly no, but would be a good addition indeed. Instead of creating a Wiki page or Markdown, I'm thinking maybe making the `swiftgen template` be able to print the description of each template would be nice, like maybe we can check if there is a comment `{# … #}` on top of each template and consider that as the template description, so that each `*.stencil` file can embark its description directly (instead of having to maintain a wiki page or something). That can be part of #123 probably.

AliSoftware:
@akaralar any news on this? I'm planning on releasing 2.1.0 in a few days (probably next week-end), it would be awesome if I could include your template in it!

akaralar:
Hey, I totally forgot about this because I since discovered some flaws with the workflow I mentioned and stopped using that template :) But someone else might need it so I'll add the tests and the CHANGELOG today

akaralar:
@AliSoftware I added the tests and added relevant bit of information to the changelog.

CHANGELOG.md.orig
```diff
@@ -0,0 +1,325 @@
...
+  [Olivier Halligon](https://github.com/AliSoftware)
+* Add more `swiftgen templates` subcommands.  
+  [Olivier Halligon](https://github.com/AliSoftware)
+  * `swiftgen templates list` lists all the available templates
+  * `swiftgen templates which <name>` prints the path to the template named `<name>`
+  * `swiftgen templates cat <name>` prints the content to the template named `<name>`
+  * `<name>` here can be either a subcommand name like `colors` or a
+    composed name `colors-rawValue` for a specific template.
+<<<<<<< HEAD
+=======
```
arnonhongklay: diff

AliSoftware:
Rebased and Merged via 864b61df9e3b302d9d029e5a6d2ad77ed490686c

Issue #101, opened 2016-03-09
AppKit (Mac) support


toshi0383:
I think it'd be awesome if we add AppKit support.

I should consider what to do, but this is just an idea for now.😉

AliSoftware:
I agree with the idea, but not sure about the implications.

Probably only affect templates (which would need to use `NSColor` instead of `UIColor`, etc) and not so much SwiftGen's own source code (which is dedicated for parsing input files like storyboards, asset catalogs and color palette files, which probably have the same format on an OS X project, we'll have to check but I guess), so that will probably be an easy enough PR to implement :+1:

tbaranes:
+1 on this, I'd love to use `SwiftGen` in OS X app :tada:

AliSoftware:
Took a quick look and confirms it will only be a matter of duplicating existing templates and just adjusting them for AppKit.

But as I'm not developing for AppKit and have no OS X projet, I'd be glad if someone actually using OS X projects and knowing what code they expect to just propose the templates and I'll happily merge them!

tbaranes:
I started to play with the storyboards template, but finally, I don't have nice usage for them since we don't have any navigations on mac. For now, I think it's the less important templates, so let's find nice use case before.

@toshi0383 @AliSoftware Any suggestions?

AliSoftware:
I think @toshi0383 would have more insight than me as I'm not an OS X developer, so idk how OS X developers use Storyboards and would prefer the template to look like :wink:

I'll merge the Images, Colors and Fonts PRs soon — Adding a note in the README could be worth it too — and then it's up to you to tell me if I should also wait for a Storyboard template before releasing a new version (if you think you'll be able to create one soon) or if you don't need it or will provide it later and can wait for a later release.

tbaranes:
Since the OS X platform is mainly using one window and work inside, it's a lot of more limited than iOS for most of the apps (there always exception). I don't think to add the storyboard features now, but I will definitely do it if I find a use case matching the need.

These four new generators will already make a nice release and it's a first step for SwiftGen on OS X! I'm looking forward for it :)

toshi0383:
Users can instantiate controllers by NSStoryboard's `instantiateInitialController()` or `instantiateControllerWithIdentifier(_:)` methods.
Even we don't have navigations on mac apps, I think we can simply have a wrapper around this. So at least we can provide a type-safe view loading experience to users.

[...]

PR #125, opened 2016-06-03
Images templates for OS X

tbaranes:
Here the first templates (and their tests) for OS X: Images
I tested the template in my own project, it seems to be working without any issues!

I didn't update the README, since the usage doesn't change, not sure what we should add, maybe a note? Also, the same localizable templates for iOS can be reused for OS X, nothing to change. 

#101 🎉


AliSoftware:
Super cool, thanks!! Will look at it soon but seems ace! 👍

tbaranes:
When I wake up this morning, I got an idea and can't get away from it, I have to do it! Here a nicer way to handle Images for OS X: don't duplicate any templates (that was the easiest way), reuse the same templates, and the same tests for both since the output difference is just a type:

```swift
#if os(iOS)
  import UIKit.Image
  typealias Image = Image
#elseif os(OSX)
  import AppKit.NSImage
  typealias Image = NSImage
#endif
```

AliSoftware:
Ohhh indeed great idea! :100:!

AliSoftware:
Hey @tbaranes 

To thank you for your work on these latest PRs and also in an effort to follow the same rules as [Moya's Contributors Guidelines](https://github.com/Moya/contributors), I've added you as a collaborator on SwiftGen.
That doesn't mean any special expectation from you, it just gives you the possibility now to review PRs and Issues and merge PRs yourself and contribute more easily to SwiftGen in the future if you ever want to.

Thx 👍

tbaranes:
Thanks for the invitation!

PR #126, opened 2016-06-03
Colors templates for OS X

tbaranes:
This one contains the same things as #125 (less the changelog update) but for the Colors
Related to #101


templates/colors-osx-default.stencil
```diff
@@ -0,0 +1,45 @@
...
+{% if colors %}
+{# Note: We don't use a UInt32 rawValue in this template so we can have multiple colors with the same rgbaValue #}
+enum {{enumName}} {
+  {% for color in colors %}
+  /// <span style="display:block;width:3em;height:2em;border:1px solid black;background:#{{color.rgb}}"></span>
+  /// Alpha: {{color.alpha|hexToInt|int255toFloat|percent}} <br/> (0x{{color.rgba}})
+  case {{color.name|swiftIdentifier}}
+  {% endfor %}
+
+  var rgbaValue: UInt32! {
```
AliSoftware: I'm wondering why `rgbaValue` is an implicitly-unwrapped-optional? I think we can drop the `!` here.

_(I mean, we have the same problem with the iOS templates, it's not specific to your addition in this PR… it's just that I'm realizing this as I'm reading the code again while reviewing your template, and it's probably from legacy when the template had a different structure but now it could be the occasion to remove them)_

tbaranes: Agree with you, if there no reason, let's drop it 👍

PR #127, opened 2016-06-03
Fonts templates for OS X

tbaranes:
Same as #125 and #126 but for fonts! Related to #101 
Last one will be the storyboards templates, but also the longest, so it will be for next week 🔜


AliSoftware:
You're unstoppable! I like that 😆 👍

PR #128, opened 2016-06-04
Storyboard templates for OS X

tbaranes:
Following the comment of @toshi0383, I give a second try. This PR is not completed yet, it still missing more examples and advanced use cases. For the example, I have two tests running:
- Empty storyboard template (maybe the most important right now)
- Anonymous storyboard: `NSWindowController` -> `NSSplitViewController` (as window content) -> left view controller / right view controller

Since `instantiateInitialController` and `instantiateControllerWithIdentifier` are returning `AnyObject`, and since `NSWindowController` and `NSViewController` doesn't have any commun parent, the template is returning `AnyObject` too, it will be up to the user to do the cast. I'm not really fan of this, but don't have any elegant solution for now.

In order to keep the useful `performSegue` as iOS, I added an extension on `NSWindowController` as well. I did a quick look on the others controllers available (split, tab...) are inheriting of `NSViewController`.

@toshi0383: what do you think of the output?
@AliSoftware I have `testAllStoryboardsWithCustomName` and `testAllStoryboardsWithDefaults` due to the OS X storyboard. Should we split the storyboards iOS and OS X in two different directories?

### DIFF
--- a/SwiftGen.xcodeproj/project.pbxproj
+++ b/SwiftGen.xcodeproj/project.pbxproj
@@ -85,6 +85,10 @@
 		E2B1F5841D0272A10000E40A /* storyboards-osx-default.stencil in Resources */ = {isa = PBXBuildFile; fileRef = E2B1F5831D0272A10000E40A /* storyboards-osx-default.stencil */; };
 		E2B1F5861D0272CF0000E40A /* Storyboards-osx-Anonymous-Defaults.swift.out in Resources */ = {isa = PBXBuildFile; fileRef = E2B1F5851D0272CF0000E40A /* Storyboards-osx-Anonymous-Defaults.swift.out */; };
 		E2B1F5881D0274B90000E40A /* Storyboards-osx-Empty.swift.out in Resources */ = {isa = PBXBuildFile; fileRef = E2B1F5871D0274B90000E40A /* Storyboards-osx-Empty.swift.out */; };
+		E2B1F58C1D0282460000E40A /* Placeholder-osx.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = E2B1F58B1D0282460000E40A /* Placeholder-osx.storyboard */; };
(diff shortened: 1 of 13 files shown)


tbaranes:
Just added one more use case: Message (adapted from the iOS Message.storyboard), it contains different controller: NSSplitController, NSTabController, NSWindowController, and at least one sample of each segue... well, an interesting case 😆 

The test is currently failing, SwiftGen replace all the `NS` prefix by `UI`, I think it's an issue in the core since the template doesn't contains anything related. You can take a look at `[Storyboards-osx-Anonymous-Defaults.swift.out](https://github.com/tbaranes/SwiftGen/blob/feature/storyboards_templates/UnitTests/expected/Storyboards-osx-Message-Defaults.swift.out)` to have an idea of the output that will be.

AliSoftware:
> I have testAllStoryboardsWithCustomName and testAllStoryboardsWithDefaults due to the OS X storyboard. Should we split the storyboards iOS and OS X in two different directories?

@tbaranes Indeed, that's a good idea to have dedicated subdirectories for those fixtures.

AliSoftware:
> The test is currently failing, SwiftGen replace all the NS prefix by UI, I think it's an issue in the core since the template doesn't contains anything related

That's indeed hard-coded in the source code, which is bad.

I annotated the code with 2 `// TODO:` comments just recently — which will show up as warnings thanks to the new SwiftLint script I recently added — to highlight where that needs to be changed. You might want to rebase your PR on top of master btw to get them here.

We'll probably need to add a new field to the stencil context in addition to `"identifier"` and `"class"`, like `vcType` or something, to store the `scene.tag` there (which is the XML tag name and is usually either `<viewController …>`, `<navigationController …>`, `<tableViewController …>` and the likes). Then we'll have to alter the templates to use either `{{scene.class}}` if it's defined (if it's a custom class), or the `UI{{scene.vcType|titlecase}}` if `scene.class` is `nil`. And then we can easily change that `UI` into `NS` for the OS X template.

tbaranes:
Got it. I was sure that the storyboards will be more complex than the others 😆 
Since you give me some rights, I will move my fork in a new branch directly here, I will also close this PR in favour of a new one from the branch.

PR #130, opened 2016-06-04
Update UIFont example on readme to match default template

phatblat:
I really like the `init(family:size:)` convenience constructor in the [usage example](https://github.com/AliSoftware/SwiftGen#usage-1). This changes the default template to match.

### DIFF
--- a/README.md
+++ b/README.md
@@ -429,7 +429,7 @@ struct FontFamily {
 // Helvetica Bold font of point size 16.0
 let font = FontFamily.Helvetica.Bold.font(16.0)
 // Another way to build the same font
-let sameFont = UIFont(family: FontFamily.Helvetica.Bold, size: 16.0)
+let sameFont = UIFont(font: FontFamily.Helvetica.Bold, size: 16.0)
 ```
 
 ---


AliSoftware:
Hi @phatblat!

It seems you forgot to update the test fixtures as well :wink:

But anyway, I don't think it's a good idea to make that change. If anything has to be fixed it's the typo in the README and not the constructor signature.

This is because the term "Font Family" in the fonts vocabulary means the group of fonts with the same name but only style variants. For example, "Helvetica" is a font family, containing the font variants "Helvetica-Regular", "Helvetica-Bold", "Helvetica-Oblique", "Helvetica-BoldOblique", and so on.

Here the constructor expects a specific font *variant*, like the enum case representing "Helvetica-Oblique", not a font family. So the name of this parameter should definitely not be `family:`. We might consider renaming it from `font:` to `fontVariant:` maybe but that's it.

We still need to fix that typo in the README though so feel free to amend your PR to do that instead :wink:

phatblat:
Ah, that's right. I'll revert the change and update this PR to change the readme instead.

phatblat:
All fixed up! Wait, does it even make sense to update the changelog for this simple read change?

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ---
 
+## Master 
+
+* Update UIFont example on readme to match default template
```
AliSoftware: Please add a full-stop + 2 spaces at the end of this line. That's the Markdown syntax so that the next info (your name + PR link) goes to next line in same paragraph.

phatblat:
Updated changelog syntax

AliSoftware:
Yeah indeed for this small change that's maybe not worth having a CHANGELOG entry indeed

phatblat:
OK, I'll pull that change out.

PR #131, opened 2016-06-05
Storyboards templates for OS X

tbaranes:
Just migrated all the works did in my fork in a local branch, closing #128 in favour of this one.

Things to do (following the previous discussion):
- [x] Separate iOS and OS X storyboards in different fixtures directory in order to fix tests `testAllStoryboardsWithDefaults` and `testAllStoryboardsWithCustomName`
- [x] Drop the hard coded `UI` to handle `NS` prefix (cf @AliSoftware explanation below)
- [x] Find a satisfying output 😆 

> I annotated the code with 2 // TODO: comments just recently — which will show up as warnings thanks to the new SwiftLint script I recently added — to highlight where that needs to be changed. You might want to rebase your PR on top of master btw to get them here.

> We'll probably need to add a new field to the stencil context in addition to "identifier" and "class", like vcType or something, to store the scene.tag there (which is the XML tag name and is usually either <viewController …>, <navigationController …>, <tableViewController …> and the likes). Then we'll have to alter the templates to use either {{scene.class}} if it's defined (if it's a custom class), or the UI{{scene.vcType|titlecase}} if scene.class is nil. And then we can easily change that UI into NS for the OS X template.

⚠️  The current state of this PR is more a proof of concept to discuss the output than a final implementation ⚠️  

PS: I will be in vacation starting mid of next week for two weeks, I will get this PR ready once I'm back if anyone finish it 👯


AliSoftware:
Hey! I had a problem with Travis on master :confused: — which was still using Xcode 7.0 and failed compilation even if it were working locally.

You might want to rebase onto after that fix I just pushed (otherwise even if your own tests are OK travis will fail on your branch and the PR won't be able to be merged)

tbaranes:
Small updates:

>  Separate iOS and OS X storyboards in different fixtures directory in order to fix tests testAllStoryboardsWithDefaults and testAllStoryboardsWithCustomName

I ended by adding an `osx` option to `parseDirectory`, which will just check if the storyboard name contains "osx" or not and add it to the list to check. It's less nicer than split the directory, but allow us to keep the same structure as the others generators.

>  Drop the hard coded UI to handle NS prefix (cf @AliSoftware explanation below)

First version of this. @AliSoftware is that what you were thinking? I didn't do for the segues type yet, but I drop the `UI` prefix hard coded, to put it in the template. About the template, is there any better syntax to have multiple if conditions? I check in stencil README but didn't find anything related.

AliSoftware:
My idea was more along the lines of:

```
return ["identifier": scene.storyboardID, "class": scene.customClass, "vcType": scene.tag]
```
And then let the template do all the logic, including using `{{class}}` if not nil, and using `UI{{vcType|titlecase}}` otherwise, and thus also being responsible for the title-casing in the template and not in the code.

AliSoftware:
> I ended by adding an osx option to parseDirectory, which will just check if the storyboard name contains "osx" or not and add it to the list to check. It's less nicer than split the directory, but allow us to keep the same structure as the others generators.

tbh I don't like this option. Because that means that people using SwiftGen will _have to_ name their Storyboards with `osx` in their name. I don't want to impose that (and I don't understand why it would be required).

I recently grouped the different files in `UnitTests/expected` into subfolders in Xcode. It's probably time we do the same for the `UnitTests/fixtures`  folder to split it into subfolders like `Images/`, `Colors/`, `Strings/`, `Storyboards-iOS/` and `Storyboards-OSX/` (the subfolder `Fonts/` already exists anyway)

>  About the template, is there any better syntax to have multiple if conditions? I check in stencil README but didn't find anything related.

I don't think there is yet. kylef/Stencil#52 is not yet implemented and it probably will be the first step in then supporting operators like `&&` and `||`, by they're not there yet it seems.

> I'm also encountering another issue for which I don't have a great solution after some though: […]

I think that logic should be handed in the template itself, as I said in my previous comment. But we'll probably need kylef/Stncil#52 first so that the template can do comparisons that would probably look like this:

```
{% if scene.vcType == 'viewController' %}
/* branch without the guard let */
{% else %}
/* branch with guard let … as? … */
{% endif %}
```

AliSoftware:
Maybe until kylef/Stencil#52 is implemented and released we can workaround by providing the comparison result in a dedicated field in the Stencil context:

```
return ["identifier": scene.storyboardID, "class": scene.customClass, "vcType": scene.tag, "isBaseViewController": (scene.tag == "viewController")]
```

This way both template (iOS & OS X) will have everything they need to do the comparisons. In the iOS template you'll be able to check `{% if isBaseViewController %}` and avoid the `guard let` in that case — because you know it's neither a custom class nor a special `vcType` like `navigationController` or others — and in the OSX template you won't use that `isBaseViewController` test if you don't need it at all.

AliSoftware:
_btw you might want to rebase on `master` before reorganizing the `UnitTests/*` folder into subfolders, as I renamed some of the `UnitTests/expected/*` files recently — so rebasing would avoid a potential conflict when merging later_ :wink:

tbaranes:
> tbh I don't like this option. Because that means that people using SwiftGen will have to name their Storyboards with osx in their name. I don't want to impose that (and I don't understand why it would be required).

Oh, didn't though that will make a mandatory condition for all the users, I'm completely joining you on this to say that is a bad idea. I though it was used only unit tests, my bad. I'm gonna change it and come back to the first solution.

Thanks for your answer, it's useful to have a better understand of about how all the logic is implemented SwiftGen 👍

AliSoftware:
np, don't hesitate to ask if something in the code or the project organization is not clear!

---

Side note; the more I think about it, the more I believe the fields in the Stencil context should be named `customClass` and `baseClass` or `baseType` instead of `class` and `vcType`, to make their intend more clear:

```swift
return [
  // The scene identifier, as set in IB
  "identifier": scene.storyboardID,
  // The custom class set in IB, e.g. "FooBarViewController"
  "customClass": scene.customClass,
  // The base class type on which this scene is base
  // (depending on the type of VC you drag & dropped in your Storyboard from the library pane)
  // possible values include 'viewController', 'navigationController', 'tableViewController'…
  "baseType": scene.tag,
  // indicate if the baseType is 'viewController' or anything else
  "isBaseViewController": (scene.tag == "viewController")
]
```

This way it's clear that the `customClass` field now only contains something if there actually _is_ a custom class, and would be empty if there is no special custom class set in IB — contrary to the old behavior before this PR where the `class` field was computed in SwiftGen's `StoryboardParser` code to contain either the custom class or the inferred one, but now that we'll do that logic in the template.

GenumKit/Stencil/Contexts.swift
```diff
@@ -82,11 +82,12 @@ extension StoryboardParser {
...
           .map { (scene: Scene) -> [String:String] in
             // Handle special scene.tag cases like navigationController, splitViewController, etc…
-            let customClass = scene.customClass ??
-              (scene.tag != "viewController" ? uppercaseFirst(scene.tag) : nil)
-            if let customClass = customClass where scene.customClass == nil {
-              return ["identifier": scene.storyboardID, "class": customClass]
-            } else if let customClass = customClass {
+            let customClass = scene.customClass ?? uppercaseFirst(scene.tag)
+            if scene.customClass == nil && scene.tag == "viewController" {
+              return ["identifier": scene.storyboardID, "class": customClass, "isBaseViewController": ""]
```
AliSoftware: More like `return ["identifier": scene.storyboardID, "class": scene.customClass, "isBaseViewController": true]`?

GenumKit/Stencil/Contexts.swift
```diff
@@ -82,11 +82,12 @@ extension StoryboardParser {
...
+            } else if scene.customClass == nil {
+                return ["identifier": scene.storyboardID, "class": customClass]
+            } else if let customClass = scene.customClass {
               return ["identifier": scene.storyboardID, "class": customClass, "vcType": customClass]
```
AliSoftware: more like `"vcType": scene.tag`

tbaranes: Didn't apply the change discuss in your comments yet, I will let you know once the code is ready for another review

UnitTests/expected/Storyboards-osx-Message-Defaults.swift.out
```diff
@@ -46,22 +46,22 @@ struct StoryboardScene {
 
     case MessageDetailsScene = "MessageDetails"
     static func instantiateMessageDetails() -> NSViewController {
-      return StoryboardScene.Message_Osx.MessageDetailsScene.controller()
+      return StoryboardScene.Message_Osx.MessageDetailsScene.controller() as! NSViewController
```
AliSoftware: We should avoid force-casts (`as!`) and use `guard let … as? … else { fatalError("some explicit message") }` instead, like I did recently on `master`.

BTW, could you confirm me that swiftlint warns you about this when you run your Unit Tests and complain that you shouldn't use `as!`? I added the SwiftLint Build Script Phase recently to automatically check code style even in the templates, but not sure if it does work in your computer or only on mine*

* you must [install swiftlint on your computer](https://github.com/realm/SwiftLint/blob/master/README.md) for the Build Scripts to run and warn you about those, though

GenumKit/Stencil/Contexts.swift
AliSoftware: ok 👍  sorry for having reacted _that_ quickly, I'll let you work now 😆 Good luck 😉

UnitTests/expected/Storyboards-osx-Message-Defaults.swift.out
tbaranes: Great idea to add SwiftLint, it's really useful. Didn't get any errors / warnings from it just by running tests. However, building the project (CMD + B) is reporting all the warnings (trailing_newline) in .swift.out for fonts, colors... but no errors for the force_cast. I'm also getting some `no such files or directory` from Xcode.

I'm not update from the master, that may be an explanation why I'm getting all these warnings, but don't know why I don't have the force_cast error. I will give a try later.

GenumKit/Stencil/Contexts.swift
tbaranes: Haha. At least, you are really fast to do your review 🎉 
I will do all the changes discuss above, so you may see some commits not fully ready to review!

AliSoftware:
Additional notes / checklists (sorry to spam you again ^^)

* [ ] Don't forget to re-generate the Storyboard resources used in the Playground if you've modified them, so that the Playground demo pages are using the latest compiled version of those storyboard fixtures to use in their demo code.  
This is done by using the `rake playground:resources` task in a terminal, which compiles the storyboards in `UnitTest/fixtures` into `*.xcplayground/Resources`
* [ ] You will probably need to update the `Rakefile` if you move the fixtures in subfolders btw, especially [those 3 lines here](https://github.com/AliSoftware/SwiftGen/blob/master/Rakefile#L145-L153) just to update the paths of the fixtures the Rakefile needs to compile

And another little checklist for once the PR is ready:

* [ ] It would be good to check if the Playground demos are still working once you've made all those changes. Idk if it's worth adding a dedicated demo page for OS X storyboards as well in that playground or not?
* [ ] Dropping a line in the README about that OS X storyboard support would probably be a good idea as well

UnitTests/expected/Storyboards-osx-Message-Defaults.swift.out
AliSoftware: Indeed, your branch seems to be based on a commit _after_ I added the SwiftLint script in the main target… but _before_ I decided to split that script in two (one part in the `swiftgen-cli` target to only check the swiftlint + GenumKit codebase, another in the `UnitTests` target to test the templates generated code).

That's why you have SwiftLint running only when Building but not when running tests, while as I now decided to split them in the 2 targets since that commit you branched from, I now test the templates when doing a `Cmd-U` instead of a `Cmd-B`.

That doesn't explain why you don't have any `force_cast` warning though. Might be worth checking it warns properly once you'll have rebased your branch, and investigate why it doesn't catch those if you still don't have any warning about them when in `master`.

tbaranes: Indeed, after rebasing I have swiftlint running when testing, still no force_cast error. Also, do you also have in master all these warnings:

screen shot 2016-06-05 at 23 40 22

AliSoftware: Ah, I don't, but that's probably because you did a Ctrl-I to re-indent your code and now have all the empty lines "indented" (i.e. your empty lines actually contain spaces)

You can check a CheckBox in Xcode > Preferences > Text Editing to prevent Xcode from adding those every time.

AliSoftware: Now that I'm on my Mac, here's the screen capture of that setting in Xcode that make Xcode create empty lines unintended
xcode-setting-newlines

_Note that you can fix all those warnings automatically using some `swiftlint autocorrect` command in the Terminal instead of fixing them one by one manually_

---

Also, I just added a `as!` in a template just to check, and confirm that SwiftGen did generate a warning for it, so not sure why you don't have it:
Force-Cast SwiftLint error
I realize by reading my script again though that it does scan for `*.swift.out` files only at the root level of the `UnitTests/expected` folder, but not in subfolders if any. I'm gonna fix that in an upcoming commit.

tbaranes: I enable the option just in case, but the warnings doesn't seem to come from here. In fact, I don't even know why I'm getting that warnings since the `trailing_newline` rule is respected in all the files. I tried different solution: fixing myself, `swiftlint autocorrect`... nothing changed. 

I'm gonna try on master to see if I'm getting the same warnings

tbaranes:
Alright, this time you can take a look at it @AliSoftware 😆 

I updated the following:
- System to drop UI from hard code to move it in the templates. Using `baseType`, `customClass`...
- Renaming stencil fields as you suggested
- Handle `isBaseViewController` in template for iOS
- Drop the `as!` from templates 

Also, SwiftLint created errors for each `try!` used in `StoryboardsTests`, for now I just replaced them by:

```
    do {
      let result = try template.render(parser.stencilContext())
      let expected = self.fixtureString("Storyboards-Message-Lowercase.swift.out")
      XCTDiffStrings(result, expected)
    } catch {}
```

Let me know which behaviour do you prefer in the catch: just a log, fatalError, nothing, test failure...

tbaranes:
Alright, this time you can take a look at it @AliSoftware 😆 

I updated the following:
- System to drop UI from hard code to move it in the templates. Using `baseType`, `customClass`...
- Renaming stencil fields as you suggested
- Handle `isBaseViewController` in template for iOS
- Drop the `as!` from OSX templates

AliSoftware:
Cool 😎 Will take a look soon!

Beginning of week will probably be quite busy for me, so don't think I'll have time to review & merge before you leave for your vacation :wink: (you leave in the upcoming days, right?)

Will take a closer look in the next days and let you know. In the meantime if I don't see you online before that, enjoy your vacation!

PS : I've just released SwiftGen 1.1.0 (which include the first OS X templates + the bug fix for the Fonts template) 🎉

tbaranes:
Just see your release! Awesome, I pass the link to the other OS X dev 🎉 
Yeah living thursday (in fact, I'm already in vacation but had few days working remotely to do), anyway, thanks!

I will take care of the change once I'm back. 👍

UnitTests/expected/Storyboards-osx-Message-Defaults.swift.out
AliSoftware: Ah, indeed the warning was not about trailing whitespaces in empty lines, but about trailing newlines at end of file, I misread it sorry.

That's probably because of the way I parse `.swift.out` files with SwiftLint. Since SwiftLint is only parsing files with extension `*.swift` given a directory, I had to cheat a little in the `scripts/swiftlint-templates.sh` script and instead use something like `cat *.swift.out | swiftlint lint --use-stdin`. That's probably because of that `cat`+pipe that SwiftLint considers there is a missing line at end of that input, probably because `cat` eats that trailing line?
Strange that I don't have that warning on my end, will try to investigate more.

---

In the meantime, [I added the `force_unwrapped` opt-in rule in `.swiftlint.yml` on `master`, which revealed that my templates for storyboards still had a `!` somewhere in their code](https://github.com/AliSoftware/SwiftGen/commit/9326ef4376086fd716d45c765c203621b39d674c). So I fixed that since then but you might want to pick or rebase on that change to ensure your OS X storyboard templates don't have those `!` hidden somewhere as well.

AliSoftware: @tbaranes: Could you try in a terminal on your Mac to run the following command and tell me if it finds any warning, especially the one about the trailing newline?

```sh
cat UnitTests/expected/Storyboards-All-Default.swift.out | swiftlint lint --use-stdin
```

If so, maybe that means that your `cat` appends some newline at the end of the output after printing the content of the file, while mine doesn't? (maybe you use a different shell? I tried with `zsh` and `sh` both but couldn't reproduce the warning)

tbaranes: After a rebase on master, here the output: `Done linting! Found 0 violations, 0 serious in 1 file.`
I'm using zsh, but I got no warnings, I have them only when building when Xcode. Did you try to build using `feature/storyboards_osx`branch? I don't think that will change anything but...

tbaranes: I just understand why I'm getting these warnings: I have a space in a folder name: `/path/with space/to/SwiftGen`. Dropping this space solves the issues, I just have warnings about an `!` left (thanks to your update).

So, I'm gonna change my folder name now 😆

AliSoftware: Ah! Interesting, and makes sense indeed!

But the proper fix would be to patch my scripts so that they can handle paths with spaces, so that it works for everyone :wink:  I'm guessing that would just be a matter of wrapping some `$XX` variables into quotes so should be easy to test at least.

tbaranes:
@AliSoftware I'm back here 😆 
So, I just updated some small stuff:
- Add a `NSStoryboard` section in README since the output and usage is different from iOS, I think it can have a full section for itself
- Updated playgrounds fixtures path / ressources as you asked

Since I got no feedbacks on the output, I think we can go with that for a first version. We can still improve it later. 

Still missing two points:
- lowercase and uppercase templates: are they really necessary? I'm have no used for them personally, so should I port them directly or wait for someone that needs them?
- add a playground page for storyboard: it's up to you if you think that we should have one or not. Just let me know what do you prefer

AliSoftware:
Thx!

The uppercase vs lowercase template is because a lot of people use the same name for the storyboard and the VC class and this might cause naming conflicts, hence the lowercase variant.

We can't really add a page for that in the playground as playgrounds are tied with a SDK; he one we have is an iOS playground so if we wanted to demo the OSX templates we'd need a separate ~OSX~ macOS playground. I don't think it's worth it given the similarities with iOS.

tbaranes:
Just added the lowercase template, but I prefer to wait for a feature request to do the uppercase one. I don't think it's a good idea to add too much templates before they are needed. 

I just have the CHANGELOG to update, otherwise, everything has been implemented for this PR. Finally ready to go 🎉

AliSoftware:
Cool! Will take a look later and merge once you've added the CHANGELOG entry.

_Not being an OS X developer either, I think I'll still merge it anyway even if not broadly tested by us non-OSX devs, and see if there are any feedback once it'll be up._

GenumKit/Parsers/StoryboardParser.swift
```diff
@@ -84,10 +84,11 @@ public final class StoryboardParser {
     storyboardsSegues[storyboardName] = delegate.segues
   }
 
-  public func parseDirectory(path: String) {
+  @available(OSX 10.10, *)
+  public func parseDirectory(path: String, osx: Bool = false) {
     if let dirEnum = NSFileManager.defaultManager().enumeratorAtPath(path) {
       while let subPath = dirEnum.nextObject() as? NSString {
-        if subPath.pathExtension == "storyboard" {
+        if subPath.pathExtension == "storyboard" && subPath.containsString("osx") == osx {
```
AliSoftware: wait, how is this still in the code? Didn't we decide that we shouldn't make that logic — that use the name of the file to decide if it's OS X or not — in the source of SwiftGen?

I guess since you now reorganized the fixtures directory into subdirectories to be able to separate iOS and OS X fixtures, this is no longer necessary, right?

GenumKit/Stencil/Contexts.swift
```diff
@@ -66,7 +66,9 @@ extension AssetsCatalogParser {
     - `name`: `String`
     - `scenes`: `Array` (absent if empty)
        - `identifier`: `String`
-       - `class`: `String` (absent if generic UIViewController)
+       - `customClass`: `String` (absent if generic UIViewController)
+       - `isBaseViewController`: `Bool`
```
AliSoftware: maybe we should comment that a little more, to explain what we call "baseViewController" / "baseType", so that people wanting to implement their template understand what values those fields can have and what they mean

README.md
```diff
@@ -320,6 +319,167 @@ override func prepareForSegue(_ segue: UIStoryboardSegue, sender sender: AnyObje
 initialVC.performSegue(StoryboardSegue.Message.Back)
 ```
 
+## NSStoryboard
+
+```
+swiftgen storyboards --template storyboards-osx-default.stencil /dir/to/search/for/storyboards
```
AliSoftware: The `--template` option of SwiftGen doesn't expect/need you to specify the `.stencil` extension :wink:

README.md
```diff
@@ -320,6 +319,167 @@ override func prepareForSegue(_ segue: UIStoryboardSegue, sender sender: AnyObje
...
+    case MessageListScene = "MessageList"
+    static func instantiateMessageList() -> NSViewController {
+      guard let vc = StoryboardScene.Message_Osx.MessageListScene.controller() as? NSViewController
+      else {
+        fatalError("ViewController 'MessageList' is not of the expected class NSViewController.")
+      }
+      return vc
+    }
+
+    case MessagesTabScene = "MessagesTab"
```
AliSoftware: This example / sample code is probably a bit too long for a README. Maybe add a `[… snip …]` in the middle to truncate it, or simply remove some cases to only show 1 or 2 in the README. After all, it's there only to show the general principle of how the generated code will look like, to give the new user an overlook, not to detail every use case :wink:

GenumKit/Parsers/StoryboardParser.swift
tbaranes: This is still necessary but just for running tests. I didn't find a solution to get the right path from NSBundle. If you have a workaround to avoid this, I will definitely change it.

Otherwise, when used in a real case, the user will never have to add `osx` to their storyboard names, it will work transparently. It's really used only for the unit tests.

AliSoftware: Found a way to fix it :wink:

AliSoftware:
@tbaranes I've fixed the hack on StoryboardParser for tests.

Now the fixture folders `Storyboards-iOS` and `Storyboards-OSX` folders have beed added as **folder references** (blue folder icon in Xcode) instead of Groups (yellow folders in Xcode).

By doing this, this means that when the `UnitTests.xctest` bundle is build by Xcode, instead of copying those `*.storyboard` files at the root of `UnitTests.xctest/Contents/Resources/` and putting them all at the same level in the compiled xctest bundle, the folders are copied as-is, keeping the folder hierarchy, and thus copied into separate subfolders `UnitTests.xctest/Contents/Resources/Storyboards-iOS/` and  `UnitTests.xctest/Contents/Resources/Storyboards-OSX`.

This way, I can make the Unit Tests only scan for the specific subfolder in the bundle and thus only scan for iOS _OR_ OS X storyboards but not both. I've adapted the `TestsHelper` accordingly so that the `fixturePath` and `fixturesDir` functions now takes an optional parameter to specify an optional subdirectory for the fixture and there you go!

tbaranes:
Awesome. Didn't know that. Thanks for the fix!

GenumKit/Stencil/Contexts.swift
```diff
@@ -80,13 +83,17 @@ extension StoryboardParser {
...
           .sort({$0.storyboardID < $1.storyboardID})
-          .map { (scene: Scene) -> [String:String] in
-            // Handle special scene.tag cases like navigationController, splitViewController, etc…
-            // TODO: Fix this to extract the 'UI' prefix and be able to implement OSX support in #128
-            let customClass = scene.customClass ??
-              (scene.tag != "viewController" ? "UI" + uppercaseFirst(scene.tag) : nil)
-            if let customClass = customClass {
-              return ["identifier": scene.storyboardID, "class": customClass]
+          .map { (scene: Scene) -> [String:AnyObject] in
+            if scene.customClass == nil && scene.tag == "viewController" {
```
AliSoftware: Just realized that this condition is the same as the next on and both could be grouped on only one condition:

```swift
 if scene.customClass == nil {
  return [
    "identifier": scene.storyboardID,
    "baseType": uppercaseFirst(scene.tag)
    "isBaseViewController": (scene.tag == "viewController")
  ]
}
```

GenumKit/Stencil/Contexts.swift
```diff
@@ -66,7 +66,10 @@ extension AssetsCatalogParser {
     - `name`: `String`
     - `scenes`: `Array` (absent if empty)
        - `identifier`: `String`
-       - `class`: `String` (absent if generic UIViewController)
+       - `customClass`: `String` (absent if generic UIViewController)
+       - `isBaseViewController`: `Bool`, indicate if the baseType is 'viewController' or anything else
+       - `baseType`: `String` (absent if class is a custom class). The base class type on which a scene is base.
+          Possible values include 'viewController', 'navigationController', 'tableViewController'…
```
AliSoftware: Reading th code, it seems you always uppercase the first letter of this value. So:

1. either the documentation should reflect that and list the examples with an uppercase letter
2. or the code should be fixed to stop upper-casing the first letter and keeping the value intact (letting that transformation responsibility to the template using a filter if it needs it)

I think the 1st solution is the easiest and more useful, as people will always use the `uppercaseFirst` variant of this in their template anyway.

GenumKit/Stencil/Contexts.swift
```diff
@@ -80,13 +83,17 @@ extension StoryboardParser {
...
+                "identifier": scene.storyboardID,
+                "baseType": uppercaseFirst(scene.tag),
+                "isBaseViewController": true
+              ]
+            } else if scene.customClass == nil {
+                return ["identifier": scene.storyboardID, "baseType": uppercaseFirst(scene.tag)]
+            } else if let customClass = scene.customClass {
+              return ["identifier": scene.storyboardID, "customClass": customClass]
             } else {
               return ["identifier": scene.storyboardID]
```
AliSoftware: I don't see when that `else` case can happen. the two above conditions covers both `if scene.customClass == nil` and `if let customClass = scene.customClass` (which pass only if `scene.customClass != nil`) so by that time one of those two path should have been taken and the `else` doesn't cover any case left.

This also means that you can rewrite that whole `if` as a simple:

```swift
if let customClass = scene.customClass {
  return ["identifier": scene.storyboardID, "customClass": customClass]
} else {
  /* customClass is nil */
return [
    "identifier": scene.storyboardID,
    "baseType": uppercaseFirst(scene.tag)
    "isBaseViewController": (scene.tag == "viewController")
  ]
}
```

Which reduces from 4 conditions to only two and makes the whole conditional block more readable :wink:

CHANGELOG.md
```diff
@@ -9,7 +9,10 @@
 * Added support for tvOS and watchOS in images, fonts and color templates.  
 [Tom Baranes](https://github.com/tbaranes),
   [#145](https://github.com/AliSoftware/SwiftGen/pull/145)
-  
+* Added support for OSX in storyboards.
```
AliSoftware: We should probably mention here that `class` has been renamed `customClass` and the Stencil context for storyboards has changed a bit. This would be especially important for people that potentially implemented their own custom templates for Storyboards and used the `class` entry in them.

GenumKit/Stencil/Contexts.swift
tbaranes: We can't do that (or I missed  feature from `stencil`). In order to check if it's a `isBaseViewController`, we are using `{% ifnot scene.isBaseViewController %}` which consist into simply check if the key exists are no whatever is the value.

Anyway, you are still right that the current conditions are more complex than they should do.

tbaranes:
Thanks for your comments! All of them has been cared of 👍

CHANGELOG.md
tbaranes: If you have any better formulation / explanation than my update, let me know, or update the CHANGELOG directly :)

AliSoftware:
_Come on Travis, what's taking you so long? I want to publish a release tonight and the CI makes the suspens killing me_ :trollface:

tbaranes:
Seems like wants us to move away from github for tonight 👮

GenumKit/Parsers/StoryboardParser.swift
```diff
@@ -84,6 +84,7 @@ public final class StoryboardParser {
     storyboardsSegues[storyboardName] = delegate.segues
   }
 
+  @available(OSX 10.10, *)
```
AliSoftware: Why did you add this restriction? This prevented me from compiling the SwiftGen OS X executable when I was trying to release so I removed it, but wanted to be sure what the primary intent was.
People even not in OS X 10.10 should still be able to parse iOS storyboards for example.

tbaranes: My bad, old code that I forgot to remove.

PR #132, opened 2016-06-05
Update IB metadata

phatblat:
It annoys me to no end that Xcode stores (seemingly unnecessary) tool and OS version information in storyboard and XIB files. I like to isolate these changes from features so that any conflicts are obvious that they come from a finicky tool instead of an feature change to the file.

While updating this metadata in the test fixtures (by simply looking at them), I noticed a bunch of generated binary files for the storyboard inside the playground are checked in. These don't appear to be necessary to be versioned as they will simply be regenerated when someone views the storyboard.


AliSoftware:
Are you sure storyboards _are_ compiled automatically when simply placed as resources in the playground?

In my experience just putting a storyboard in the Resources folder of a playground doesn't make it be compiled and thus the playground pages end up not being able to demonstrate the `StoryboardScenes.messagesList.initialVC` features and stuff.

That's actually exactly why I had to add a `rake playground:resources` task to manually recompile the storyboard to a storyboardc every time I need to change the storyboard content in IB, so that the playground code can work and not generate any errors from an impossibility for it to load the VC from the SB if it's not previously compiled manually…

phatblat:
I'm not sure of anything 😓. Something I was doing was regenerating all of the .nib files inside `Wizard.storyboardc`, but looking closer I see now that `Wizard.storyboard` lives in the UnitTests and that rake task does the compiling.

I'll revert the change to ignore .storyboardc paths as without them the Storyboards-Demo breaks.

AliSoftware:
Should I then close this PR or not?

phatblat:
Ugh, I think I dropped the wrong commit.

AliSoftware:
Ok, seems fine reading this from GitHub. Note that we'll probably have to re-run `rake playground:resources` again soon given the work done on #128 though.

PR #136, opened 2016-06-06
Update README to include OS X

tbaranes:
Just few updates in the README to add the support of OS X:
- Add the NS{Image,Color,Font} to the titles
- Updated the templates output, I removed the `typealias` declaration since the templates doesn't show the full version, but is that understable? I think so.

You can have a look of the result [here](https://github.com/AliSoftware/SwiftGen/tree/feature/update_osx_readme).

Suggestions:
- About the summary, we can make the color / font title more readable, something like: 
    - `enums for your colors`.
    - `enums for your fonts`.
- Maybe add a badge beside the travis one to show which platform we support?


tbaranes:
All updated 👍

AliSoftware:
SwiftGen is a bit of an UFO in the pods world, because it's actually not a library that you'll use in your project code, like `OHHTTPStubs` or `Alamofire` or such. It's not even supposed to be a pod in the first place, as it's only a tool (it originally was only released over homebrew as a precompiled binary only).

I added CocoaPods support due to public demand, as people wanted to be able to update SwiftGen as easily as a `pod update` along with their pods, but the `SwiftGen.podspec` actually doesn't include code in your Pods project, but only make the `swiftgen` binary downloaded by CocoaPods somewhere in the Pods so you can use `$(PODS_ROOT)/SwiftGen/bin/swiftgen/…` in your Build Scripts. But there is no code or swift file associated that would be added to the Pods project like any other classic pod, so it's a bit cheating.

That's why the `SwiftGen.podspec` is marked as being compatible with iOS, OSX, tvOS and watchOS: so that people can include it whatever platform their project is building on, because we don't care as we don't include code in the pod, only the binary, so it doesn't matter if the project is a tvOS project or whatnot.

That being said, we could consider matching the platform listed in the `SwiftGen.podspec` with the one for which we officially provide a template, to indicate for which platforms the templates bundled with SwiftGen are valid. But that would prevent anyone targeting other platforms to use SwiftGen in their Podfile even if they intended to provide custom templates tailored to their project to generate ad-hoc code.

So because anyone can create their template, even create a watchOS template locally next to their project and use that, even if SwiftGen itself doesn't provide that watchOS template when it's installed, I think it's still fair to say that SwiftGen — the tool — is compatible with watchOS — as long as the user provides the appropriate template… in case the one for iOS doesn't match.
People could still make PRs to submit their watchOS template or amend the default template if that `if os()` needs to be amended to consider watchOS / tvOS.

README.md
```diff
@@ -358,12 +357,13 @@ NamedColor   : Translucent
...
 ```swift
-extension UIColor {
+// The Color type below is typealias'ed to UIColor on iOS and NSColor on OSX
+extension Color {
   /* Private Implementation details */
   ...
 }
 
-enum ColorName : UInt32 {
+enum ColorName {
```
AliSoftware: That `: UInt32` is generated when we use the old template now renamed `colors-rawValue.stencil`, which generated  `case Translucent = 0xffffffcc` cases.

Now that the new default template `colors-default.stencil` doesn't use `rawValues` but instead use a dictionary to map each `case` to the `UInt32` output reba value, then indeed that `: UInt32` is not part of the generated code anymore… but those `case Translucent = 0xffffffcc` should disappear as well!

So we should also update the body of that `enum` currently pasted in that README too so that the example code pasted here is still consistent with what is generated by the default template.

AliSoftware:
Should we drop a note in the "Storyboards" paragraph to mention that « there is no OS X storyboards yet (but a PR #131 is pending to add that template if you want to help) » in the README?

README.md
tbaranes: 😆 In fact I had updated the body, bu when I was review my changes, I though their was voluntary more colors. On my way to fix this.

tbaranes:
Updated:
- Add a note for OS X storyboard 
- Updated the expected output of colors

PR #138, opened 2016-06-09
Gotta return

phatblat:
Adds missing return to generated `initialViewController` method in storyboard templates.

screen shot 2016-06-08 at 8 59 34 pm


AliSoftware:
Thanks for the PR!

Damn, I really need to make [this script checking that the templates compile](https://github.com/AliSoftware/SwiftGen/blob/master/scripts/check_templates.sh#L1-L8) work some day!

AliSoftware:
Damn I merged before realizing you forgot to add entry in CHANGELING ^^ I will add it on master.

PR #145, opened 2016-06-22
Add tvOS and watchOS support for Images, Fonts, Colors

### LINKED ISSUE #143: Failing to compile on tvOS
This is due to the lines:

#if os(iOS)
  import UIKit.UIImage
  typealias Image = UIImage
#elseif os(OSX)
  import AppKit.NSImage
  typealias Image = NSImage
#endif

which does not include a case for tvOS

tbaranes:
Fix #143 
**Note:** as said in the other thread, I didn't test the output in a real project. Based on @Megatron1000 comments, it should work since it was working before.

OS X support didn't change localizables and storyboards templates, so if it was working before, it should still work as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,10 @@
 
 * Fix issue with txt files bailing on comments  
 [Derek Ostrander](https://github.com/dostrander), [#140](https://github.com/AliSoftware/SwiftGen/issues/140)
-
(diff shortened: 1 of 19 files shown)


AliSoftware:
Could you please add a CHANGELOG entry?

Otherwis e:+1:

tbaranes:
CHANGELOG updated 👯

AliSoftware:
Thanks again @tbaranes 💯

Issue #135, opened 2016-06-06
Add CONTRIBUTING.md


AliSoftware:
- Mention the Moya Contributors Guidelines and the fact that I wish to give push access to every contributor
- Document a bit the organization of the project: `swiftgen-cli` for the parsing of arguments (using `Commander`) and calling `GenumKit`, `GenumKit` for all the logic parsing the files and creating some Stencil contexts from that, `Stencil` for template rendering, …
- Document the steps to contribute: e.g. for every new template, don't forget to update Unit Tests too, for every new feature don't forget to include a Playground page, …
- Improve code documentation to help contributors understand the codebase better.

AliSoftware:
Perfect timing: We'll likely base that `CONTRIBUTING.md` file and all the related documentation and CoC on https://github.com/Moya/contributors/pull/7 :+1:

Dschee:
I'm **all in for a CONTRIBUTING.md** file since I just happened to have problems or at least questions when I started to work on #93.

I think the file should also include **a section on how to setup the project** from the contributor side. Here's **what I have done** and **what issues I faced** and am still facing:

- make sure Homebrew and CocoaPods are installed and up-to-date
- run `brew install swiftlint` (I had it installed, but this is what should be in the docs)
- run `pod install` in the root folder

**My first issue** was that `pod install` lead to the following warnings:

```
[!] The `swiftgen [Debug]` target overrides the `EMBEDDED_CONTENT_CONTAINS_SWIFT` build setting defined in `Pods/Target Support Files/Pods-swiftgen/Pods-swiftgen.debug.xcconfig'. This can lead to problems with the CocoaPods installation
    - Use the `$(inherited)` flag, or
    - Remove the build settings from the target.

[!] The `swiftgen [Release]` target overrides the `EMBEDDED_CONTENT_CONTAINS_SWIFT` build setting defined in `Pods/Target Support Files/Pods-swiftgen/Pods-swiftgen.release.xcconfig'. This can lead to problems with the CocoaPods installation
    - Use the `$(inherited)` flag, or
    - Remove the build settings from the target.
```

My CocoaPods version is 1.0.1. These warnings confused me – I can't tell if they are of any relevance.

**My second issue** was after opening `SwiftGen.xcworkspace` and hitting `Cmd+B` to build the swiftgen-cli target. Xcode built for some time but then **failed** with the following output:

```
[...]

PR #147, opened 2016-06-25
Initial CONTRIBUTING.md file

Dschee:
As a followup on our discussion in #135 I kicked off a CONTRIBUTING.md file with some initial content inspired from Alamofire (Moya doesn't seem to have one yet) and the project setup section I talked about. I hope this works fine as a base to start with.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,34 +1,26 @@
 # Contributing Guidelines
 
-This document contains information and guidelines about contributing to this project.
-Please read it before you start participating.
+This document contains information and guidelines about contributing to this project. Please read it before you start participating.
 
 **Topics**
 
-* [Asking Questions](#asking-questions)
-* [Reporting Issues](#reporting-issues)
+- [Asking Questions](#asking-questions)
+- [Reporting Issues](#reporting-issues)
 
 ## Asking Questions
 
-We don't use GitHub as a support forum.
-For any usage questions that are not specific to the project itself,
-please ask on [Stack Overflow](https://stackoverflow.com) instead.
-By doing so, you'll be more likely to quickly solve your problem,
-and you'll allow anyone else with the same question to find the answer.
-This also allows maintainers to focus on improving the project for others.
+We don't use GitHub as a support forum. For any usage questions that are not specific to the project itself, please ask on [Stack Overflow](https://stackoverflow.com) instead. By doing so, you'll be more likely to quickly solve your problem, and you'll allow anyone else with the same question to find the answer. This also allows maintainers to focus on improving the project for others.
 
 ## Reporting Issues
 
-A great way to contribute to the project
-is to send a detailed issue when you encounter a problem.
-We always appreciate a well-written, thorough bug reports.
+A great way to contribute to the project is to send a detailed issue when you encounter a problem. We always appreciate a well-written, thorough bug reports.
 
 When reporting issues, please include the following:
 
-* The version of Xcode you're using
-* The version of iOS or macOS you're targeting
-* The full output of any stack trace or compiler error
-* A code snippet that reproduces the described behavior, if applicable
-* Any other details that would be useful in understanding the problem
+- The version of Xcode you're using
+- The version of iOS or macOS you're targeting
+- The full output of any stack trace or compiler error
+- A code snippet that reproduces the described behavior, if applicable
+- Any other details that would be useful in understanding the problem
 
 This information will help us review and fix your issue faster.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,6 +6,7 @@ This document contains information and guidelines about contributing to this pro
 
 - [Asking Questions](#asking-questions)
 - [Reporting Issues](#reporting-issues)
+- [Project Setup](#project-setup)
 
 ## Asking Questions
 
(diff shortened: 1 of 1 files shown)


AliSoftware:
Seems like a great start, thanks.

Definitely want to include Moya's Contributor Manifest (which they now call `Community.md` and explain my will to give push access to every contributor too.

We'll end up probably having two different documents, one explaining the motivations, how to contribute, the community guidelines (CoC) and spirit of the project, and another one explaining the technical details (the need to have SwiftLint installed, why we can't Cmd-B on the command line app target, …), but at least that's a start, so 👍

CONTRIBUTING.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Project Setup
+
+Before you can start developing a fix or a new feature in this project you need to be aware of how this project is structured. Especially since SwiftGen is a command line tool written in Swift some things need to be handled differently.
+
+### Requirements
+
+First you need to make sure your machine fulfills the following requirements:
+
+- You have [Homebrew](http://brew.sh), [CocoaPods](https://cocoapods.org) and [xcpretty](https://github.com/supermarin/xcpretty) installed
```
AliSoftware: If I'm not mistaken, Homebrew and CocoaPods are only necessary if you intend to release a new version of SwiftGen.

If you clone the repository and start modifying the code, you can still build and compile it even if you don't have Homebrew, and CocoaPods is only necessary if you intend to update the dependencies SwiftGen depends upon.
And about `xcpretty`, it shouldn't be required at all.

CONTRIBUTING.md
```diff
@@ -0,0 +1,49 @@
...
+- You have [SwiftLint](https://github.com/realm/SwiftLint) installed – otherwise run `brew install swiftlint`
+- Install the dependencies by running `pod install` from the project root directory
```
AliSoftware: I commit/push my Pods project alongside the app project. So there is no need to `pod install` after you clone the repository and the `Pods/` project and all its pods are pulled alongside the project. (This would only be needed if you intend to add/update the existing SwiftGen dependencies)

CONTRIBUTING.md
Dschee: Ah I see, makes sense. But for `xcpretty` the `rake install` failed with an error and stating something about `xcpretty` so I installed it and it went away. That's why I thought there must be some dependency here.

AliSoftware: Ah. If the Rakefile fails when xcpretty isn't present then we must fix that so it isn't required anymore. After all it's only used to make the Xcode output more readable but everything should build fine without too.

Dschee: Good, shall we create a separate issue for that?

Also what of the requirements remains? Is SwiftLint needed? Looks like everything else is optional (although we could mentions those, too somewhere).

AliSoftware: Yes I think we should state in the CONTRIBUTING that only SwiftLint is required and only mention the other tools but don't require them. If one of the other tools especially xcpretty is technically required given the current state of the code we should make a separate PR to remove that requirement.

(Note that technically SwiftLint isn't really _required_ per se (I think the script only emits a warning if it's not installed) but I still want to consider it required because it ensures that the templates we bundle with SwiftGen are consistent and respect the coding conventions that many swift developers use and expect so I don't want to ship a new release of SwiftGen if it were not to respect those coding conventions)

Dschee: I agree on all points. Btw (and this is a bit off topic) when working on the dot syntax feature today I saw some code that I'm used to get warnings for from SwiftLint within the project like the `[String:AnyObject]` [here](https://github.com/AliSoftware/SwiftGen/blob/master/GenumKit/Stencil/Contexts.swift#L129). Just mentioning it, we could also create an issue for that if there isn't one already. I'm keeping those lines as is for now.

AliSoftware: You mean that you expected SwiftLint to warn about this line (because you forgot the space after the colons) but it didn't? So it seems to miss some warnings?
Maybe try to rebase your branch on top of master then as I know I fixed some bugs in my Script Build Phases invoking SwiftLint recently maybe that would solve them in your branch once rebased?

Dschee: Exactly, that's what I'm expecting.

I could rebase, but I started my fork on June 20th (so just a few days ago) and there are only 5 newer commits since then which were all not from you and have tvOS as their topic. So I doubt that rebasing will help at all. Also the [link from above]() links to the current master version and still has the same line of code ... and last but not least I already finished my feature as you can see in #148 and won't have much time for additional improvements for some time now. Just wanted to mention it.

I could also mention that you are using two whitespaces instead of the Xcode default (which is 4) for indentation which made it a hassle for me to edit code as I had to do all the indenting manually and Xcode broke them each time I opened or closed a brace (which is annoying). But that would open Pandora's box for time-consuming but nowhere-leading discussions on the pros and cons of following the conventions of a platform and so on and so forth. So, let's just keep it easy and ignore these small details. :)

AliSoftware: Noted.

PS: You can set the indentation level at project level (instead of changing it in your Xcode preferences for all your projects).

Dschee:
It says "unknown repository", so I'm closing it in favor of #149.

PR #148, opened 2016-06-28
Dot support in identifiers for structure

### LINKED ISSUE #93: Dot support in identifiers for structure
In our app, we use structured identifiers as the string keys for localizable strings, such as `error.title.password-required`. Before integrating `swiftgen`, we structured these ourselves into `struct` or  or `enum` definitions that had a hierarchy, i.e. one for `error`, one for `title`. It would be great if `swiftgen` could do the same thing and pick up the structure from dot-delimited string keys.

Dschee:
This is my attempt to fix #93 – I've setup a **new stencil template named "structured"** and improved the **tests to cover these cases**, too.

Unfortunately with enums it **isn't possible** to use the `L10n.Settings.NavigationBar.Title` dot syntax structure in a clean way. If this is required then we need to basically rewrite a lot of code as we would probably need to use Structs for that (that's is how [Laurine](https://github.com/JiriTrecak/Laurine) works). But given we have enums I tried to still get the advantage of scoped autocompletion and succeeded, the structure now looks like this: `L10n.Settings(.NavigationBar(.Title))`.

The **typing behavior is as good as with normal dot syntax**:

- type "L10n."
- hit down key until autocompletion finds the right key
- hit enter
- press "." and hit down key until found (via autocompletion)
- hit enter
- press "." and hit down key until found (via autocompletion)
- hit enter
- ...

Note that I have **not changed any of the existing features**, so releasing this should be easy as it just adds a new feature. Sometime later in a newer version we could of course use the dot syntax by default but as this will mean people using SwiftGen already will need to update their code this is not a good approach as a first step.

Also there will probably be some **edge cases** we will encounter once this is used in production (I tested in one of my own projects and it worked like a charm – no issues for me). As an example I think using "strings" as part of a key path like "settings.navigation_bar.strings.title" will be a problem as it is a reserved key. But I think these issues can be fixed over time and don't prevent from releasing a new version so long as this feature is opt-in (which it is right now).

Note that I couldn't find a way to do deep looping within the stencil so I needed to do this manually. The downsides of this approach are that there is a **maximum level of depth** (currently 5 dots) and that the stencil repeats a lot of code 5 times. I'm not sure if this is even possible in a different way when using the Stencil framework.

I also think it would be good to **document how to use this feature** somewhere – both, how to change up the commands to generate structured localization enums and how to use it in code. I don't have time right now for this though, just wanted to complete the feature and push it to get some feedback - here it is. :)


AliSoftware:
Cool thanks! 👍 
I'll try to take a look soon (even if I have a busy week but will do my best).

I think maybe we could address this dot-support feature in multiple steps.

1. The first step being this very PR, providing basic initial support with a single template based on enums, like you did.
2. Propose an alternate template that use `struct` instead of `enum` so we can use `x.y.z.t` syntax instead of enums with nested associated values `.x(.y(.z(.t)))`. It would be up to the use to choose the template they prefer
3. Try to implement it using recursion:
  * either using [Stencil's `include` tag](https://github.com/kylef/Stencil#include) so that the part being recursively repeated could be in a separate file, the root template include it at some point, and the included file could include itself recursively then? (just an idea, not sure it can really work by probably worth a try).
  * Or implement it using a custom tag (maybe [the custom `set` node I added in GenumKit to extend Stencil](https://github.com/AliSoftware/SwiftGen/blob/master/GenumKit/Stencil/SetNode.swift) could help? Basically it's used to store a part of a stencil expression into a variable and re-execute (re-render) it each time — it can be seen a bit like a local function that you then call when you need it except it doesn't take any argument, but it re-interpret its content each time so can re-interpret inner variables so it might be enough. It potentially could help (in a similar way Stencil's include tag would work, except it would be all in the same file), but again not totally sure it covers everything until we try
   * Or lastly if `SetNode` doesn't cover what we need, and doesn't allow us to implement it the way we want, we could consider creating a new custom Stencil node that would let us implement it the way we want

So we are at step 1 with this PR and I think I'll merge it soon once reviewed, we can still imagine steps 2 and 3 as future improvements for future versions, but let's keep it one step at a time

AliSoftware:
PS: Please add an entry in the `CHANGELOG.md` to explain your changes and credit yourself :wink:

templates/strings-structured.stencil
```diff
@@ -0,0 +1,233 @@
+// Generated using SwiftGen, by O.Halligon — https://github.com/AliSoftware/SwiftGen
+
+{% if structuredStrings %}
```
AliSoftware: It seems like you fixed an issue when previously using `if structuredStrings.strings` in previous commit (5aec504abf8aba42b08129750e68a087759ded03)… but the travis build before this fixing commit was still successful when you had that.

So maybe, if with the old code you had an issue and not the expected generated output, but the tests did still pass at that time despite the issue… that means that we are missing a Unit Test to cover that use case and prevent that this issue reappear in the future?

Dschee: Good catch, just added a test that ensures the expected behavior works in cases where there is content within deeper levels but none at the root level (that was the issue). See the message description in 9b3e7d66ea4c2a9f55de07e99cc918a1db8b7b57 for further details.

Dschee:
Sounds all reasonable to me, also added an entry to the `Changelog.md` and wrote an additional test to ensure everything works fine when no keys are specified in root level.

AliSoftware:
Cool!

Seems like travis hand and failed to report the CI status, and the branch seems to have conflicts anyway (probably just the CHANGELOG.md entries because I merged other requests in the meantime), preventing me to merge. Could you please rebase your branch on top of master so I can merge the PR? Thx!

UnitTests/fixtures/LocStructuredOnly.strings
```diff
@@ -0,0 +1,3 @@
+"settings.navigation-bar.title.even.deeper" = "Settings";
+"SETTINGS.USER_PROFILE_SECTION.HEADER_TITLE" = "User Profile Settings";
+"seTTings.uSer-proFile-sectioN.footer_text" = "Here you can change some user profile settings.";
```
AliSoftware: Could be interesting to add some examples with parameters, to ensure that template properly handle Localized strings with placeholders. Like:

```
"settings.user_profile_section.points" = "Welcome %@! You have %d points on your account.";
```

AliSoftware: Also wondering what happens if a structure happens to be both a leaf and a node of the structure, like:

```
"settings.title" = "Titre of section";
"settings.title.details" = "Subtitle";
```
In that case wouldn't there be some conflict like inside `enum SettingsL10n` you'd have both an case for `title` and a subenum for `title` containing `detail`?

I guess that will work ok because if I understand correctly, the names inside `enum SettingsL10n` will be `case title` and `enum TitleL10n` so they won't conflict, but it would be nice to add such a use case in that `LocStructuredOnly.strings` file to be sure we handle it properly without conflicts.

Dschee:
Rebased.

UnitTests/fixtures/LocStructuredOnly.strings
Dschee: As you can see in `Strings-File-Structured.swift.out` the result looks like this:

```
case Settings(SettingsL10n)
enum SettingsL10n {
```

As you can see there is also a **case** names `Settings` so if there was also a leaf names as such then this would lead to two entries techincally rendering the result as invalid code. If we want nodes to be autocompleting with the same name I think there's no way to fix this so long as we use enums. We could of course choose one of the two – but that would lead to confusions on the user side so I think it's best to actually keep it the way it is and document well. Long term we can fix this by using one of the other approaches.

AliSoftware: Ok this will probably be ok for our first step then, even if it doesn't cover _all_ the cases and generates invalid code if we have leaf and node with the same name, that's still a great start for all those cases where we don't have this special case, so I'm ok for a first version.

Would sill be interested to add a `"settings.user_profile_section.points" = "Welcome %@! You have %d points on your account.";` line in `LocStructuredOnly.strings` and ensure the generated code looks as expected though (as looking at the code it seems it would work and be handled by your template, so let's include that in the Unit Test :wink:)

Dschee: Also just added a parameterized and structured key to the tests.

UnitTests/expected/Strings-File-Structured-Only.swift.out
```diff
@@ -27,6 +27,8 @@ enum L10n {
       case HeaderTitle
       /// Here you can change some user profile settings.
       case FooterText
+      /// Welcome %@! You have %d points on your account.
+      case Points(String, Int)
```
AliSoftware: 🎉

UnitTests/fixtures/LocStructuredOnly.strings
Dschee: Haha, seems like you were writing while I was coding. ;)

AliSoftware: Hehe indeed 😆 
Will merge as soon as Travis validates the PR 😉 

Great job anway, thanks for the awesome work! 👍
_(I'm looking forward to see where we'll go from here to improve it even better in future versions 😉 )_

UnitTests/expected/Strings-File-Structured-Only.swift.out
Dschee: I hope it's all there now. You're probably just waiting for the tests to pass.

I've got one question: When do you plan a new release so people can actually use this feature? I'd appreciate a quick 1.2.0 release but you may have other plans of course.

AliSoftware: I was planning to do a release as soon as your PR + #131 were merged, which should be tonight 🎉

UnitTests/fixtures/LocStructuredOnly.strings
Dschee: Thanks for reviewing and helping to improve the PR so quickly. 👍 

Yeah, I'm also interested in what parts of SwiftGen will become obsolete by Xcode 8 – if any. If I remember correctly than Xcode 8 will include some kind of new image and colors within code. Haven't tested though, we're gonna see how that will improve working with resources.

AliSoftware:
@Dschee As explained in the CONTRIBUTING.md file you are helping me put together, I just given you push access to SwiftGen!

This means that you can directly use the repository (instead of keeping using your fork and having to keep it up-to-date at all times — now you can remove your fork and work on SwiftGen directly!) and create branches on the SwiftGen repository, add labels on issues and PR, close issues, etc.
Master is still locked to force us ensure that travis builds our changes and pass to green before we add anything on master, but otherwise you can push on any branch you create on SwiftGen.

Dschee:
Thanks. I'm uploading my contributing branch from my fork then directly to this repo. That'll probably close the MR so I'll have to create a new one then, right? Or is there any trick for such things? :)

AliSoftware:
Sadly GitHub doesn't allow to _modify_ an already open PR to point to a different fork or branch, so I think you'll have to create a new one indeed.

Dschee:
Let's try, maybe it picks it up automagically ... a moment ...

Dschee:
Didn't work, created a new MR: #149.

PR #152, opened 2016-07-20
Fix `testWizardsStoryboardsWithSwift3()` test and `storyboards-swift3.stencil`

ikesyo:


AliSoftware:
Hi @ikesyo 

It seems that you changed the storyboard used for the tests (from `Anonymous.storyboard` to `Wizard.storyboard`) and fixed the template (`storyboards-swift3.stencil`) but didn't change the `expected/*` files accordingly?

I wonder how the test is passing on travis given that the template changed but the expected output wasn't modified? (Haven't the tools to test it locally here and check more in details myself, as I'm on holiday)

ikesyo:
The followings are the details for the issue:

- `storyboards-swift3.stencil` and `expected/Storyboards-Wizard-Swift3.swift.out` are added in #117.
    - __Those were correct at that time__, but those are not really tested since the test was broken: `testWizardsStoryboardsWithSwift3()` didn't test `Wizard.storyboard` but tested `Anonymous.storyboard` with `storyboards-swift3.stencil`.
- `isBaseViewController` was added to `StoryboardParser.stencilContext` in #131.
    - __`storyboards-swift3.stencil` was not fully updated in that PR__, but the test for the combination of the template and `Wizard.storyboard` didn't run, so __the template got broken silently__.

AliSoftware:
Gotcha! Thanks for the detailed explanation!

PR #153, opened 2016-07-26
[Storyboard] Support typed initial view controller

ikesyo:


AliSoftware:
Looks :ok_hand:, thanks for that PR!

Could you just also add an entry in the CHANGELOG.md to credit yourself please? Thx!

ikesyo:
Thanks! Updated ✨

AliSoftware:
Finally merged! Sorry it took so long.

I've also added you as contributor to the project to thank you for your recent PRs, that will soon be integrated in the upcoming release. Use those superpowers at your convenience :wink:

AliSoftware:
@ikesyo Someone just pointed out that there's an issue (completion error) with that new template in case in have a Storyboard with **no** `initialViewController` set. See #160.

Would be nice to have a Unit Test for that. The fix is probably as simple as adding a conditional `{% if … %}` in the templates to avoid generating the initialViewController code if the variable is empty.

ikesyo:
> @ikesyo Someone just pointed out that there's an issue (completion error) with that new template in case in have a Storyboard with no initialViewController set. See #160.
> 
> Would be nice to have a Unit Test for that. The fix is probably as simple as adding a conditional {% if … %} in the templates to avoid generating the initialViewController code if the variable is empty.

That is already handled [here](https://github.com/AliSoftware/SwiftGen/blob/0d49961257ba72dd9b95715c7e2ec93761765242/GenumKit/Parsers/StoryboardParser.swift#L54-L57), [here](https://github.com/AliSoftware/SwiftGen/blob/0d49961257ba72dd9b95715c7e2ec93761765242/GenumKit/Parsers/StoryboardParser.swift#L97), [here](https://github.com/AliSoftware/SwiftGen/blob/0d49961257ba72dd9b95715c7e2ec93761765242/GenumKit/Stencil/Contexts.swift#L86-L92) and by [the template checks](https://github.com/AliSoftware/SwiftGen/blob/0d49961257ba72dd9b95715c7e2ec93761765242/templates/storyboards-default.stencil#L52). The root cause for that the issue is that picking up a newer template and using it in an older SwiftGen is not a safe nor supported operation (use case), I think.

AliSoftware:
Good point on the fact that using a new template with an old version might not work given that the new version introduces new keys used by the template but not present in previous versions.
That's indeed probably the reason of the bug reported in #160

However I'm wondering if the `{% ifnot storyboard.initialScene.isBaseViewController %}` is sufficient to handle all the cases. If there is an initialScene, then both the case when this initial scene is a UIViewController or a custom VC are handled. But if there is _no_ initialScene, then I think `storyboard.initialScene` is absent and thus `storyboard.initialScene.isBaseViewController` is evaluated to false, and the `info` branch is executed, generated the code for `initialViewController` while there is none. So that `{% ifnot … %}` test might probably itself be wrapped in some `{% if storyboard.initialScene %}` conditional block to handle that third case, right?

And even if the template is already correctly handling all the cases, adding a Unit Test for that use case seems like a good idea anyway :wink:

ikesyo:
> But if there is no initialScene, then I think storyboard.initialScene is absent and thus storyboard.initialScene.isBaseViewController is evaluated to false, and the info branch is executed, generated the code for initialViewController while there is none. So that {% ifnot … %} test might probably itself be wrapped in some {% if storyboard.initialScene %} conditional block to handle that third case, right?

I think that the case is already tested through [`Dependency.storyboard`](https://github.com/AliSoftware/SwiftGen/blob/58747812ac579c0f6b0d5df46708ea08513db5e0/UnitTests/fixtures/Storyboards-iOS/Dependency.storyboard#L2), [`Storyboards-All-CustomName.swift.out`](https://github.com/AliSoftware/SwiftGen/blob/58747812ac579c0f6b0d5df46708ea08513db5e0/UnitTests/expected/Storyboards-All-CustomName.swift.out#L44-L51) and [`Storyboards-All-Default.swift.out`](https://github.com/AliSoftware/SwiftGen/blob/58747812ac579c0f6b0d5df46708ea08513db5e0/UnitTests/expected/Storyboards-All-Default.swift.out#L44-L51) in this exact pull request.

If a storyboard does not have `initialViewController`, the placeholder `InitialScene(objectID: nil, tag: "viewController", customClass: nil)` instance is created [here](https://github.com/AliSoftware/SwiftGen/blob/0d49961257ba72dd9b95715c7e2ec93761765242/GenumKit/Parsers/StoryboardParser.swift#L97) and that is converted to an `initialScene` dictionary of the following:

```swift
[
    "viewController": "UIViewController",
    "isBaseViewController": true
]
```

> And even if the template is already correctly handling all the cases, adding a Unit Test for that use case seems like a good idea anyway 😉

The use case (a storyboard which does not have `initialViewController`) is already tested as I said above 😉

AliSoftware:
Ok sorry, I totally missed that this was already part of unit tests.

So if I understand you (and the code) correctly, that means that if there's no initial scene, then the generated StencilContext is the same as if there was an initial scene being a `UIViewController`, right? Is that a good idea?
I mean, it works given how you did write the current template, because it happens that what you're interested into in your template is to test if there is an initial scene that happens to not be a base view controller, so you can have only one test to handle both cases… But maybe it wouldn't be that obvious for people wanting to create their own template that having no initial scene mean that `{% if initialScene %}` would still evaluate to true — I wonder if they wouldn't assume instead that not having an initial scene would lead to `{{ initialScene }}` to be empty?

ikesyo:
Thanks for your thought @AliSoftware! I'm now convinced that we should differentiate the case of "no initial view controller" with the case of `UIViewController` is the initial scene. I will tackle it.

PR #156, opened 2016-08-12
Support for fonts with multiple descriptors

chrisellsworth:
I ran into an issue where I tried to use the font "Avenir" which contains descriptors for "Light", "Medium", "Heavy", etc., and the output from `swiftgen` only contained the descriptor for "Book."

The intent of this change is to support all of the descriptors from a given font, rather than just the first.


AliSoftware:
Perfect, nice one 👌 

Could you just please add an entry in the CHANGELOG.md to mention that change and credit yourself? Thanks!

chrisellsworth:
Done. Thanks! Looks like there's now a conflict though. Do you prefer that I rebase my feature branch or make a merge commit, or will you make the merge on your side when pulling in the PR?

AliSoftware:
If you can rebase your PR on top of `master` that would be great, thanks!

AliSoftware:
Mmmmh idk what happened with Travis, it seems your rebase didn't properly trigger a new build like it should've. Maybe pushing a dummy new commit might help re-triggering the travis hook?!

GenumKit/Parsers/FontsFileParser.swift
```diff
@@ -35,16 +35,18 @@ public func == (lhs: Font, rhs: Font) -> Bool {
 // MARK: CTFont
 
 extension CTFont {
-  static func parseFontInfo(fileURL: NSURL) -> Font? {
+  static func parseFontInfo(fileURL: NSURL) -> [Font]? {
```
AliSoftware: Now that this method returns an array and not a single font, maybe we could return a non-optional array instead (and return an empty array on any `guard` failure)?

GenumKit/Parsers/FontsFileParser.swift
```diff
@@ -69,9 +71,9 @@ public final class FontsFileParser {
             continue
           }
           guard UTTypeConformsTo(uti, "public.font") else { continue }
-          guard let font = CTFont.parseFontInfo(file) else { continue }
+          guard let fonts = CTFont.parseFontInfo(file) else { continue }
```
AliSoftware: This way, this line simply becomes `let fonts = CTFont.parseFontInfo(file)` and this simplifies the possible states: instead of having to deal with two empty cases (`nil` and `[]`) representing similar "no font found in this file" failure, we would now only have one case to consider (the empty array `[]` case) and using `[].forEach { … }` won't generate any entry in the output anyway.

_… and here goes your dummy commit to force Travis-CI to re-trigger a build_ 😆

chrisellsworth:
Less code, I like it! And so does Travis 👍

AliSoftware:
🎉

PR #159, opened 2016-08-18
Struct-based structured keys for dot-syntax support

Dschee:
This is a continuation of the work started in #148. Specifically this PR attempts to implement step 2 which @AliSoftware stated as follows:

> Propose an alternate template that use struct instead of enum so we can use `x.y.z.t` syntax instead of enums with nested associated values `.x(.y(.z(.t)))`. It would be up to the use to choose the template they prefer

Now the following should be possible in usage:
``` Swift
L10n.Settings.NavigationBar.Title
```

What do you think about updating the README to explicitly mention this as an example alternate template? I'm sure quite some people will be interested in using this template in favor of the default template (even if it only supports a depth level of 5).


AliSoftware:
> What do you think about updating the README to explicitly mention this as an example alternate template? I'm sure quite some people will be interested in using this template in favor of the default template (even if it only supports a depth level of 5).

Totally.

TBH I'm even wondering if that support shouldn't be the default one (but it's always quite tough to decide to replace the default template, because every project out there using SwiftGen with that default template will have its generated code modified and the syntax to call `L10n.xx` modified (from `L10n.Settings_NavigationBar_Title` to `L10n.Settings_NavigationBar_Title`).

Or maybe releasing SwiftGen as a new *major* version update (instead of minor) would solve that dilemma? (I'm never sure what's the best play regarding semantic versioning on tools like SwiftGen, should I base my semantic version on API differences on the tool itself, or on the API differences on the code SwiftGen generates?)

Dschee:
In case we replace the default template we should definitely release a major version. Semantic Versioning 2.0 defines that major versions should be released if the API changes in a breaking way, which is the code in usage for us. So if the default code used by people doesn't work any more (which is the case if it changes from "L10n.A_B_C" to "L10n.A.B.C") then this sure is a reason to make a major release regarding to semver imho.

Dschee:
At first I even wanted to suppose to replace the default template myself. But then I changed my mind since I thought, you might want to wait until we support many levels of depth since some people using SwiftGen will have more than 5 and you don't want them to be irritated by the major update. They could of course use the older default then if they need to if we keep it in some way. That should be documented then in a migration guide.

AliSoftware:
Indeed, on second thoughts, maybe keep it as an alternate template at first. As long as it's clearly documented in the README to let people interested in this feature know it exists and incite them to use it!

Dschee:
So would you want to do the documentation yourself and merge this or should I add it to the PR?

AliSoftware:
If that's OK with you, I think it would make more sense for you to add it to the PR, this way you'll think about all the details (like the maximum number of levels the template supports) as you're the one who implemented it 😉 

Thx!

GenumKit/Stencil/Contexts.swift
```diff
@@ -151,7 +151,8 @@ extension StringsFileParser {
           "count": entry.types.count,
           "types": entry.types.map { $0.rawValue },
           "declarations": (0..<entry.types.count).map { "let p\($0)" },
-          "names": (0..<entry.types.count).map { "p\($0)" }
+          "names": (0..<entry.types.count).map { "p\($0)" },
```
AliSoftware: Code style: I just realized that with Swift 2 now we can use `entry.types.indices` rather than `0..<entry.types.count` :wink:

GenumKit/Stencil/Contexts.swift
```diff
@@ -151,7 +151,8 @@ extension StringsFileParser {
...
+          "typednames": (0..<entry.types.count).map { "p\($0): \(entry.types[$0].rawValue)" }
```
AliSoftware: You can probably use `enumerate` instead here, like `entry.types.enumerate().map { p\($0): \($1.rawValue) }`

UnitTests/expected/Strings-File-Dot-Syntax.swift.out
```diff
@@ -0,0 +1,70 @@
...
+
+  struct Settings {
+
+    struct NavigationBar {
+
+      struct Title {
+
+        struct Even {
+          /// Settings
+          static let Deeper = L10n.tr("settings.navigation-bar.title.even.deeper")
```
AliSoftware: We should probably add another string which has more than 6 levels, so that we can write a Unit Test for such cases, and ensure it doesn't generate code that fails to compile if people happen to use deeper levels (but rather generate code that just stops splitting/nesting at level 6 and use `_` for higher level… but would at least compile)

GenumKit/Stencil/Contexts.swift
Dschee: True, updated the code in 2bedddba9bfc304d619cfd9879562dffb5ecaeb8.

GenumKit/Stencil/Contexts.swift
Dschee: Nice catch, implemented in 8aa31e72fd1cf629a92b73e68513cd74ca1ec18b.

UnitTests/expected/Strings-File-Dot-Syntax.swift.out
Dschee: Very good idea. I just implemented this in 2f9512141edd98713e7b5090532125359ea73929.

Dschee:
@AliSoftware Thank you for your suggestions, I've walked through them all. Only thing left to do is documentation. Planning to do this sometime this weekend – and yes, next time I'll do my changes directly within the project. Forgot that I had access. ^^

Dschee:
I just added a section to the README. Anything else missing for a merge? :)

Dschee:
Ahh yeah, rebase missing, just a minute ...

Dschee:
Ok, rebased. Should be merge-ready now. :)

CHANGELOG.md
```diff
@@ -33,12 +33,13 @@
   [Olivier Halligon](https://github.com/AliSoftware)
 * Now reports an error when it failed to parse a color in a color input file.  
   [Olivier Halligon](https://github.com/AliSoftware), [#162](https://github.com/AliSoftware/SwiftGen/issues/162)
-
+* Add struct-based structured identifiers using dot syntax via `-t dot-syntax` option.
```
AliSoftware: Maybe specify that this is for the `strings` part of SwiftGen only? Like:

> * New Strings template (available via `-t dot-syntax`), allowing string keys containing dots
>    (like `foo.bar.baz`) to be organized as a hierarchy and accessible via dot syntax.

Pods/PathKit/Sources/PathKit.swift
```diff
@@ -673,6 +673,6 @@ internal func +(lhs: String, rhs: String) -> Path {
 
 extension Array {
   var fullSlice: ArraySlice<Element> {
-    return self[0..<self.endIndex]
+    return self[self.indices]
```
AliSoftware: Don't modify sources of dependencies!

AliSoftware:
Apart from the two little comments, seems good. Will let you fix those then probably merge it tomorrow 👍

CHANGELOG.md
Dschee: Done – simply copied your suggestion.

Pods/PathKit/Sources/PathKit.swift
Dschee: Oh, sorry for that, should have been more careful with search & replace. Fixed now in a288a41e900df5b16e60b2e6a279f85d885a17d2.

Dschee:
Fixed. :)

PR #163, opened 2016-08-30
[Storyboard] Handle `initialViewController` absence correctly

ikesyo:
Discussed in #153 and #160.

PR #165, opened 2016-09-03
Fix initialViewController #153 to be UIKit-agnostic

AliSoftware:
Previous PR #153 and #163 did hardcode the "UI" prefix before the `viewController`/`tableviewController`/… tag names, making the code impossible to use for the OS X templates.

Now the prefix is handled at the template level, so that the iOS templates add the `UI` prefix themselves.

When we have time, we could add that support for typed initial VC in the OS X templates as well now that the code SwiftGen core doesn't hardcode the "UI" prefix anymore. 

\cc @ikesyo


GenumKit/Stencil/Contexts.swift
```diff
@@ -83,13 +85,20 @@ extension StoryboardParser {
     let storyboards = Set(storyboardsScenes.keys).union(storyboardsSegues.keys).sort(<)
     let storyboardsMap = storyboards.map { (storyboardName: String) -> [String:AnyObject] in
       var sbMap: [String:AnyObject] = ["name": storyboardName]
+      // Initial Scene
       if let initialScene = initialScenes[storyboardName] {
-        var initial: [String:AnyObject] = ["viewController": initialScene.customClass ?? "UI\(uppercaseFirst(initialScene.tag))"]
```
AliSoftware: That was the problematic line, because we don't want the SwiftGen binary itself to add the `UI` prefix (so that we can also be compatible with OS X and its NS prefix) but want that to be handled at the template level (so that OS X templates could add the `NS` prefix)

Pods/Stencil/Sources/IfTag.swift
```diff
@@ -0,0 +1,80 @@
...
+
+  public func render(context: Context) throws -> String {
+    let result = try variable.resolve(context)
+    var truthy = false
+
+    if let result = result as? [Any] {
+      truthy = !result.isEmpty
+    } else if let result = result as? [String:Any] {
+      truthy = !result.isEmpty
+    } else if let result = result as? Bool {
```
AliSoftware: This very line was added recently in the `master` branch of the `Stencil` pod (isn't even present in `6.0.0-beta.1`, that's why we had to point onto `master`). Before that change, `{% ifnot initialScene.isBaseViewController %} wasn't evaluated correctly and generation using Stencil defaulttdmplates would fail to generate the expected result, as they rely on those kind of tests.

PR #166, opened 2016-09-04
Migrate Swift 3 templates to use lowercase enums

AliSoftware:
@Sajjon could you test this in your Swift 3 project and check all the templates work and generate expected code that compiles properly? Thx!

PR #167, opened 2016-09-04
Add Swift3 support to Dot-Syntax template

Dschee:
This is a followup of #166.


Dschee:
Oh wait, wrong order of branches. :D

PR #168, opened 2016-09-04
Add Swift3 support to Dot-Syntax template [fixed]

Dschee:
This is a followup of #166.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ## 3.0.0
 
-* Add template that calls `NSLocalizedString()` separately for each string, 
+* Add template that calls `NSLocalizedString()` separately for each string,
   which is useful when trying to extract strings in the app to a `.strings` file.  
   [Ahmet Karalar](https://github.com/akaralar)
 * Add some `file_length` and similar SwiftLint exceptions in bundled templates, as
@@ -40,6 +40,8 @@
   [Cihat Gündüz](https://github.com/Dschee), [#159](https://github.com/AliSoftware/SwiftGen/pull/159)
 * Update Swift 3 templates to use lowercase enums.  
   [Olivier Halligon](https://github.com/AliSoftware), [#166](https://github.com/AliSoftware/SwiftGen/pull/166)
+* New Strings template (available via `-t dot-syntax-swift3`), allowing keys with dots in Swift 3 (see above).
+  [Cihat Gündüz](https://github.com/Dschee), [#168](https://github.com/AliSoftware/SwiftGen/pull/168)
 
 > 💡 You can now **create your custom templates more easier than ever**, by cloning an existing template!
 >
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,7 @@
   [Cihat Gündüz](https://github.com/Dschee), [#159](https://github.com/AliSoftware/SwiftGen/pull/159)
 * Update Swift 3 templates to use lowercase enums.  
   [Olivier Halligon](https://github.com/AliSoftware), [#166](https://github.com/AliSoftware/SwiftGen/pull/166)
-* New Strings template (available via `-t dot-syntax-swift3`), allowing keys with dots in Swift 3 (see above).
+* New Strings template (available via `-t dot-syntax-swift3`), allowing keys with dots in Swift 3 (see above).  
   [Cihat Gündüz](https://github.com/Dschee), [#168](https://github.com/AliSoftware/SwiftGen/pull/168)
 
 > 💡 You can now **create your custom templates more easier than ever**, by cloning an existing template!
--- a/UnitTests/expected/Strings-File-Dot-Syntax-Swift3.swift.out
+++ b/UnitTests/expected/Strings-File-Dot-Syntax-Swift3.swift.out
@@ -1,0 +1,79 @@
+// Generated using SwiftGen, by O.Halligon — https://github.com/AliSoftware/SwiftGen
+
+import Foundation
+
+// swiftlint:disable file_length
+// swiftlint:disable type_body_length
+// swiftlint:disable nesting
+// swiftlint:disable variable_name
+// swiftlint:disable valid_docs
+
+struct L10n {
+  /// Title of the alert
+  static let AlertTitle = L10n.tr("alert_title")
+  /// Some alert body there
+  static let AlertMessage = L10n.tr("alert_message")
+  /// Hello, my name is %@ and I'm %d
+  static func Greetings(p0: String, p1: Int) -> String {
+    return L10n.tr("greetings", p0, p1)
+  }
+  /// These are %3$@'s %1$d %2$@.
+  static func ObjectOwnership(p0: Int, p1: String, p2: String) -> String {
+    return L10n.tr("ObjectOwnership", p0, p1, p2)
+  }
+
+  struct Apples {
+    /// You have %d apples
+    static func Count(p0: Int) -> String {
+      return L10n.tr("apples.count", p0)
+    }
+  }
+
+  struct Bananas {
+    /// Those %d bananas belong to %@.
+    static func Owner(p0: Int, p1: String) -> String {
+      return L10n.tr("bananas.owner", p0, p1)
+    }
+  }
+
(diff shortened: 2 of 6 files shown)


AliSoftware:
Don't forget the CHANGELOG entry 😉

Dschee:
Uuh, this "update branch" button ... :(

Okay, sure.

Dschee:
Okay, @AliSoftware now it should be ready to merge (once the checks pass). :)

Dschee:
@AliSoftware I just noticed that a key like `OBJECT.NUMBER.NAME` is now `L10n.Object.Number.NAME` in usage, not `L10n.Object.Number.Name` like it was before #166 was merged.

AliSoftware:
Yes, that's what I explained here https://github.com/AliSoftware/SwiftGen/pull/166#issuecomment-244611672

You can't satisfy every user convention, both people writing `URL.Name`, `URL_NAME` and `url_name` and treat/filter all those correctly for all cases.

One solution would be to make one template for people using all-uppercase keys and a different one for people using camel-case or dot-separated keys. That would probably be too much templates, but after all, to each their own. I think for 3.0.0 this one will be enough, 3.1.0 has plenty of room for a lot of new templates (like dot-syntax for macOS, templates for people using all-uppercase keys, …) to make everybody happy.

---

That would also mean that it would be great to have a markdown file or a wiki page summarizing the description of each template, so that people have an idea of which one they should choose amongst that many available templates and have an idea of what the conventions associated for each template.

Dschee:
Yeah, I actually read that but didn't quite understand it when the `dot-syntax` template is considered. That template was created with the goal to satisfy the need of all people using a dot in their keys and result to a camel cased dot-separated structure.

So, taking your examples `URL.Name` should be `L10n.Url.Name`, `URL_NAME` should be `L10n.UrlName` and `url_name` should be `L10n.UrlName` also. This is what I tried to create so I wonder how come you changed the behavior without that anyone seemed to need that change?

Maybe you changed a method which had this side effect and you decided to go for the change elsewhere and live with the broken template? If that's the case I'm completely okay with that, I just need to know why you changed this, ;)

Dschee:
Rebased.

AliSoftware:
I did the change because of the use case (that we forgot in our unit tests) of `ObjectOwnership` (camel-cased key) in other templates, that generated odd all-lowercase result, and I feel like using `ObjectOwnership` is more common than using `OBJECT_OWNERSHIP` anyway.

But in dot-syntax template, I'm sure there's a way to avoid that: you should try playing with different filters to achieve that, I'm sure its possible but it's hard to make it work for all people's conventions. Maybe I did it in a rush and applied it everywhere including dot-syntax where we could do better, but 3.0 is long overdue.

So what we're probably gonna do is to merge this PR and include it in 3.0.0 as is, because the `dot-syntax` template is quite cool and I think this is gonna be the future default later… but then you should try to improve the template to avoid that all-uppercased `NAME` suffix but also still keep compatibility with all conventions (both `URL_NAME`, `URLName`, `URL.Name`, `URL.Name_And_Params`, `URL.NAME-AND_PARAMS` and whatnot), and I'll merge that fix after 3.0.0 and include it in a 3.1.0 once it's ready (even if that means releasing again as soon as next week), because doing last-minute changes before a rushed release is never a good idea.

Because 3.0.0 is 1 month overdue already and it's ages I haven't done any new release (people are waiting for one to be able to use SwiftGen with the latest Swift 3 changes) and we can't let users wait any longer (especially since I'll be away to give some talks in upcoming weeks and probably won't have time to release later this month). I prefer releasing often with small changes rather than making people wait too long, and that 3.0.0 was definitely too long given the number of changes it contains already.

CHANGELOG.md
```diff
@@ -40,6 +40,8 @@
   [Cihat Gündüz](https://github.com/Dschee), [#159](https://github.com/AliSoftware/SwiftGen/pull/159)
 * Update Swift 3 templates to use lowercase enums.  
   [Olivier Halligon](https://github.com/AliSoftware), [#166](https://github.com/AliSoftware/SwiftGen/pull/166)
+* New Strings template (available via `-t dot-syntax-swift3`), allowing keys with dots in Swift 3 (see above).
```
AliSoftware: Please end line with two spaces to avoid breaking Markdown formatting

Dschee: Oh sure, sorry. On it!

Dschee:
That's all okay with me, I was not to say "stop the release, this is more important!" but instead just wanted to understand the change. Now I have and in fact I would have done the same, so no problem. ;)

I was also waiting for a new release since some time, so I'm totally happy with the current 3.0.0! 🎉

AliSoftware:
One thing that seems strange to me in the template `dot-syntax-swift3` you just added is I'm not sure I see any difference with the Swift 2 template? Especially, the generated constants still are titlecased instead of lowercased.

Agreed, that's arguable as we actually generate `struct` in this structured template, so using `Bananas` and `Owner` titlecased makes sense if you see them as being type names:

```swift
  struct Bananas {
    /// Those %d bananas belong to %@.
    static func Owner(p0: Int, p1: String) -> String {
      return L10n.tr("bananas.owner", p0, p1)
    }
  }
```

But at call site, that will look like:

```swift
L10n.Bananas.Owner(2, "John")
```

While this is correctly following Swift 3 conventions as `Bananas` and `Owner` are type names, it seems to me that at call site it could make more sense to use lowercase to make it more Swift-3:

```swift
L10n.bananas.owner(2, "John")
```
But if that would make the `struct` names lowercase themselves… so I'm a bit torn here.

What's your thoughts? Should we turn them lowercase to improve call-site, or keep them titlecase to respect Swift 3 even if it's implementation details for the user?

Dschee:
Sorry for getting back so late, a lot of things to do lately. :)

Well, I think you've got a point there. I'm just thinking about this: What if we keep all structs titlecased and only lowercase the functions? That way we would still follow the conventions and communicate to the caller that `L10n.Bananas` is not yet going to return a `String` as the last components is not lowercased. The final call would look like this:

```Swift
L10n.Bananas.owner(2, "John")
```

What do you think?

AliSoftware:
LGTM, good idea!
Le lun. 3 oct. 2016 à 14:37, Cihat Gündüz <notifications@github.com> a
écrit :

zzyspace:
I used dot syntax like:
```swift
L10n.Bananas.Owner(p0: 2, p1: "John")
```

how to use it like:
```swift
L10n.Bananas.Owner(2, "John")
```

Dschee:
Good question, it wasn't `p0: 2, p1: "John"` when this PR was merged afaik. Must have been a side effect of other changes. Do you have some more insight @AliSoftware?

AliSoftware:
For the same function declaration, Swift 3 now makes the parameter names visible to the call site. The template must state `case owner(_ p0: Int, _ p1: String)` instead of `case owner(p0: Int, p1: String)` to fix that for Swift 3.

I haven't checked in detail how the `dot-syntax-swift-3.stencil` template was implemented in that regard, maybe we have some changes to do to this template? (That should have been covered by the Unit Tests though…)

UnitTests/expected/Strings-File-Dot-Syntax-Swift3.swift.out
```diff
@@ -0,0 +1,79 @@
...
+  struct Apples {
+    /// You have %d apples
+    static func Count(p0: Int) -> String {
+      return L10n.tr("apples.count", p0)
+    }
+  }
+
+  struct Bananas {
+    /// Those %d bananas belong to %@.
+    static func Owner(p0: Int, p1: String) -> String {
```
AliSoftware: This is where the problem lies. We should adapt the template so that it should instead generate this:
```
static func owner(_ p0: Int, _ p1: String) -> String
```

Not sure how I missed that…

AliSoftware: Note that #237 refactored all those templates to use recursion and get rid of copy/pasting, so maybe wait for it to be merged to fix that Swift 3 syntax on the recursive version, that would be much easier.

zzyspace:
Great! Waiting for it!

PR #171, opened 2016-09-14
Feature/swift 2.3

### LINKED ISSUE #169: Convert project to Swift 3 and app bundle
- [X] Convert the SwiftGen project itself to Swift 3, so that we can continue developing it using Xcode 8 instead of Xcode 7 (at least convert it to Swift 2.3, but at this point we might as well migrate to Swift 3 directly) _(#201)_
- [x] Reorganize the project to that it's possible to Cmd-B in Xcode to build the project from Xcode, the same way SwiftLint does it. _(#204)_
  - This would build a command-line app wrapped in a `.app` bundle (which Xcode knows how to produce, including linking the Swift dynamic libraries etc) so we'd be able to debug it more easily, debugging it directly from Xcode using Cmd-R (and adjusting the command-line arguments via the Xcode scheme while debugging)
  - The `Rakefile` would still be necessary to reorganize the build `.app` into a package shippable via Homebrew (moving build dylibs around, changing their `install_name`, etc) so that the end user doesn't see a `.app` bundle with no UI but still sees a command-line tool

vknabel:
This pull request migrates SwiftGen to Swift 2.3 and would temporarily fixes #169 particularly.
A migration to Swift 3.0 would be much heavier and would take more time.

Currently, one can compile `swiftgen-cli` and all unit tests pass.
But I haven't checked the stencil templates yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: objective-c
-osx_image: xcode7.3
+osx_image: xcode8
 
 # cache: cocoapods
 # podfile: Example/Podfile
--- a/Pods/Pods.xcodeproj/project.pbxproj
+++ b/Pods/Pods.xcodeproj/project.pbxproj
@@ -945,6 +945,7 @@
 				PRODUCT_NAME = Pods_UnitTests;
 				SDKROOT = macosx;
 				SKIP_INSTALL = YES;
+				SWIFT_VERSION = 2.3;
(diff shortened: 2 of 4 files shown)


AliSoftware:
Seems cool 👌  will check that more thoroughly a bit later in the week when I've time to actually test the resulting project with the templates (don't see a reason why they would be affected)

ikesyo:
Superseded by #179. Thanks for the pull request though!

PR #176, opened 2016-09-24
Updates SwiftGen to Swift 2.3

dzenbot:
The automated changes on the project config are different than https://github.com/AliSoftware/SwiftGen/pull/171 for some reason. Maybe different Xcode versions? I used **Version 8.0 (8A218a)**, which is the latest.

Most importantly, the Rakefile is also updated to Xcode 8.* so running SwiftGen doesn't throw.

NOTE: I tried updating to Swift 3, but there's so many many migrations issues! Most of all are known issues in https://swift.org/migration-guide/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: objective-c
-osx_image: xcode7.3
+osx_image: xcode8
 
 # cache: cocoapods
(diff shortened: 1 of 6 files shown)


ikesyo:
Superseded by #179. Thanks for the pull request though!

PR #182, opened 2016-10-04
explicitly set pods to use legacy swift language version (i.e. 2.3)

AndrewSB:
I'm unsure how the license changes in this diff happened - I ran `pod update`, and then changed the `SWIFT_VERSION` flag on all the frameworks in the Pods.xcodeproj


ikesyo:
Thanks for the pull request, but you need CocoaPods 1.1.0.rc.2 for Xcode 8. See https://github.com/AliSoftware/SwiftGen/pull/179#issuecomment-251291172. 🙏

PR #188, opened 2016-10-21
Adding code gen

plivesey:
I've mostly finished all the code for the initial commit. But need some advice/help/more work on a few things:

1. I haven't been able to test end to end yet. But rebased on master, and should be able to try that now.
2. I want to output static files as part of the code gen. I can't work out how to do this.
3. I want to allow for arbritrary parameters on the command line and parse them into the stencil dictionary (as keys and values). Is this possible? I can't work out how to do this.
4. There's only one test right now which tests all the features in one model. Is this OK? Or do you want more testing?
5. TODO: Needs docs.

Input:

```
{
  "type": "UnitTestClass",
  "fields": [
             {"type": "string", "optional": true, "name": "firstName", "doc": "First name of the unit test"},
             {"type": "string", "name": "lastName"},
             {"type": "array", "element": "UnitTestClass", "name": "friends"},
             {"type": "array", "element": "UnitTestClass", "name": "optFriends", "optional": true, "doc": "An optional array"},
             {"type": "int", "name": "age", "optional": true}
             ]
}
```

Output:

Usage:
```
// Can initialze with throws init
do {
  let model = try UnitTestClass.make(dictionary: myDictionary)
} catch {}

// Can init with an optional init
let model: UnitTestClass? = UnitTestClass(dictionary: myDictionary)

// Can init with an optional init and an error reporter closure
let model: UnitTestClass? = UnitTestClass(dictionary: myDictionary) { error in
  print(error)
}
```

```

public final class UnitTestClass: DataModel, Equatable {

  // MARK: Instance Variables

  /**
   First name of the unit test
   */
  public let firstName: String?
  public let lastName: String
  public let friends: [UnitTestClass]
  /**
   An optional array
   */
  public let optFriends: [UnitTestClass]?
  public let age: Int?

  // MARK: Initializers

  public init(firstName: String?, lastName: String, friends: [UnitTestClass], optFriends: [UnitTestClass]?, age: Int?) {
    self.firstName = firstName
    self.lastName = lastName
    self.friends = friends
    self.optFriends = optFriends
    self.age = age
  }

  public convenience init?(dictionary: [String: Any]) {
    try? self.init(dictionary: dictionary, error: { _ in })
  }

  public convenience init?(dictionary: [String: Any], error: ((String) throws -> Void)) rethrows {
    try? self.init(dictionary: dictionary as NSDictionary, error: error)
  }

  public convenience init?(dictionary: NSDictionary) {
    try? self.init(dictionary: dictionary, error: { _ in })
  }

  public init?(dictionary: NSDictionary, error: ((String) throws -> Void)) rethrows {
    let parser = Parser(dictionary: dictionary)
    self.firstName = try parser.fetch("firstName", error: error)
    guard let lastName: String = try parser.fetch("lastName", error: error) else {
      try error("Required field 'lastName' is missing from type 'UnitTestClass'")
      return nil
    }
    self.lastName = lastName
    self.friends = try parser.fetch("friends", error: error)
    self.optFriends = try parser.fetch("optFriends", error: error)
    self.age = try parser.fetch("age", error: error)
  }

  public static func make(withDictionary dictionary: [String: Any]) throws -> Self {
    return try make(withDictionary: dictionary as NSDictionary)
  }

  public static func make(withDictionary dictionary: NSDictionary) throws -> Self {
    let model = try self.init(dictionary: dictionary) { error in
      throw NSError(domain: "com.swiftgen.modelParsing", code: 0, userInfo: [ "parseMessage": error ])
    }
    if let model = model {
      return model
    } else {
      throw NSError(domain: "com.swiftgen.modelParsing", code: 0, userInfo: [ "parseMessage": "Model returned nil with no error." ])
    }
  }

  public static func parse(data: Any, error: (String) throws -> Void) rethrows -> UnitTestClass? {
    if let data = data as? NSDictionary {
      return try self.init(dictionary: data, error: error)
    } else {
      try error("Models must be initialized with NSDictionary. Instead, got \(type(of: data))")
      return nil
    }
  }

  // MARK: Data

  public func data() -> [String: Any] {
    var data = [String: Any]()
    data["firstName"] = firstName?.dataRepresentation()
    data["lastName"] = lastName.dataRepresentation()
    data["friends"] = friends.map { $0.dataRepresentation() }
    data["optFriends"] = optFriends?.map { $0.dataRepresentation() }
    data["age"] = age?.dataRepresentation()
    return data
  }

  // MARK: Builder

  public func map(update: (Builder)->()) -> UnitTestClass {
    let builder = Builder(copy: self)
    update(builder)
    return builder.build()
  }

  public final class Builder {
    public var firstName: String?
    public var lastName: String
    public var friends: [UnitTestClass]
    public var optFriends: [UnitTestClass]?
    public var age: Int?

    public init(copy: UnitTestClass) {
      firstName = copy.firstName
      lastName = copy.lastName
      friends = copy.friends
      optFriends = copy.optFriends
      age = copy.age
    }

    public func build() -> UnitTestClass {
      return UnitTestClass(firstName: firstName, lastName: lastName, friends: friends, optFriends: optFriends, age: age)
    }
  }
}

public func ==(lhs: UnitTestClass, rhs: UnitTestClass) -> Bool {
  return lhs.firstName == rhs.firstName && lhs.lastName == rhs.lastName && lhs.friends == rhs.friends && lhs.optFriends == rhs.optFriends && lhs.age == rhs.age
}
```


plivesey:
Edited the previous comment with the new changes. I think this is ready for review now and merge once I've worked out the problems listed above.

plivesey:
FYI, I also tested this on https://github.com/bwhiteley/JSONShootout for performance and this solution was faster than any solution there. I'll create a pull request for the repo once I've merged in these changes.

AliSoftware:
I'm not sure it makes sense to see that new feature as a command specifically to generate models, rather than seeing it as a generic JSON input parser? And make the command `swiftgen json -t model …` instead of `swiftgen model …`?

Haven't had time to review _all_ the code, but I'm not sure to see why your code would only be tied to model generation and not arbitrary content generation?

plivesey:
So, this code takes JSON as an input and produces models. Naming it JSON seems weird, because that's the input not the output. It seems like all the other commands are based on the output. As in:

```
swiftgen images ./myImagesFile
```

Will produce an enum of images. Possibly, you're saying this should be able to create anything? Which is sort of true I suppose...but by default, it produces models (if you don't specify a template), so seems more user friendly to name it model?

plivesey:
Update on original issues and @AliSoftware's comment:

1. Tested end to end. It works!
2. Added a `base` command which outputs static code
3. I want to allow for arbritrary parameters on the command line and parse them into the stencil dictionary (as keys and values). Is this possible? I can't work out how to do this. (Don't need this for this commit. Will do this in a future commit.
4. The test seems to exersize all the code paths I need to test
5. Added docs

I think it's ready for review and merge now :D

GenumKit/Stencil/Filters.swift
```diff
@@ -80,12 +80,32 @@ struct StringFilters {
...
+    } else if let array = value as? [String] {
+      strings = array
+    } else {
+      throw FilterError.InvalidInputType
+    }
 
     return strings.joinWithSeparator(", ")
   }
+
+  static func joinAnd(value: Any?) throws -> Any? {
```
AliSoftware: It has been a while since I checked how Stencil evolved and what new features were brought by more recent versions, but I remember that @kylef wanted to implement filters with parameters. In case that already landed we could use that instead of defining a dedicated filter for each joiner string we need!

GenumKit/Stencil/Filters.swift
```diff
@@ -106,3 +126,83 @@ struct NumFilters {
     return "\(percent)%"
   }
 }
+
+struct ModelParsingFilters {
```
AliSoftware: Aren't all those too specific? Not sure why we can't implement that in the template instead?

swiftgen-cli/base.swift
```diff
@@ -0,0 +1,28 @@
...
+//  Created by Peter Livesey on 11/10/16.
+//  Copyright © 2016 AliSoftware. All rights reserved.
+//
+
+import Commander
+import PathKit
+import GenumKit
+import Stencil
+
+let baseCommand = command(
```
AliSoftware: Why the need for that command?

`swiftgen templates cat json-base` already covers outputting the verbatim content of a template…

GenumKit/Stencil/Filters.swift
plivesey: So, this took me the longest time to work out. AFAIK, it is impossible to implement them in the template because of the limitations of Stencil. To take them one by one:

plivesey: `parameterList`
This one needs to create a list of params like: `x: String, y: Int`
The main problem here is that there is no trailing `,`. This means it's impossible to do in a template alone. So, I decided to use `join`! But `join` takes an array as a parameter, so I need to convert my input to an array. This is impossible to do in stencil. Basically, the main functionality which stencil seems to be missing is a map filter. Sadly, you can't just implement a `map` componenet, because component's are rendered. You need a `map` filter. But, since filters can't take parameters, you need to implement seperate filters for each `map`.

plivesey: `equalityList`. Basically, the same problem as above. Need to create an array of `==` statements, and there is no `map` function.

plivesey: `initializerList` yep same thing here too

plivesey: The main concern here is that this will grow as people add their own templates. This is unfortunate, but sadly, Stencil just really doesn't support lists well because the lack of a `map`  (oh, how I love `map`). As people more methods details, we should be able to change these up to be as generic as we need. I tried to keep stuff as generic as possible creating lists for joining and not just outputting strings.

swiftgen-cli/base.swift
plivesey: Ah. I didn't realize that. Perfect.

plivesey: Pushed

GenumKit/Stencil/Filters.swift
kylef: I've implemented this in master now, you can find an example of implementing a filter with arguments at https://github.com/kylef/Stencil/commit/dfd57e957125e5178bf4e9d401fa892992783306.

It's going to be a few days until I cut a release with this though, there are other things I want to get into this release.

plivesey: @kylef sadly, I can't upgrade stencil because SwiftGen inherits some classes to modify some behavior. Since things aren't open, this won't compile.

Take a look at https://github.com/AliSoftware/SwiftGen/blob/master/GenumKit/Stencil/GenumTemplate.swift

Template and Namespace are both subclassed here. The code seems reasonable to me (so wondering if you'd consider making these open), but wondering if you have other ideas.

plivesey: Is the problem that Template isn't actually `open`? Because I'm confused. For me on 0.7.1:

`public class Template: ExpressibleByStringLiteral`
https://github.com/kylef/Stencil/blob/master/Sources/Template.swift

plivesey:
@AliSoftware I've updated to use a filter with an argument instead of join&&.
Also rebased on master and migrated to swift 3

There is one outstanding issue: I had to manually edit the stencil pod to make template open. Waiting on @kylef to see what to do about that.

AliSoftware:
Ok, cool, thx!

I'm adding the WIP label to wait for #219 to be merged and Stencil to be updated in SwiftGen, so we could then rebase this PR on top of the changes (that would make your PR cleaner and easier to review if it doesn't contain the Stencil pod update along the rest of the changes) and review it then 😉

kylef:
> There is one outstanding issue: I had to manually edit the stencil pod to make template open. Waiting on @kylef to see what to do about that.

I'll make a release (0.7.2) which makes Template open soon and update my PR.

GenumKit/Stencil/GenumTemplate.swift
```diff
@@ -41,20 +41,24 @@ open class GenumTemplate: Template {
...
+  namespace.registerFilter("join", filter: ArrayFilters.join)
+  namespace.registerFilter("lowerFirstWord", filter: StringFilters.lowerFirstWord)
+  namespace.registerFilter("snakeToCamelCase", filter: StringFilters.snakeToCamelCase)
+  namespace.registerFilter("snakeToCamelCaseNoPrefix", filter: StringFilters.snakeToCamelCaseNoPrefix)
+  namespace.registerFilter("titlecase", filter: StringFilters.titlecase)
+  namespace.registerFilter("hexToInt", filter: NumFilters.hexToInt)
+  namespace.registerFilter("int255toFloat", filter: NumFilters.int255toFloat)
+  namespace.registerFilter("percent", filter: NumFilters.percent)
+  namespace.registerFilter("escapeReservedKeywords", filter: StringFilters.escapeReservedKeywords)
+  namespace.registerTag("ifstrequal", parser: StrEqualsNode.parse)
```
kylef: I don't think this tag or the one below would be necessary with Stencil 0.7, you can use the new `if` statement that takes operators. For example:

```html+django
{% if something == "example" %}

{% endif %}
```

```html+django
{% if something != "example" %}

{% endif %}
```

plivesey:
@AliSoftware I've rebased and made the suggested changes. There are no Pods diffs anymore.
Thanks @kylef for the great additions to stencil.

AliSoftware:
Cool thx! 👌  Not at home this week-end, but will take a look later!

GenumKit/Stencil/Filters.swift
```diff
@@ -114,12 +114,26 @@ extension StringFilters {
...
+      strings = array.flatMap { $0 as? String }
+      guard array.count == strings.count else { throw FilterError.invalidInputType }
+    } else if let array = value as? [String] {
+      strings = array
+    } else {
+      throw FilterError.invalidInputType
+    }
 
-    return strings.joined(separator: ", ")
+    let separator: String
```
AliSoftware: Shorter version: `let separator = (arguments.first as? String) ?: ", "`

GenumKit/Stencil/Filters.swift
```diff
@@ -114,12 +114,26 @@ extension StringFilters {
 }
 
 struct ArrayFilters {
-  static func join(_ value: Any?) throws -> Any? {
-    guard let array = value as? [Any] else { throw FilterError.invalidInputType }
-    let strings = array.flatMap { $0 as? String }
-    guard array.count == strings.count else { throw FilterError.invalidInputType }
+  /// Joins an array of values with a separator. By default, it uses `, `
+  /// but you can pass in a different value as an argument.
+  static func join(value: Any?, arguments: [Any?]) throws -> Any? {
```
AliSoftware: @kylef Wouldn't it be useful to integrate such a filter in Stencil itself in the future?

GenumKit/Stencil/MapNode.swift
```diff
@@ -0,0 +1,63 @@
...
+      let error = "'map' statements should use the following " +
+      "'map array with element set varname' `\(token.contents)`."
+      throw TemplateSyntaxError(error)
+    }
+
+    let variable = components[1]
+    let mapVariable = components[3]
+    let resultName = components[5]
+
+    let forNodes = try parser.parse(until(["endmap", "empty"]))
```
AliSoftware: `let mapNodes` (or some more appropriate name instead) :wink:

GenumKit/Stencil/MapNode.swift
```diff
@@ -0,0 +1,63 @@
...
+    self.mapVariable = mapVariable
+    self.resultName = resultName
+    self.nodes = nodes
+  }
+
+  public func render(_ context: Context) throws -> String {
+    let values = try variable.resolve(context)
+
+    if let values = values as? [Any], values.count > 0 {
+      let mappedValues: [String] = try values.enumerated().map { index, item in
```
AliSoftware: Why do we need the `enumerated()` and the `index` here? Doesn't seem to be used in the code anyway

README.md
```diff
@@ -12,6 +12,8 @@ SwiftGen is a suite of tools written in Swift to auto-generate Swift code (or an
 * [`enums` for your NSStoryboards and their Scenes](#nsstoryboard)
 * [`enums` for your Colors](#uicolor-and-nscolor).
 * [`enums` for your Fonts](#uifont-and-nsfont).
+* [`classes` for your models](#models)
+* [`any code` for your JSON spec](#json)
```
AliSoftware: Everything else is in plural, so I'd say "JSON specs" here too

GenumKit/Stencil/Filters.swift
plivesey: Fixed

GenumKit/Stencil/MapNode.swift
plivesey: Fixed

GenumKit/Stencil/MapNode.swift
plivesey: Looks like from a previous iteration of the code. Fixed.

README.md
plivesey: Fixed

GenumKit/Stencil/Filters.swift
krzysztofzablocki: it's already in stencil http://stencil.fuller.li/en/latest/builtins.html#join

README.md
```diff
@@ -12,6 +12,8 @@ SwiftGen is a suite of tools written in Swift to auto-generate Swift code (or an
 * [`enums` for your NSStoryboards and their Scenes](#nsstoryboard)
 * [`enums` for your Colors](#uicolor-and-nscolor).
 * [`enums` for your Fonts](#uifont-and-nsfont).
+* [`classes` for your models](#models)
+* [`any code` for your JSON specs](#json)
```
djbe: This whole `any code` section seems extraneous to me. I think there should be one section "Parse your `json` files", and then one subsection in that about parsing models.

README.md
```diff
@@ -614,8 +616,196 @@ let sameFont = UIFont(font: FontFamily.Helvetica.Bold, size: 16.0) // iOS
 let sameFont = NSFont(font: FontFamily.Helvetica.Bold, size: 16.0) // OS X
 ```
 
----
+## Models
+
```
djbe: See comment above. I'd rename this to `JSON`, and a general description about how a user could parse whatever they want using custom templates. Explain that there's a `spec` variable.

Then add a subsection about model-parsing.

README.md
```diff
@@ -614,8 +616,196 @@ let sameFont = UIFont(font: FontFamily.Helvetica.Bold, size: 16.0) // iOS
...
+```sh
+swiftgen json /path/to/model/file -t model
+swiftgen templates cat model-base > /path/to/output/file
+```
```
djbe: I'm not really a fan of having a command that's effectively split into 2 dependent parts. I understand you want to reduce code duplication, and I wholeheartedly agree with that sentiment.

That said, would it be better to move the contents of `model-base` to a gist or mini-library? There are new swift JSON parsing libraries appearing every minute, and everyone has their preferences about which one to use. I'm just wondering if it's worth adding a static template to SwiftGen just for this.

README.md
```diff
@@ -614,8 +616,196 @@ let sameFont = UIFont(font: FontFamily.Helvetica.Bold, size: 16.0) // iOS
...
+
+This takes a JSON spec and converts it into a model file which supports parsing from a dictionary, converting back into a dictionary, equality, mapping, and three different initialization options.
```
djbe: Make bullet points about what's supported.

README.md
```diff
@@ -614,8 +616,196 @@ let sameFont = UIFont(font: FontFamily.Helvetica.Bold, size: 16.0) // iOS
...
+
+The input is a JSON file which creates a spec for the object. The schema is a custom schema but is based off a few standards (AVRO, Pegasus, JSON Schema). The spec takes a type name and a list of fields. Each field specifies it's type, name, whether it's optional and a doc. It can also specify `array` and `element` for an array of elements. The type and element field can be anything. The model generator simply converts it to an uppercase string for the type.
+
+The model template also requires some static code which makes the generated code smaller. This static code is outputed by `swiftgen templates cat model-base > /path/to/output/file`. You should generate this code as well as the model code for each of your JSON specs.
```
djbe: See above comment, this could simply become "Add the line `pod 'MyMiniParsingLibrary'` to your Podfile or library of choice".

README.md
```diff
@@ -614,8 +616,196 @@ let sameFont = UIFont(font: FontFamily.Helvetica.Bold, size: 16.0) // iOS
...
+  }
+}
+
+public func ==(lhs: , rhs: PersonModel) -> Bool {
+  return lhs.firstName == rhs.firstName && lhs.lastName == rhs.lastName && lhs.friends == rhs.friends && lhs.optFriends == rhs.optFriends && lhs.age == rhs.age
+}
+```
+</details>
+
+## JSON spec
```
djbe: See first comments above, I'd rearrange this.

UnitTests/TestSuites/JSONModelTests.swift
```diff
@@ -0,0 +1,23 @@
...
+  func testFullModel() {
+    let parser = JSONFileParser()
+    try? parser.parseFile(path: Fixtures.path(for: "Model.json", subDirectory: "Models"))
+
+    let template = GenumTemplate(templateString: Fixtures.string(for: "json-model.stencil"), environment: genumEnvironment())
+    let result = try! template.render(parser.stencilContext())
+
+    let expected = Fixtures.string(for: "Models-Dir-Basic.swift.out")
+    XCTDiffStrings(result, expected)
+  }
```
djbe: Add more tests:
- empty input model
- incorrect input model

swiftgen-cli/json.swift
```diff
@@ -0,0 +1,32 @@
...
+let jsonCommand = command(
+  outputOption,
+  templateOption(prefix: "json"),
+  templatePathOption,
+  Argument<Path>("FILE", description: "JSON to parse", validator: fileExists)
+) { output, templateName, templatePath, path in
+  let parser = JSONFileParser()
+
+  do {
+    try parser.parseFile(path: path)
```
djbe: It might be interesting to rewrite your parser to accept multiple input files, see #213 for more information. If you do, don't forget to add unit tests for those.

templates/json-model.stencil
```diff
@@ -0,0 +1,120 @@
+// Generated using SwiftGen, by O.Halligon — https://github.com/AliSoftware/SwiftGen
+
+public final class {{ spec.type }}: DataModel, Equatable {
```
djbe: See comment above, would be cool to accept multiple `spec`s.

djbe:
Hi @plivesey, with the latest SwiftGen release we've split the whole project into many sub-parts. If you'd like I can help you get started on reworking this to match the project changes.

The best place to start is with the parser, which'd go in the [SwiftGenKit](https://github.com/SwiftGen/SwiftGenKit) repository. It might be best to open an issue there first though, so we can discuss the scope of this feature, and what would fit in the SwiftGen project and what wouldn't.

hebertialmeida:
@AliSoftware @djbe This was a need that I also had. So as I use RocketData that was created by @plivesey and SwiftGen I arrived here.

I decided to create a project and that satisfy this need, the biggest motivation to create a separated project is to support more languages like Kotlin and Java. I would love to hear what do you think.
https://github.com/hebertialmeida/ModelGen

This is also somehow related to #30 from @Liquidsoul

AliSoftware:
@hebertialmeida Hey, thx for the feedback

> the biggest motivation to create a separated project is to support more languages like Kotlin and Java

I'm not sure I understand. SwiftGen, despite its name, is capable of generating whatever you like, as it's based on templates. In fact, in a side projects I once used it to generate Java code using a custom template 😉 

The next big pressing feature for SwiftGen after the long-awaited 5.0 release tomorrow will be support for Xcode 9 (Swift 4 templates + support for colors in xcassets). Once this lands, we'll probably start working on the support for swiftgen.yml config file, a JSON parser (which will address the current issue) and maybe merging some repos (in light of what splitting in 4 GH repos led to in term of maintenance and speed of introducing new features).

So JSON parsing is definitely on the radar — and the solution will probably be way more simple to implement now, as it's simply gonna be a pass-thru class conforming to Parser protocol which will automatically leverage support for multiple input files, ability to use `--param` to toggle the generation of the common static code or not, etc.

AliSoftware:
BTW I see that you already implemented the "use an yml file to configure the tool" feature in ModelGen (which is something we definitely want for SwiftGen, there's an issue pending about that somewhere), would love some contribution on that ;)

AliSoftware:
Quickly looking at your SourceCode, and seing it uses SwiftGenKit already 👍 , I think we could indeed share some implementations there, like your `JSONParser.swift` file is exactly what I had in mind in the first place.

hebertialmeida:
I can definitely try to help with the yml file, I did that for ModelGen and it was simple. I don't have experience with installers, later I want to add it to Homebrew or something to make it easy.

It is built on top of SwiftGenKit and Stencil with Commander already, I have based on SwiftGen and Sourcery 😄

djbe:
Correct me if I'm wrong, but you don't actually use SwiftGenKit, do you? You depend on it, sure, but I don't see you importing it anywhere. You *do* use StencilSwiftKit though 👍

hebertialmeida:
@djbe OMG, Yeah you are right, sorry about that, I started using it but then I have created my parsers based on it, then I forgot to remove the dependency. 👍

PR #190, opened 2016-10-26
remove key label from tr function in swift3

AndrewSB:


AliSoftware:
Good idea 👍
But don't the other Swift3-compatible templates have to have the same change as well to make it consistent whatever Swift3-compatible template people use?

PS: Don't be surprised but I'll wait a bit to merge that one, as this will generate code with breaking changes, so that means that I'll need to do a _major_ version bump (SwiftGen 4.0.0) so that people using SwiftGen 3 with one of the built-in Swift3 templates won't see their code break. And I've already some other PRs late which will fit in a minor version before deciding to bump to major breaking one… But otherwise I'm 👌 for the idea/change and will merge once all those other pending PRs not requiring major bump are finally merged first!

AndrewSB:
That sounds 💯

I'm using this for now
screen shot 2016-10-26 at 4 44 21 pm
So a `4.0.0` release sounds perfect 😄

AndrewSB:
Oh, I totally missed the first part of your response.
Yes, the other swift3 templates need to be updated too.

I'll add a commit shortly

AndrewSB:
It looks as if the `strings-dot-syntax-swift3.stencil` template doesn't have a free `tr(:)` function. So it doesn't require a change.

So I think this may be the only change required for Strings in swift3

AndrewSB:
^ merge was so cool! It was done through Github, not the CLI 😮

AliSoftware:
Yup, that's one of the nice things GitHub recently added to its interface 👍

https://github.com/blog/2051-protected-branches-and-required-status-checks
> The new “Update branch” button lets you merge the latest changes from the base branch of the Pull Request with one click.

dostrander:
@AndrewSB This looks like it might need to be updated :)

dostrander:
@AndrewSB Can you add a changelog entry as well

AndrewSB:
done @dostrander! 😄

dostrander:
Thanks @AndrewSB merging

PR #192, opened 2016-11-03
Add multiline support for strings file

### LINKED ISSUE #187: Constants not generated for multiline text
This seems to be\* valid way of defining multiline text in `strings` file:

```
"welcome.intro" = "Multiline intro.\
\
Many lines!";
```

`swiftgen strings` omits it and it's not available in generated file.

*It's displayed correctly in the app and Xcode highlighting picks this up.

alldne:
Fixes #187

I would like to add an test case covering this feature but I don't even know how to run test cases 😂
Help needed!


alldne:
I found the command, `rake tests`.. 

The reason of test failure is that the ordering of `Entry` was broken.

GenumKit/Parsers/StringsFileParser.swift
```diff
@@ -17,12 +31,41 @@ public final class StringsFileParser {
...
+    let plist = try NSPropertyListSerialization.propertyListWithData(data, options: NSPropertyListReadOptions.Immutable, format: nil)
+
+    guard let dict = plist as? Dictionary<String, String> else {
+      throw StringsFileParserError.InvalidFormat
+    }
+
     var encoding: NSStringEncoding = NSUTF16StringEncoding
     let fileContent = try NSString(contentsOfFile: path, usedEncoding: &encoding)
+
+    func extractKey(line: String) -> String? {
```
AliSoftware: Not sure why you'd need that.
Now that we parse the `.strings` file as being an `NSDictionary`, we don't deal with _lines_ in the text files, but already have key/value pairs, so you already have the key. No need to get the lines from the textual format anymore nor extract their keys!

GenumKit/Parsers/StringsFileParser.swift
```diff
@@ -17,12 +31,41 @@ public final class StringsFileParser {
...
+      }()
+
+      let range = NSRange(location: 0, length: (line as NSString).length)
+      if let match = regex.firstMatchInString(line, options: [], range: range) {
+        return (line as NSString).substringWithRange(match.rangeAtIndex(1))
+      }
+      return nil
+    }
+
     let lines = fileContent.componentsSeparatedByCharactersInSet(NSCharacterSet.newlineCharacterSet())
```
AliSoftware: This is not needed anymore, we don't need to treat the `.strings` file as a textual file now that we load it as a dictionary

GenumKit/Parsers/StringsFileParser.swift
```diff
@@ -17,12 +31,41 @@ public final class StringsFileParser {
 
   // Localizable.strings files are generally UTF16, not UTF8!
   public func parseStringsFile(path: String) throws {
+    guard let data = NSData(contentsOfFile: path) else {
+      throw StringsFileParserError.FailureOnLoading(path: path)
+    }
+
+    let plist = try NSPropertyListSerialization.propertyListWithData(data, options: NSPropertyListReadOptions.Immutable, format: nil)
```
AliSoftware: If I'm not mistaken, all of those lines could be done via a single `let plist = Dictionary(contentsOfFile: path)`

GenumKit/Parsers/StringsFileParser.swift
```diff
@@ -17,12 +31,41 @@ public final class StringsFileParser {
...
+    }
+
     let lines = fileContent.componentsSeparatedByCharactersInSet(NSCharacterSet.newlineCharacterSet())
 
-    for case let entry? in lines.map(Entry.init) {
-      addEntry(entry)
+    for case let key? in lines.map(extractKey) {
+      if let translation = dict[key] {
+        addEntry(Entry(key: key, translation: translation))
+      }
```
AliSoftware: Now that we have a `Dictionary<String, String>` of the keys and their translation directly (and don't use the textual representation of lines anymore) we can simply do something like:

```swift
for (key, translation) in dict {
  addEntry(Entry(key: key, translation: translation))
}
```

GenumKit/Parsers/StringsFileParser.swift
```diff
@@ -82,6 +125,11 @@ public final class StringsFileParser {
       self.init(key: key, translation: translation, types: types)
     }
 
+    public init(key: String, translation: String) {
+      let types = PlaceholderType.fromFormatString(translation)
+      self.init(key: key, translation: translation, types: types)
+    }
+
     private static let lineRegEx: NSRegularExpression = {
```
AliSoftware: This won't probably be needed anymore now that we process the file as a dictionary PLIST instead of a textual file.

GenumKit/Parsers/StringsFileParser.swift
alldne: The reason why I didn't use `Dictionary(contentsOfFile: )` is that when it fails, it just returns `nil`. `Dictionary(contentsOfFile: )` can fail with various reasons: a file does not exist, an invalid encoding, malformed plist file, etc.. But we are not able to know the reason of the failure with `Dictionary(contentsOfFile:)`.

`NSPropertyListSerialization.propertyListWithData(..)` throws an error by itself. By using it, we can propagate errors with rich information that should be helpful for users.

Anyway, I also agree that `let plist = Dictionary(contentsOfFile: path)` is much more simple and easy to read. It seems that both have pros and cons so I will follow your decision.

GenumKit/Parsers/StringsFileParser.swift
alldne: Right. The first look of this PR didn't have that `extractKey` and _line-by-line_ something. But after running test cases, I encountered that many cases had failed. That's because `NSDictionary` does not guarantee the order of keys. By order I mean the order that the key was added by.

I couldn't find a good way to preserve the original order of the keys. (We can't sort these keys in alphabetical order, right?😅)
So I decided reading the file line-by-line and extracting the keys, which seems quite redundant and ugly.

This code just works. Do you have any idea to make this good?

AliSoftware: Ah, good point.

Not sure what's best. I would consider sorting keys alphabetically upon generation (= don't sort the dictionary keys, but sort the array of `Entry` once it's build or when building the StencilContext from it), that doesn't seem too bad.

Sure that would mean that the original order in the b`.strings` file isn't kept but that would avoid parsing the file twice, once as a PLIST and once as text, so… 🤔

AliSoftware: The more I think about it the less I'm sure people care about the order, as:

* they probably never go read the generated file, but rather rely on auto completion
^ most people often use reverse-DNS naming for their keys, like `"home.header.title"` so sorting alphabetically would still keep related keys together

At my company we actually _do_ sort our `Localizable.strings` file alphabetically (they are auto-generated by a tool of ours exporting localization from the POEditor service), so…

GenumKit/Parsers/StringsFileParser.swift
AliSoftware: Ah indeed. It's very unlikely (if even possible) that a valid Localizable.strings generates an invalid PLIST. And an invalid Localizable.strings generally makes Xcode generate a warning at compile time iirc. But I appreciate the idea of having a more explicit error so maybe you're right, that's a good idea to keep it that way

AliSoftware:
Btw didn't answer your original comment, but to run tests sure you can use `rake tests` but you can also run them from Xcode like you'd do for any other project, using Product > Tests (Cmd-U)

dostrander:
@alldne This looks like it might need to be updated :)

dostrander:
@alldne Just found the magic "Update Branch" button :)

AliSoftware:
About the still-pending requested changes, what do you guys think about the debate of having to parse the file twice (once as Plist to parse multi-lines correctly, once as text to get the original keys order) vs only paring the file once and sort the keys alphabetically in the generated file?

I personally opt for the latter, I think it's not a big deal to order the generated keys as people generally don't look at the generated code anyway but use auto-completion, and that most people already order their keys with reverse-dns notation… What's your thoughts?


About the build failed, there seem to be one test case failing with generated code being different than expected, this needs to be fixed too before we can merge 😉

dostrander:
I tend to agree that most people don't look at the generated code and just use autocomplete, however there are probably some people who do and want it grouped like the file. If we can benchmark it with a large number of entries to determine if it takes too long and if not I'd say leave it in and keep a benchmark test to mark sure it doesn't exceed a certain time.

alldne:
I also prefer sorting the keys and I can even say that if it was my personal project, I would decide in that way without any hesitation. 

But this project is public and many other people might be using it. Some people may not care the order of keys as we expected, but the other people may not as @dostrander mentioned. 
And the point is that, it's impossible we predict what users think important.

So..🤔 In my opinion, if you consider the users, keeping the original behavior is quite matter.
By keeping the original behavior, we might lose a little bit of performance and get some ugly(?) code.

In addition, personally I think, especially for open source project, people writing the code are as important as people using the program.

alldne:
Well, now entries are sorted by its key and quite a lot of files under `/UnitTests/expected` are updated along.

Note that when you see the generated file, the keys might not be in alphabetical order. (_what?!_)
This sounds weird but.. the reason is that after sorting the keys, template engine change keys into camel case.

For example, assume that we have some keys: `ObjectOwnership`, `alert_title`, `AAA` and `bbb`.
If we sort these keys, we get the following result:
```swift
["AAA", "ObjectOwnership", "alert_title", "bbb"]
```
And then, template engine converts snake case into camel case:
```swift
["AAA", "ObjectOwnership", "AlertTitle", "bbb"]
```

I think it's not a big deal.. but FYI.. 😉

djbe:
Hmm, some unit tests still depend on that code, namely: testLinesWithDefaults()

AliSoftware:
Right… but if we don't do line-parsing but only plist-parsing now, shouldn't we get rid of those tests too then, as they don't make sense anymore?

djbe:
True, was just looking through the documentation of NSPropertyListSerialization on how that works just in case 😆

PR #194, opened 2016-11-04
Templates for Swift 3 Compliant with API Design Guidelines

afonsograca:
In order to comply to [SE-0023](https://github.com/apple/swift-evolution/blob/master/proposals/0023-api-guidelines.md), some changes had to be made to the template stencils. Specifically:

- Names of types and protocols are UpperCamelCase. Everything else is lowerCamelCase.
- Prune redundant type names - Matched text extends to the end of the type name


dostrander:
@afonsograca this looks good, just need an update with master :)

afonsograca:
@dostrander done!

This also solves a namespace issue that I came across today. Given the following strings:
```
"user.confirm" = "Confirm";
"user.confirm.title" = "Do you want to confirm?";
"user.confirm.description" = "Once you do you cannot go back.";
```

This creates a static string called `Confirm` and a struct `Confirm` making it ambiguous for the compiler.

dostrander:
@afonsograca NICE that was one of the problems I had when I had attempted the `dot-syntax` for this. GOOD CATCH! Just waiting on building

AliSoftware:
@dostrander don't forget to add CHANGELOG entries as you merge PRs 😉 so we keep track of changes as we merge them and avoid forgetting them later at release time.
 (I usually even ask the OP to credit themselves in the CHANGELOG)

PR #195, opened 2016-11-07
Core data rebase

Igor-Palaguta:
Merged with recent master, please review


GenumKit/Stencil/Contexts.swift
```diff
@@ -295,3 +295,94 @@ extension FontsFileParser {
     return Context(dictionary: ["enumName": enumName, "families" : families], namespace: GenumNamespace())
   }
 }
+
+/* MARK: - Stencil Context for CoreData
```
kylef: Btw, with #216 you should be able to greatly simplify this code since Stencil 0.7.0 supports Swift type introspection and you can pull variables directly from Swift types in Stencil.

Example:

```html+django
There are {{ articles.count }} articles.

<ul>
  {% for article in articles %}
    <li>{{ article.title }} by {{ article.author }}</li>
  {% endfor %}
</ul>
```

```swift
struct Article {
  let title: String
  let author: String
}

let context = Context(dictionary: [
  "articles": [
    Article(title: "Migrating from OCUnit to XCTest", author: "Kyle Fuller"),
    Article(title: "Memory Management with ARC", author: "Kyle Fuller"),
  ]
])
```

Igor-Palaguta: @kylef Stencil 0.7 is huge release! Thank you so much for it!!!

PR #200, opened 2016-11-16
Add changelog entry for escaping reserved keywords

djbe:
Apparently we can't directly modify the changelog


AliSoftware:
Yeah, that's normal, I've locked the `master` branch to force all contributors to only do changes via Pull Requests and thus ensure that other people can vet the changes and that the CI always pass before merging… even if for trivial changes like that this might feel cumbersome ^^

PR #201, opened 2016-11-16
Update project to Swift 3

ahtierney:
- [x] run migrator to swift 3
- [x] update pods to latest, swift 3 compatible versions
- [x] audit APIs for swift 3 compliance

Here's a migration to swift 3 as per issue #169 with a goal to bring all the project dependencies up to date.


AliSoftware:
Cool, thanks for this work on the migration!

I think you should add an item to your checklist: "refactor the API to be compliant with Swift 3 naming guidelines". It's cool if it compiles, but this migration is the best occasion to also make the internal method names Swift3-compliant too :wink:

As per switching to Swift3 being the default template, that would mean a major version bump of SwiftGen IMHO, so that people using the current version will still have their generated code compatible even if they kept the default template.

If we manage to merge this PR before the imminent SwiftGen 4.0.0, it might be the right time. But it might be better to do separate PRs (this one to convert these internal code of SwiftGen too Swift 3 without any visible external changes for the users, and a separate one to rename there templates to make the Swift 3 templates the default and update the documentation accordingly) this way we can merge this one quickly then decide about the other, public migration of the templates if we have time to integrate in 4.0 or will do in a 5.0. I'm more in favor of the last solution to allow more flexibility at least it will allow us to decide art the time we decide to release there next version!

ahtierney:
Sounds like a plan! I'll do an API audit and aim to keep this branch internal / non-breaking. Thanks!

ahtierney:
I've done the API audit and merged up with master. Naming is hard, so let me know if there's anything you take issue with!

GenumKit/Parsers/StoryboardParser.swift
```diff
@@ -86,7 +86,7 @@ public final class StoryboardParser {
     }
   }
 
-  public func addStoryboardAtPath(_ path: String) {
+  public func addStoryboard(atPath path: String) {
```
djbe: To be consistent with your naming in fonts & color parsers, I'd change this to addStoryboard(at path: )

GenumKit/Parsers/StringsFileParser.swift
```diff
@@ -59,8 +59,8 @@ public final class StringsFileParser {
       }
     }
 
-    public static func fromFormatString(_ format: String) -> [PlaceholderType] {
-      return StringsFileParser.typesFromFormatString(format)
+    public static func placeholders(fromFormatString str: String) -> [PlaceholderType] {
```
djbe: Leave the parameter type out: placeholders(fromFormat str: String)

GenumKit/Parsers/StringsFileParser.swift
```diff
@@ -130,7 +130,7 @@ public final class StringsFileParser {
     }()
 
   // "I give %d apples to %@" --> [.Int, .String]
-  fileprivate static func typesFromFormatString(_ formatString: String) -> [PlaceholderType] {
+  fileprivate static func placeholders(fromFormatString formatString: String) -> [PlaceholderType] {
```
djbe: Same as above: placeholders(fromFormat formatString: String)

GenumKit/Parsers/StringsFileParser.swift
ahtierney: 👍  agree

GenumKit/Parsers/AssetsCatalogParser.swift
```diff
@@ -41,7 +41,7 @@ private enum AssetCatalog: String {
 extension AssetsCatalogParser {
   static let imageSetExtension = "imageset"
 
-  fileprivate func process(_ items: [[String: AnyObject]], prefix: String = "") {
+  fileprivate func processCatalogItems(_ items: [[String: AnyObject]], prefixedWith prefix: String = "") {
```
ahtierney: 👍 agree

ahtierney:
It looks like CI is having an issue with it's pods repo. Based on what I'm seeing it seems like I need to add a `pod repo update` to the `.travis.yml` file's install step. Any thoughts regarding this? Apart from CI I think I've got everything in now.

djbe:
Problem with that is that it takes a really long time. See here:
https://github.com/travis-ci/travis-ci/issues/6473

AliSoftware:
Indeed.
`pod repo update` isn't an *installation* step _per se_, so I'd instead add it as a step before the linting, or even in the same command: `pod repo update && pod lib lint …`

djbe:
In the thread I linked, they mention using a new xcode image, which version do we use?

AliSoftware:
Another solution to avoid it taking too long is to use the `--quick` flag on `pod lib lint`. Would probably be better.

ahtierney:
yea if we can avoid updating that pods that would be best, does --quick skip repo validation?

djbe:
try changing the osx image to:
osx_image: xcode8.1
and see if it fixes the issue

AliSoftware:
`--quick` avoids any linting that would need to download anything, so yeah.

I think maybe we should do both: upgrade to the Travis image `xcode8.1` and use `--quick`, right?

ahtierney:
Sounds good to me

ahtierney:
success 🎉

AliSoftware:
Good job! 👍

Haven't much time to audit the new API, maybe if someone else can audit before merging (as it's the kind of migration where it's easy to miss something)?

ahtierney:
Definitely in the more eyes the better camp. I'll see if I can recruit @dostrander to it.

djbe:
I'll see if I can get this to work, could be useful:
https://github.com/dduan/needless

djbe:
All right, either `needless`  doesn't find much, or we've got most of it covered. The only thing it found was in SwiftIdentifier.swift:
```swift
func swiftIdentifier(fromString string: String
```
should become
```swift
func swiftIdentifier(from string: String
```

I tested GenumKit, swiftgen-cli, the unit tests and the expected output.

ahtierney:
I'm down for that. I'll push the change in a minute.

GenumKit/Stencil/GenumTemplate.swift
```diff
@@ -7,42 +7,41 @@
 import Stencil
 
 // Workaround until Stencil fixes https://github.com/kylef/Stencil/issues/22
-public class GenumTemplate: Template {
+open class GenumTemplate: Template {
   public override init(templateString: String) {
     let templateStringWithMarkedNewlines = templateString
-      .stringByReplacingOccurrencesOfString("\n\n", withString: "\n\u{000b}\n")
-      .stringByReplacingOccurrencesOfString("\n\n", withString: "\n\u{000b}\n")
+        .replacingOccurrences(of: "\n\n", with: "\n\u{000b}\n")
```
AliSoftware: Not sure why I did the replacement twice, but I think that was intentional (like for cases like `\n\n\n`, the first replacement only replaces the first 2 `\n` and the second takes care of the remaining ones)… so we should keep that dual replacement in place even after the Swift 3 migration.

SwiftGen.playground/Pages/Fonts-Demo.xcplaygroundpage/Contents.swift
```diff
@@ -40,7 +40,7 @@ struct FontFamily {
 // Using the UIFont constructor…
 let helvetica = UIFont(font: FontFamily.Helvetica.Regular, size: 20.0)
 // Or using the enum value and its `font` method
-let helveticaNeue = FontFamily.HelveticaNeue.Regular.font(20.0)
+let helveticaNeue = FontFamily.HelveticaNeue.Regular.font(size: 20.0)
 
-let helveticaBoldBig = FontFamily.Helvetica.Bold.font(100.0)
+let helveticaBoldBig = FontFamily.Helvetica.Bold.font(size: 100.0)
```
AliSoftware: Don't we generate lowercase cases now, like `Helvetica.bold` and not `Helvetica.Bold`? I mean, if the Playground is in Swift 3 (which this `font(size: 100)` instead of `font(100)` seems to suggest, should we assume that the code in the Playground might reflect the code from the `-t swift3` template, or should we still use the code generated by the default template in the Swift 3 playground?

SwiftGen.xcodeproj/project.pbxproj
```diff
@@ -548,12 +548,14 @@
 			isa = PBXNativeTarget;
 			buildConfigurationList = 09A87B451BCC9B8000D9B9F5 /* Build configuration list for PBXNativeTarget "swiftgen" */;
 			buildPhases = (
+				54570B6ACBB962B5ED01A405 /* [CP] Check Pods Manifest.lock */,
 				BEC9F3A158B764C68AF03EF9 /* [CP] Check Pods Manifest.lock */,
```
AliSoftware: Seems strange that the Build Phase got duplicated here. We gotta double-check that.

SwiftGen.xcodeproj/project.pbxproj
```diff
@@ -576,6 +579,8 @@
 				09A87B4E1BCCA2C600D9B9F5 /* Resources */,
 				C40C2D80E9C1FEEAA4FD6B14 /* [CP] Embed Pods Frameworks */,
 				6C7EBDA1392A1BFCC9F0A2DA /* [CP] Copy Pods Resources */,
+				1CAEE405320C3DBD251A68B3 /* 📦 Embed Pods Frameworks */,
+				B8FA4CC6E874346645F002E3 /* 📦 Copy Pods Resources */,
```
AliSoftware: Same, seems like you used an outdated CocoaPods version to regenerate that project, the version of CP when I (it happens that it was me who did that on CP 😄) introduced the emoji in the build phases names, and which got reverted to `[CP]` in a later version… Gotta clean that up and regenerate the project with a recent version of CP probably…

SwiftGen.xcodeproj/project.pbxproj
```diff
@@ -844,6 +909,21 @@
 			shellScript = "\"${SRCROOT}/Pods/Target Support Files/Pods-swiftgen/Pods-swiftgen-resources.sh\"\n";
 			showEnvVarsInLog = 0;
 		};
+		F815D65995A8B21A935964D2 /* 📦 Copy Pods Resources */ = {
```
AliSoftware: Yup, it's clear that this build phase is the duplicate of the one just before…

SwiftGen.xcodeproj/project.pbxproj
dostrander: Do you think it would it be beneficial to put a Gemfile in the project to make sure people are on the same version of cocoapods?

AliSoftware: Probably a good idea indeed 👍

AliSoftware: @djbe mine are also installed system-wide, that doesn't impact that :wink:

When you run `bundle install` bundler will install all the required gems, in the right versions, system-wide (unless told otherwise). Then you just need to run `bundle exec pod install` (instead of just `pod install`) to be sure to run the right version of your system-wide-installed CocoaPods, even if that one (the one defined in the Gemfile) isn't there last version installed in your system.

GenumKit/Parsers/AssetsCatalogParser.swift
```diff
@@ -12,7 +12,8 @@ public final class AssetsCatalogParser {
 
   public init() {}
 
-  public func addImageName(name: String) -> Bool {
+  @discardableResult
+  public func addImageName(_ name: String) -> Bool {
```
ahtierney: I went back and fourth on this one, I think because of the collection is actually `imageNames` I have to confess I'm still a little unclear on if there is an objective differentiator between `addImage()` `add(image:)` 😬 I agree though, `addImage(named name: String)` is best for api continuity.

GenumKit/Parsers/ColorsFileParser.swift
```diff
@@ -52,53 +52,53 @@ private func parseHexString(hexString: String) throws -> UInt32 {
 // MARK: - Text File Parser
 
 public final class ColorsTextFileParser: ColorsFileParser {
-  public private(set) var colors = [String:UInt32]()
+  public fileprivate(set) var colors = [String:UInt32]()
```
ahtierney: agree, I'll do a full audit. A lot of these I think are code style between `extension` grouping.

GenumKit/Stencil/Contexts.swift
```diff
@@ -5,12 +5,23 @@
 //
 
 import Stencil
+fileprivate func < <T: Comparable>(lhs: T?, rhs: T?) -> Bool {
+  switch (lhs, rhs) {
+  case let (l?, r?):
+    return l < r
+  case (nil, _?):
+    return true
```
ahtierney: Agree. I'm kind of impressed the converter produces this 🤔

ahtierney:
Thanks for the feedback!

ahtierney:
Feedback is up:
- Rebase off master
- Naming changes
- converted all `fileprivate` to `private` where possible (left one case in the asset parser where extensions *could* be restructured, but didn't want to be too opinionated about it)
- re-gen'd playgrounds
- cleaned up pods but running `deintegrate` and rebuilding

GenumKit/Stencil/GenumTemplate.swift
djbe: If so, then even a double replacement won't cover all cases. It might be better to use a regex for this:
```swift
templateString.replacingOccurrences(of: "\n\n+", with: "\n\u{000b}\n", options: .regularExpression)
```

ahtierney:
@djbe I think I agree in principle, but if I understand the code correctly the intent is not to flatten *inbound* new lines, but to preserve them.
So `"\n\n\n\n\n\n foobaz \n\n\n\n\n\n\n"` should output the same, not `\n\n foobaz \n\n`
Correct me if I'm wrong here. Either way might make sense to have some tests for this so expected behavior is clear and it always behaves the same.

djbe:
Huh, yeah. I honestly just took a look at the issue (https://github.com/kylef/Stencil/issues/22) and there they just mention extra empty lines they want to get rid of.

djbe:
@AliSoftware could we get this merged? I think al your feedback was covered.

GenumKit/Parsers/AssetsCatalogParser.swift
```diff
@@ -13,7 +13,7 @@ public final class AssetsCatalogParser {
   public init() {}
 
   @discardableResult
-  public func addImageName(_ name: String) -> Bool {
+    public func addImage(named name: String) -> Bool {
```
AliSoftware: Huh, why the indentation change here? Seems like a misindentation was introduced by mistake

GenumKit/Parsers/ColorsFileParser.swift
```diff
@@ -52,12 +52,12 @@ fileprivate func parseHex(_ hexString: String) throws -> UInt32 {
 // MARK: - Text File Parser
 
 public final class ColorsTextFileParser: ColorsFileParser {
-  public fileprivate(set) var colors = [String:UInt32]()
+  public fileprivate(set) var colors = [String : UInt32]()
```
AliSoftware: SwiftLint rules usually prefer no space before `:` (but one space after)

GenumKit/Stencil/Contexts.swift
```diff
@@ -195,7 +195,7 @@ extension StringsFileParser {
     }
 
     let strings = entries
-        .sorted() { $0.key < $1.key }
+        .sorted() { $0.key.lowercased() < $1.key.lowercased() }
```
AliSoftware: Might be better to use `str1.caseInsensitiveCompare(str2)` here. Or anything that would ensure that any string case but also any diacritics and other niceties of Unicode are insensitively-compared properly (even if those are very rare to uncommon in _keys_)

CHANGELOG.md
```diff
@@ -28,7 +28,8 @@
   ensuring that the parser correctly handles namespaced folders.  
   [David Jennes](https://github.com/djbe)
   [#199](https://github.com/AliSoftware/SwiftGen/pull/199)
-* Swift 3 migration [ahtierney](https://github.com/ahtierney)
+* Swift 3 migration
```
AliSoftware: Missing the full-stop + 2-spaces at the end of this line so that the rendered markdown is formatted properly

GenumKit/Parsers/ColorsFileParser.swift
```diff
@@ -8,7 +8,7 @@ import Foundation
 import AppKit.NSColor
 
 public protocol ColorsFileParser {
-  var colors: [String: UInt32] { get }
+  var colors: [String : UInt32] { get }
```
AliSoftware: SwiftLint usually prefers no space before `:` (but one after)

CHANGELOG.md
```diff
@@ -28,6 +28,9 @@
   ensuring that the parser correctly handles namespaced folders.  
   [David Jennes](https://github.com/djbe)
   [#199](https://github.com/AliSoftware/SwiftGen/pull/199)
+* Swift 3 migration
```
AliSoftware: Missing the full-stop + 2-spaces at the end of this line so that the rendered markdown is formatted properly.

GenumKit/Stencil/Contexts.swift
```diff
@@ -165,29 +166,43 @@ extension StoryboardParser {
...
+                "translation": entry.translation.newlineEscaped,
+                "keytail": keytail
+        ]
       }
     }
 
-    let strings = entries.sort { $0.key < $1.key }.map { entryToStringMapper($0, []) }
-    let structuredStrings = structure(entries, mapper: entryToStringMapper, currentLevel: 0, maxLevel: 5)
+    let strings = entries
+        .sorted { $0.key.lowercased() < $1.key.lowercased() }
```
AliSoftware: Might be better to use str1.caseInsensitiveCompare(str2) here. Or anything that would ensure that any string case but also any diacritics and other niceties of Unicode are insensitively-compared properly (even if those are very rare to uncommon in keys)

GenumKit/Parsers/AssetsCatalogParser.swift
```diff
@@ -67,12 +68,13 @@ extension AssetsCatalogParser {
 // MARK: - ACTool
 
 extension AssetsCatalogParser {
-  private func loadAssetCatalogContents(path: String) -> [[String: AnyObject]]? {
+  fileprivate func loadAssetCatalog(at path: String) -> [[String : AnyObject]]? {
```
AliSoftware: SwiftLint rules usually prefer no space before : (but one space after)

GenumKit/Parsers/AssetsCatalogParser.swift
```diff
@@ -12,7 +12,8 @@ public final class AssetsCatalogParser {
 
   public init() {}
 
-  public func addImageName(name: String) -> Bool {
+  @discardableResult
+    public func addImage(named name: String) -> Bool {
```
AliSoftware: Huh, why the indentation change here? Seems like a misindentation was introduced by mistake

GenumKit/Parsers/ColorsFileParser.swift
```diff
@@ -52,42 +52,42 @@ private func parseHexString(hexString: String) throws -> UInt32 {
 // MARK: - Text File Parser
 
 public final class ColorsTextFileParser: ColorsFileParser {
-  public private(set) var colors = [String:UInt32]()
+  public private(set) var colors = [String : UInt32]()
```
AliSoftware: Ditto (note that some of those might be from me and my old code-style, not all from changes in this PR? But would be nice to uniform this to match SwiftLint's rules anyway)

djbe:
wooo, that works!
I've made the changes + searched globally for ` : `

AliSoftware:
Hell yeah! Let's merge! :tada:

djbe:
Wooo! 😄

ahtierney:
Awesome! Thanks a lot guys. I'll rebase and tune up #204 this weekend!

AliSoftware:
@ahtierney Cool!
Just invited you as an official Collaborator on the project, in case you want to create branches directly on the main repo (instead of in your fork) and some other niceties :wink:

ahtierney:
@AliSoftware Awesome, thank you!

AliSoftware:
@dostrander @djbe I've added a `Gemfile` in 152d48ee1c2f37f2e2f5738d14a38929f47b430f to be able pin the version of CocoaPods used by SwiftGen (while still allowing you to have whatever version you prefer system-wide)

For those of you not familiar with the usage of a `Gemfile` to pin gems like CocoaPods to a specific version, you can read more on the topic [here](https://guides.cocoapods.org/using/a-gemfile.html), but basically, to use it:

* If you haven't the `bundler` gem installed, run `gem install bundler`. This has to be done only once, for the very first time.
* Run `bundle install` to install (system-wide) all the gems (with their corresponding version) listed in the `Gemfile`in case you don't have them already.
  * For example as for now, because the `Gemfile` contains `gem 'cocoapods', '1.2.0.beta.1'` this will install exactly this version of CocoaPods if you don't have it already
* When you want to execute a CocoaPods command on the SwiftGen project, like `pod install` for example, use `bundle exec pod install` instead, to tell it to use the `pod` binary specified in the Gemfile, and not any other default one used by the system
  * For example on my Mac I currently have both CocoaPods `1.1.0` and `1.2.0.beta.1`. If (anywhere, system-wide) I do `pod --version` this will default to `1.1.0` (because stable versions have priority in RubyGems). But if I'm in the SwiftGen's directory and I do `bundle exec pod --version` this will use `1.2.0.beta.1` instead, even if `1.1.0` is the default system-wide and even if I have a `1.3.0` installed on my system in the future.

This ensures consistency in the version every contributor uses when working as a team on SwiftGen together.

PR #203, opened 2016-11-17
Escape String constant names

jkoepcke:
Currently, using reserved keywords like "any" as localization keys will result in a compiler error. This will avoid it.

### DIFF
--- a/templates/strings-dot-syntax-swift3.stencil
+++ b/templates/strings-dot-syntax-swift3.stencil
@@ -20,7 +20,7 @@ struct {{enumName}} {
     return {{enumName}}.tr("{{string.key}}", {{string.params.names|join}})
   }
   {% else %}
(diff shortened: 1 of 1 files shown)


djbe:
Wasn't this already covered in #198?

jkoepcke:
Hm I see, the solution there does not list "any" in the keywords. But the solution there also seems kind of over-engineered if you can just simply escape all constants as in my change...

djbe:
`Any` is covered, check `GenumKit/Stencil/Filters.swift` on line 20.

The reason it was implemented as it is now was probably related to code completion and/or just nice output code?

jkoepcke:
Yes it is listed but apparently it doesn't work currently. Using "any" as localization key will result in "static let Any" which results in compiler error.

AliSoftware:
@jkoepcke Are you sure you tested the `master` version of SwiftGen?
(#198 isn't yet released in homebrew, see [CHANGELOG](https://github.com/AliSoftware/SwiftGen/blob/master/CHANGELOG.md))

jkoepcke:
You are right, I am using the latest pod release. We can close this then.

PR #204, opened 2016-11-19
Issue #169 — Reorganize project as an app bundle

ahtierney:
This pull request covers the remainder of issue #169 involving the conversion to Swift 3 and project restructuring. The last 3 commits are regarding the app bundle / project restructure.

NOTE: This is built on top of PR #201 and should be merged after or instead of that request.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -51,7 +51,7 @@ end
 def defaults(args)
   bindir = args.bindir.nil? || args.bindir.empty? ? Pathname.new('./build/swiftgen/bin') : Pathname.new(args.bindir)
   fmkdir = args.fmkdir.nil? || args.fmkdir.empty? ? bindir + '../lib' : Pathname.new(args.fmkdir)
-  tpldir = args.tpldir.nil? || args.tpldir.empty? ? bindir + '../Resources/templates' : Pathname.new(args.tpldir)
+  tpldir = args.tpldir.nil? || args.tpldir.empty? ? bindir + '../templates' : Pathname.new(args.tpldir)
   [bindir, fmkdir, tpldir].map(&:expand_path)
 end
 
@@ -91,7 +91,7 @@ end
 ## [ Install Tasks ] ##########################################################
 
 desc "Install the binary in $bindir, frameworks — without the Swift dylibs — in $fmkdir, and templates in $tpldir\n" \
-     "(defaults $bindir=./build/swiftgen/bin/, $fmkdir=$bindir/../lib, $tpldir=$bindir/../Resources/templates"
+     "(defaults $bindir=./build/swiftgen/bin/, $fmkdir=$bindir/../lib, $tpldir=$bindir/../templates"
 task 'install:light', [:bindir, :fmkdir, :tpldir] => :build do |_, args|
   (bindir, fmkdir, tpldir) = defaults(args)
 
@@ -112,7 +112,7 @@ task 'install:light', [:bindir, :fmkdir, :tpldir] => :build do |_, args|
 end
 
 desc "Install the binary in $bindir, frameworks — including Swift dylibs — in $fmkdir, and templates in $tpldir\n" \
-     "(defaults $bindir=./swiftgen/bin/, $fmkdir=$bindir/../lib, $tpldir=$bindir/../Resources/templates"
+     "(defaults $bindir=./swiftgen/bin/, $fmkdir=$bindir/../lib, $tpldir=$bindir/../templates"
 task :install, [:bindir, :fmkdir, :tpldir] => 'install:light' do |_, args|
   (bindir, fmkdir, tpldir) = defaults(args)
 
--- a/Rakefile
+++ b/Rakefile
@@ -51,7 +51,7 @@ end
 def defaults(args)
   bindir = args.bindir.nil? || args.bindir.empty? ? Pathname.new('./build/swiftgen/bin') : Pathname.new(args.bindir)
   fmkdir = args.fmkdir.nil? || args.fmkdir.empty? ? bindir + '../lib' : Pathname.new(args.fmkdir)
-  tpldir = args.tpldir.nil? || args.tpldir.empty? ? bindir + '../templates' : Pathname.new(args.tpldir)
+  tpldir = args.tpldir.nil? || args.tpldir.empty? ? bindir + '../Resources/templates' : Pathname.new(args.tpldir)
   [bindir, fmkdir, tpldir].map(&:expand_path)
 end
 
@@ -91,7 +91,7 @@ end
 ## [ Install Tasks ] ##########################################################
 
 desc "Install the binary in $bindir, frameworks — without the Swift dylibs — in $fmkdir, and templates in $tpldir\n" \
-     "(defaults $bindir=./build/swiftgen/bin/, $fmkdir=$bindir/../lib, $tpldir=$bindir/../templates"
+     "(defaults $bindir=./build/swiftgen/bin/, $fmkdir=$bindir/../lib, $tpldir=$bindir/../Resources/templates"
 task 'install:light', [:bindir, :fmkdir, :tpldir] => :build do |_, args|
   (bindir, fmkdir, tpldir) = defaults(args)
 
@@ -112,7 +112,7 @@ task 'install:light', [:bindir, :fmkdir, :tpldir] => :build do |_, args|
 end
 
 desc "Install the binary in $bindir, frameworks — including Swift dylibs — in $fmkdir, and templates in $tpldir\n" \
-     "(defaults $bindir=./swiftgen/bin/, $fmkdir=$bindir/../lib, $tpldir=$bindir/../templates"
+     "(defaults $bindir=./swiftgen/bin/, $fmkdir=$bindir/../lib, $tpldir=$bindir/../Resources/templates"
 task :install, [:bindir, :fmkdir, :tpldir] => 'install:light' do |_, args|
   (bindir, fmkdir, tpldir) = defaults(args)
 
@@ -223,7 +223,7 @@ namespace :release do
     req = Net::HTTP::Post.new(uri, initheader = {'Content-Type' => content_type})
     yield req if block_given?
     req.basic_auth 'AliSoftware', File.read('.apitoken').chomp
-      
+
     response = Net::HTTP.start(uri.host, uri.port, :use_ssl => (uri.scheme == 'https')) do |http|
       http.request(req)
     end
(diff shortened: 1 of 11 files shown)


djbe:
Good! Only thing I can note is your changelog entries' format:
```
* Description of the change.••
  [Author Name](GH profile link)
  [#PRNum](link to PR)
```
With `•` to be replaced by a space (ensures that the author and PR links are on a new line).

AliSoftware (changes requested):
I haven't reviewed this in detail, but I'm gonna mark it as rejected: don't take it the wrong way, it's probably good, but it's gonna be super-hard to review the diff given how heavy it is from containing both the changes from #201 and the changes just to make it an app bundle, so it's just to keep in mind that we'll have to wait for 201 to be ready an merged, then rebase that on top of it so that this PR only contains the relevant changes to make it an app bundle, which would make it way easier to review and cleaner to merge :wink:

AliSoftware:
Time to rework this PR now that #201 has been modified then merged! :tada:

Could you please rebase this on top of `master` so that it now only include th commits related to the "app bundle migration" change?

Thanks!

ahtierney:
Okay! so this should be ready for a look.
- Rebased off master
- `.app` is runnable and buildable in xcode
    - you can pass args by editing the scheme (we may want to leave some in there as they can be toggled)
    - note I had to disable nsdocumentrevisionsdebugmode (was parsed as command line args)
- Rake now builds a more .app like `Resources` structure

Let me know what you think.

Rakefile
```diff
@@ -51,7 +51,7 @@ end
 def defaults(args)
   bindir = args.bindir.nil? || args.bindir.empty? ? Pathname.new('./build/swiftgen/bin') : Pathname.new(args.bindir)
   fmkdir = args.fmkdir.nil? || args.fmkdir.empty? ? bindir + '../lib' : Pathname.new(args.fmkdir)
-  tpldir = args.tpldir.nil? || args.tpldir.empty? ? bindir + '../templates' : Pathname.new(args.tpldir)
+  tpldir = args.tpldir.nil? || args.tpldir.empty? ? bindir + '../Resources/templates' : Pathname.new(args.tpldir)
```
AliSoftware: I'm not sure I get why you changed this?
The point of the Rakefile is to distribute SwiftGen as a command line tool in the format expected by homebrew and structured like other executables you typically have in /use/local/bin or similar non-app-bundled tools, which _don't_ have a Resources folder

ahtierney: @AliSoftware Great question, my thinking was it'd be best to keep as much parity as makes sense between the debug and the packaged version. Based on the rake install step I assumed that adding the default value would be the lowest touch way to keep things harmonious (if the debug version requires default path `Resources/templates` running install will overwrite the `main.swift` file and require reset). However I certainly understand that the `.app` is a secondary feature. Some other options I considered were:
- A build script to package the .app more like the cli tool (discussed [here](https://github.com/AliSoftware/SwiftGen/issues/169))
- Using a different relative path value depending on the target.
- Resetting the path value in `main.swift` after the install step finishes in the rake file.

Let me know if any of these options appeal to you and I'd be happy to implement it (or if I'm off in the wrong direction all together 😬)

AliSoftware: Ah, I see.

But I think even without this PR, the `main.swift` was modified by the `Rakefile` anyway, as it can always be installed in any place you like depending on the rake task arguments…

And I'd still prefer keeping the structure of the final binary untouched, first because it's more standard, second because changing it would mean adapting the Homebrew's formula too, and because the final product is more important and the .app is just a convenience structure for development, not for distribution.

So I think the best option is to:

* Keep the templates in Resources for the development mode (.app bundle) — We could even point the `templatesRelativePath` to `NSBundle.main.resourcesDirectory` to make it more dynamic? — because that's what .app bundles expect and the place where Xcode will put them
* Keep the `Rakefile` changing that value to `$tpldir`, which by default if not specified when calling `rake`, would be `../templates` to match what Homebrew and CLI binaries expect
* _Don't_ `git reset main.swift` at the end of `rake`. The `Rakefile` already did change the `main.swift` without reverting it before that PR, and I think that's OK:
  * now that the project will be organized as an app bundle, we should only have to use the `Rakefile` to do a release, so not that often anyway
  * I prefer to have a dirty git working copy after the `rake install` and handle the reset manually than having `rake` do a `git checkout main.swift` to reset it… and potentially accidentally also revert other changes that we might have forgotten to commit before calling `rake install` (another option would be to ensure that everything has been committed to git before authorizing to continue to `rake install` but ¯\_(ツ)_/¯ )

So, in summary, commit the value of the `relativePath` that matches the relative path when the .app bundle stores the templates once compiled, so that a Cmd-R (and Cmd-U) works out of the box without changes from Xcode, and keep the Rakefile behavior as before, still letting it change the `templateRelativePath` during the `rake install` to match where the templates are asked to be installed according to the `rake install[…]` parameters, which would still default to `../templates`.

ahtierney:
@AliSoftware incorporated your feedback and rebased again 👍

swiftgen-cli/main.swift
```diff
@@ -10,8 +10,7 @@ import PathKit
 
 // MARK: Common
 
-let templatesRelativePath = "../Resources/templates"
-
+let templatesRelativePath = (Bundle.main.resourcePath ?? "") + "/templates"
```
AliSoftware: I'd use `?? "."` instead. Because in case `Bundle.main.resourcePath` returns `nil`, we would end up pointing to the root of the filesystem, which seems dangerous :wink:

AliSoftware (changes requested):

AliSoftware:
Just a small change requested and I think (at least from what I read from GitHub) that we're good to go.

Maybe if I have more time later I'll pull it and test and run on my Mac in Xcode (or if anyone can do it too to validate that in practice) before we (finally!) merge.

AliSoftware:
Wait… I think at some point I disabled the `EMBEDDED_CONTENT_CONTAINS_SWIFT` build settings back in the day when I configured SwiftGen as a command-line target (because Xcode doesn't know how to embed the Swift dylibs if the build product isn't packaged as a bundle). Btw, this caused an execution of `pod install` or `pod update` to generate a warning, because the project overwrote this build settings (while we shouldn't, or at least should `$(inherit)`)

I'm surprised not to see that Build Setting being changed in this PR's diff? Shouldn't this (well, more likely the new `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES` replacing it) be present again now?

🤔 I definitely will have to pull this on my Mac and inspect/test this. Diff'ing a pbxproj directly on GitHub isn't easy ^^

djbe:
I can confirm that `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES` is set to `Yes`.

It runs from within Xcode, and also runs when moving the produced app (and deleting the other produced content such as frameworks), so there are definitely no dependencies left.

ahtierney:
@AliSoftware Good catch on the root, I've rebased and pushed the fix. Re: the `.pbxproj` file, for some reason it's been collapsed in this pull request (maybe this is what you're referring to?) there are changes and `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES` is flagged set to yes. Let me know if you'd like it to `$(inherit)` (I believe there is still a warning when you `pod update/install`).

AliSoftware:
Indeed, I think we should be sure to get rid of the `pod update/install` warning.
Probably no need to set the value to `$(inherited)`, just remove the value completely from the _Target_ level so that we just don't override the one set on the _Config.File_ level.

ahtierney:
@AliSoftware agree! pushed the fix.

AliSoftware (approved):
Still haven't have time to git pull that branch to test it in practice in Xcode, but if you assure me that it now doesn't warn upon `pod install` / `pod update` anymore now that `ALWAYS_EMBED_SWIFT_LIBRARIES` isn't overridden, then we should be good to go!

djbe:
This is the output I'm getting with the latest code:

> ~/D/L/SwiftGen2 ❯❯❯ pod update                                                       feature/ahtierney/app-bundle
> Update all pods
> Updating local specs repositories
> Analyzing dependencies
> Fetching podspec for `GenumKit` from `GenumKit`
> Downloading dependencies
> Using Commander (0.6.0)
> Using GenumKit (4.0.0)
> Using PathKit (0.7.0)
> Using Stencil (0.6.0)
> Generating Pods project
> Integrating client project
> Sending stats
> Pod installation complete! There are 4 dependencies from the Podfile and 4 total pods installed.
> ~/D/L/SwiftGen2 ❯❯❯

AliSoftware:
I've finally got some time to pull and test in Xcode. Great work!

AliSoftware:
Just opened a separate issue as testing #204 in Xcode makes me realise that the ANSI color codes are printed in the Xcode console even if Xcode console doesn't support colors (RIP XcodeColors plugin in Xcode 8 😢 ), but that's a separate issue.

AliSoftware:
I'd suggest to remove the `XcodeColors` env var (as we now use Xcode 8 so no more XcodeColors) and add some sample launch arguments to make it easy to toggle them and test the various scenarios from Xcode.

I've made the changes locally, here's what the diff looks like. @ahtierney I can't push on your fork (I thought I've read some info on GitHub allowing contributors to push on a foreign fork of a PR? But maybe I'm wrong), so I'll let you apply that patch and push it if you have time. Otherwise that's not a big deal, we could make that change in master once merged.

```diff
diff --git a/SwiftGen.xcodeproj/xcshareddata/xcschemes/swiftgen-cli.xcscheme b/SwiftGen.xcodeproj/xcshareddata/xcschemes/swiftgen-cli.xcscheme
index 87f3c10..a350b98 100644
--- a/SwiftGen.xcodeproj/xcshareddata/xcschemes/swiftgen-cli.xcscheme
+++ b/SwiftGen.xcodeproj/xcshareddata/xcschemes/swiftgen-cli.xcscheme
@@ -14,8 +14,8 @@
             buildForAnalyzing = "YES">
             <BuildableReference
                BuildableIdentifier = "primary"
-               BlueprintIdentifier = "09A87B401BCC9B8000D9B9F5"
-               BuildableName = "swiftgen"
+               BlueprintIdentifier = "234F7E6D1DEBD5C5001B3C10"
+               BuildableName = "swiftgen.app"
                BlueprintName = "swiftgen"
                ReferencedContainer = "container:SwiftGen.xcodeproj">
             </BuildableReference>
@@ -85,13 +85,28 @@
             ReferencedContainer = "container:SwiftGen.xcodeproj">
          </BuildableReference>
       </BuildableProductRunnable>
-      <EnvironmentVariables>
-         <EnvironmentVariable
-            key = "XcodeColors"
-            value = "YES"
+      <CommandLineArguments>
+         <CommandLineArgument
+            argument = "colors -t swift3 $(PROJECT_DIR)/UnitTests/fixtures/colors.xml"
             isEnabled = "NO">
-         </EnvironmentVariable>
-      </EnvironmentVariables>
+         </CommandLineArgument>
+         <CommandLineArgument
+            argument = "images -t swift3 $(PROJECT_DIR)/UnitTests/fixtures/Images.xcassets"
+            isEnabled = "YES">
+         </CommandLineArgument>
+         <CommandLineArgument
+            argument = "storyboards -t swift3 $(PROJECT_DIR)/UnitTests/fixtures/Storyboards-iOS"
+            isEnabled = "NO">
+         </CommandLineArgument>
+         <CommandLineArgument
+            argument = "strings -t swift3 $(PROJECT_DIR)/UnitTests/fixtures/Localizable.strings"
+            isEnabled = "NO">
+         </CommandLineArgument>
+         <CommandLineArgument
+            argument = "fonts -t swift3 $(PROJECT_DIR)/UnitTests/fixtures/Fonts"
+            isEnabled = "NO">
+         </CommandLineArgument>
+      </CommandLineArguments>
       <AdditionalOptions>
       </AdditionalOptions>
    </LaunchAction>
@@ -104,9 +119,9 @@
       <MacroExpansion>
          <BuildableReference
             BuildableIdentifier = "primary"
-            BlueprintIdentifier = "09A87B401BCC9B8000D9B9F5"
-            BuildableName = "swiftgen"
-	    BlueprintName = "swiftgen"
+            BlueprintIdentifier = "234F7E6D1DEBD5C5001B3C10"
+            BuildableName = "swiftgen.app"
+            BlueprintName = "swiftgen"
             ReferencedContainer = "container:SwiftGen.xcodeproj">
          </BuildableReference>
       </MacroExpansion>
```

AliSoftware:
:tada:

PR #199, opened 2016-11-13
Switch the asset catalog parser to using "actool" to parse its contents

djbe:
This is a first step to eventually implementing the dot-syntax for asset catalogs, mentioned in #52.

The good thing about this is that, without changing the template, it already fixes an issue with the generated code for catalogs with namespaces folders. Currently the generated enum cases only contain the last filename of imagesets (eg. `case myImage = "myImage"`), but if they are contained in namespaced folders, the case should actually be `case namespace1_namespace2_myImage = "namespace1\namespace2\myImage"`.

This is a preliminary pull request, I still need to add a changelog entry and a bunch of unit tests. But I'd like a quick review of the changed code and your input before continuing.

Issue #52, opened 2015-11-04
Assets Improvement: Make separate enums for image folders


gsimmons:
Like the Storyboard generator, it would be nice to have separate enums for icon folders. i.e., if I have all my main view icons in a MainViewIcons folder, it would be nice to reference them like so:
UIImage.Asset.MainViewIcons.Apple.image

AliSoftware:
The only difficulty here is if people have multiple levels of folders, like if the user have some images at the top level, then some folders at top level, containing both images and subfolders, containing both images and subfolders, etc.

Stencil doesn't have a `{% while %}` node yet that would allow to do a template that can manage that arbitrary depth. But in the meantime, we could at least handle only one level of folders, which is probably the most common case.

gsimmons:
I think handling one level would be quite reasonable. Thanks!

ChristopherRogers:
I also would like to note that this is needed to fix an issue where the incorrect image name string is generated if you use the "provides namespace" option on a folder. ("properties.provides-namespace" in the Contents.json file)

AliSoftware:
Wow I didn't even know that option existed @ChristopherRogers, TIL, thx!
Will indeed have to take that into account!

woloda375:
+ 1

mgorbach:
Also a great idea, and it is very similar to https://github.com/AliSoftware/SwiftGen/issues/93. I've got some work going on that, and once that's merged, the ideas and possibly the code should be able to be reused to do this, as well.

Dimentar:
+1

djbe:
This has been fixed, should be released really soon (in 4.1.0).

PR #206, opened 2016-11-21
Images: dot syntax

djbe:
Follow up of #199, implements the dot-syntax for asset catalogs (mentioned in #52).

For now only has the swift3 template, I want some input before I do the swift 2 version.


UnitTests/TestSuites/ImagesTests.swift
```diff
@@ -80,5 +80,27 @@ class ImagesTests: XCTestCase {
 
     let expected = self.fixtureString("Images-File-CustomName.swift.out")
     XCTDiffStrings(result, expected)
-  }
+	}
+
+	func testFileWithDotSyntax() {
```
AliSoftware: Seems like you don't have the same indentation settings as the ones defined in the project (2-spaces indentation)?

GenumKit/Parsers/AssetsCatalogParser.swift
```diff
@@ -41,34 +59,41 @@ private enum AssetCatalog: String {
...
+        result += [Entry(name: imageName, value: "\(prefix)\(imageName)")]
       } else {
         // this is a group/folder
-        let children = item[AssetCatalog.children.rawValue] as? [[String: AnyObject]] ?? []
+        let children = item[AssetCatalog.children.rawValue] as? [[String: Any]] ?? []
 
         if let providesNamespace = item[AssetCatalog.providesNamespace.rawValue] as? NSNumber,
-            providesNamespace.boolValue {
-          processCatalog(items: children, withPrefix: "\(prefix)\(filename)/")
+          providesNamespace.boolValue {
```
AliSoftware: Doesn't Swift automatically bridge boolean `NSNumber` to `Bool` with `as?`? Especially since the `AnyObject -> Any` proposal got in?

I mean, we could probably do `if let providesNamespace = item[AsstCatalog.providesNamespace.rawValue] as? Bool` directly maybe? (At least it's worth a try)

GenumKit/Parsers/AssetsCatalogParser.swift
```diff
@@ -41,34 +59,41 @@ private enum AssetCatalog: String {
 extension AssetsCatalogParser {
   static let imageSetExtension = "imageset"
 
-  fileprivate func processCatalog(items: [[String: AnyObject]], withPrefix prefix: String = "") {
+  fileprivate func process(items: [[String: Any]], withPrefix prefix: String = "") -> [Entry] {
+    var result = [Entry]()
+
     for item in items {
       guard let filename = item[AssetCatalog.filename.rawValue] as? String else { continue }
```
AliSoftware: I totally forgot to see that in the other PR about actool migration, but if you're only using the `enum AssetCatalog` as a way to group constants together (and never use the `cases` other than via their `rawValue` to get the String constant), I'd suggest you to use an `enum` with no cases, but with `static let` constants namespaced in the enum instead:

```swift
private enum AssetCatalogKeys {
  static let children = "children"
  static let filename = "filename"
  static let providesNamespace = "provides-namespace"
  static let root = "com.apple.actool.catalog-contents"
}
```

AliSoftware (changes requested):

UnitTests/TestSuites/ImagesTests.swift
djbe: I don't know, sometimes Xcode forgets them, or something else.

Personally I always use tabs so that's why it sometimes gets through, even though I try to be careful.

GenumKit/Parsers/AssetsCatalogParser.swift
```diff
@@ -8,24 +8,42 @@ import Foundation
...
+
     // process recursively
-    processCatalog(items: items)
+    entries = process(items: items)
+  }
+
+  struct Entry {
+    var name: String
+    var value: String
+    var items: [Entry]?
```
djbe: Haven't used enums with associated values yet (although I've read about them). I'll look into it.

GenumKit/Parsers/AssetsCatalogParser.swift
djbe: So we'd be using the enum just as some sort of way to namespace constants?

GenumKit/Parsers/AssetsCatalogParser.swift
djbe: I think this was written before the swift3 merge. Don't know, but you're right, it should be bridged.

GenumKit/Parsers/AssetsCatalogParser.swift
```diff
@@ -41,34 +59,41 @@ private enum AssetCatalog: String {
...
+
+    return result
   }
 }
 
 // MARK: - ACTool
 
 extension AssetsCatalogParser {
-  fileprivate func loadAssetCatalog(at path: String) -> [[String: AnyObject]]? {
+  fileprivate func loadAssetCatalog(at path: String) -> [[String: Any]]? {
```
djbe: Will do.

GenumKit/Parsers/AssetsCatalogParser.swift
AliSoftware: I think the first solution (a simple enum with associated values) would be easier to use than a struct with enum inside.

Using a `struct` with inner `enum` is only useful if all cases of the enum share multiple properties, like enough to want to avoid repeating each property in each enum case's associated values.
But here we only have one common property so I think it's not worth it and integrating `name` directly in each enum case would be easier to manipulate.

And that way you won't need dedicate inits, you can just create `Entry.image(name: "foo", value: "parent/foo")` and `Entry.namespace(name: "parent", entries: children)`.

PS: Note that I'm not sure if the `indirect` keyword is really needed. I think so, because the case references the enum being defined, but it's worth trying without to be sure we can't avoid it.

GenumKit/Parsers/AssetsCatalogParser.swift
AliSoftware: Yup. That's how I namespace all my constants in my projects now, and have been loving it!

Like I very often have stuff like this in my projects:

```swift
struct SomeModel {
  enum JSONKeys {
    static let firstname = "firstName"
    static let lastname = "lastName"
    static let dob = "dateOfBirth"
  }

  let firstname: String
  let lastname: String
  let dateOfBirth: String

  init?(json: [String: Any]) {
    guard
      let fn = json[JSONKeys.firstname] as? String,
      let ln = json[JSONKeys.lastname] as? String,
      let dob = json[JSONKeys.dob] as? String
      else { return nil }
    self.firstname = fn
    self.lastname = ln
    self.dateOfBirth = dob
  }
}
```

GenumKit/Parsers/AssetsCatalogParser.swift
```diff
@@ -93,8 +93,8 @@ extension AssetsCatalogParser {
 // MARK: - ACTool
 
 extension AssetsCatalogParser {
-  fileprivate func loadAssetCatalog(at path: String) -> [[String: Any]]? {
-    let command = Command("xcrun", arguments: "actool", "--print-contents", path)
+  fileprivate func loadAssetCatalog(at path: Path) -> [[String: Any]]? {
+    let command = Command("xcrun", arguments: "actool", "--print-contents", String(describing: path))
```
AliSoftware: Doesn't PathKit's `Path` have a dedicated property to get the path string?
I mean, `String(describing:)` feels more intended for debug representation purposes to me, while something like `path.path` or `path.string` if it exists would feel more right (a bit like you'd use `url.absoluteString`)

AliSoftware (changes requested):
Don't hesitate to *rebase* on top of master, instead of merging master into your branch.

Both have pros and cons, but merging master into your branch means that you can get quite a 🍝  git tree sometimes. And it's then harder to rebase stuff like a PR that has been opened on top of another (like 204 is on top of 201) and make the latest PR only contain the diff commits (instead it adds more 🍝 )

Sure, rebasing instead of merging also have cons, like rewriting history meaning that other people who already have pulled your commit won't have valid commits anymore, but on a feature branch used for a PR, and on which typically only one people commits onto, it seems less a problem than the git tree 🍝 brought by merging master. So sure it's a matter of preferences but I prefer clean trees personally :wink: 🌲

GenumKit/Parsers/AssetsCatalogParser.swift
djbe: You'd think that, wouldn't you? There isn't any public one, only internal methods unfortunately.

AliSoftware: I think `path.description` feels more like a property on `Path` while `String(describing: path)`, although doing the same thing, feels more like a conversion or builder to me.

djbe:
What we can do (and kylef from Commander does this), is when merging the pull request to do "squash and merge", that way you'll always have a clean clean tree.

AliSoftware:
:+1:

GenumKit/Parsers/AssetsCatalogParser.swift
```diff
@@ -8,81 +8,162 @@ import Foundation
...
+    let found = entries.contains {
+      if case let .image(imageName, _) = $0, imageName == name {
       return true
+    } else {
+      return false
     }
   }
 
+  guard !found else { return false }
+  entries.append(Entry.image(name: name, value: name))
```
AliSoftware: indentation (from lines 17 to 26) seems wrong :wink:

GenumKit/Parsers/AssetsCatalogParser.swift
```diff
@@ -8,81 +8,162 @@ import Foundation
...
+   The output of the utility is a Plist and should be similar to this:
+   ```xml
+   <?xml version="1.0" encoding="UTF-8"?>
+   <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+   <plist version="1.0">
+   <dict>
+     <key>com.apple.actool.catalog-contents</key>
+     <array>
+       <dict>
+			 <key>children</key>
```
AliSoftware: ⚠️ tabs indentation

UnitTests/expected/Images-File-Dot-Syntax.swift.out
```diff
@@ -0,0 +1,64 @@
...
+    static let Mango: AssetType = "Exotic/Mango"
+  }
+  static let Lemon: AssetType = "Lemon"
+
+  struct Round {
+    static let Apricot: AssetType = "Round/Apricot"
+    static let Orange: AssetType = "Round/Orange"
+
+    struct Red {
+      static let Apple: AssetType = "Round/Apple"
```
AliSoftware: Why isn't the value of this generated constant `"Round/Red/Apple"` instead of just `"Round/Apple"`?
Is it because the group in the asset catalogs doesn't have provides-namespace checked, so that's normal and intentional?
In that case I'm wondering if we should create the intermediate structure `struct Red` if it doesn't provide namespace. I'm thinking YES (and so in this case this output is 👍 ), as otherwise the developer wouldn't have created a group in its asset catalog… but just wanted to start the discussion in case anyone had a different opinion.

AliSoftware (changes requested):
* Fix indentation
* Add CHANGELOG entry

djbe:
BTW, I based the templates on the strings dot syntax. Any reason why we use nested `struct`s instead of nested `enum`s for namespacing?

UnitTests/expected/Images-File-Dot-Syntax.swift.out
AliSoftware: :+1:

AliSoftware:
I think it's just because I overlooked the PR implementing those templates at that time, would indeed seem better to use enums instead of structs here.

djbe:
Did this maybe depend on swift 3?

AliSoftware:
Nah I don't think it did. enums as namespaces was already possible in Swift 2…

AliSoftware (approved):

AliSoftware:
Conflict to fix so we can merge 😉

djbe:
Oh you want to do a minor release instead of patch? Sure, gimme a minute.

AliSoftware:
Ah, didn't realize that this bumped to the minor version… why is that btw?

djbe:
New feature? (new template)

AliSoftware:
I mean, before this PR, people using namespaces didn't have the dot namespace in the generated code, now they will… that means breaking, right?

djbe:
No the old workflow still works

AliSoftware:
Oh, sorry, you're right, I thought we already had a (non-working) dot-syntax template for Images, (I got confused with the one we already have for Strings). If that's a new template then yeah, minor it is 😉

In that case yeah, maybe we should do a patch version real quick first, before merging this and starting a new minor, right.

djbe:
Yeah, I mean, I still have to check this works from command line 😅 
`-t somethingsomething` to switch templates right?

AliSoftware:
yup!

djbe:
So, I think I'll wait until 4.0.1 is released before I fix the conflict in this, otherwise I'll keep having to merge changes.

AliSoftware:
Yup, probably a good idea indeed 😉

Can't wait for #204 to be able to test those command line invocations right from Xcode instead of having to launch the Terminal and `rake install` for that 😉

_(Will review #204 rn btw, it's about time we move forward on this, would help a lot the dev process)_

djbe:
Definitely! 🎉

PR #208, opened 2016-11-21
Escape newlines again in .strings file keys

ChristopherRogers:


djbe:
Could you add a unit test for this (or modify existing ones? Also please add a changelog entry.

AliSoftware:
I'm a bit surprised by that change: In which case would **keys** need newline-escaping?

ChristopherRogers:
We have a strings file that was generated (I think) using the traditional tool that scans for uses of NSLocalizedString() and sets the key to the orgiinal translation. So if the original translation had \n in it, it remains in the key.

AliSoftware:
Ok. Seems like a strange process to me (to have keys that are actual text potentially containing diacritics, newlines, and even potentially emojis), but that doesn't hurt to escape them for all of those who don't have them I guess 😉

There seems to be conflicts with `master` since this has been submitted though. Could you fix the conflict so we can merge this?

Also, you might want to add an entry in the CHANGELOG to credit yourself too :wink:

AliSoftware (changes requested):
* Please add at least one Unit Test for this use case
* Please add an entry to mention the change and credit yourself in the CHANGELOG

ChristopherRogers:
@djbe Thanks.

AliSoftware (approved):

PR #212, opened 2016-11-26
Consolidate use of PathKit

djbe:
Just did a quick review of places where paths get passed around, and tried to ensure that actual `Path` instances get used there.

The code in FontsFileParser can be simplified further pending https://github.com/kylef/PathKit/pull/25.

Hopefully I haven't missed too many, definitely needs a second look.


AliSoftware:
Nice cleanup :+1:

Issue #207, opened 2016-11-21
Multiple asset catalogs - no image found


samritchie:
I have multiple asset catalogs in my project. If I run `swiftgen images $PROJECT_DIR` I get the output `no image found`.

If I explicitly specify an asset catalog, Swiftgen will output code, however when I run multiple `swiftgen images` commands, even if I use `--enumName` I still get build failures (`Invalid redeclaration of Image`). 

Swiftgen happily handled multiple asset catalogs in the past (albeit flattened into a single enum), so I assume this is a bug introduced with the actool change?

AliSoftware:
That's right, now that swiftgen don't parse directories in the filesystem for images but a specific assets catalog that's why you now have only one assets catalog output at once… and running it multiple times, given that the template generates a `typealias Image= UIImage` each times, that explains the compiler error you have.

Until we fix this you can duplicate the images template (`swiftgen templates cat images-swift3 >images-notypealias.stencil` or whatever images template you're using), remove the typealias declaration from your copy, then use that template (with `--templatePath`) for all asset catalogs except the first one.

I don't know what's the best way we will come up to fix that. Probably reintroduce the ability to specify a directory to parse, even if that will now look for all asset catalogs in that directory and still use actool to parse each one…? (\cc @djbe)

samritchie:
Okay, thanks Olivier. I’d expect if your Swiftgenning one asset catalog, most of the time you’ll want to do them all, and see them namespaced somehow? 

I might just consolidate into a single set if this is not well supported.

djbe:
Damn, sorry we missed that use case. 

I'm thinking more along the lines of #33, where you provide multiple path values. Recursively parsing a whole folder structure is far from ideal, it's better if you tell swiftgen where to look.

I'll take a look at this after #206, as there'd be too much overlap otherwise.

AliSoftware:
This is to be confirmed with @kylef (and @peterlivesey who has probably asked him already for another feature he's working on) but iirc, Commander doesn't support specifying the same flag multiple times, so except if we fork and PR Commander to add this feature (which would be great), so far there's no way to specify multiple `--path` values. Another workaround would be to accept a single param with all the paths concatenated and separated with e.g. a comma, but that would not seen pretty to me, so I'd prefer a PR on Commander for that instead 😉

djbe:
PR Commander it is then 😆

djbe:
[...]

Issue #33, opened 2015-10-14
Accept multiple PATH in the CLI


AliSoftware:
We should accept an infinite number of input PATH from the CLI (instead of only one), and iterate over all of them to feed our `EnumBuilder` before generating a big output containing them all

AliSoftware:
@kylef How could we achieve this (parsing an arbitrary number of args at the end of the command) using `Commander`?

kylef:
@AliSoftware Just to clarify, is that with an argument or a flag?

```
tool --path a b c
```

```
tool a b c
```

AliSoftware:
Well both actually, I expect `swiftgen` to accept optional flags like, say, `--verbose`, accept optional arguments like `--output path`, and once all are consumed, I want every `Argument` left to be parsed.

```
$ swiftgen PATH1 PATH2 PATH3
$ swiftgen --verbose PATH1 PATH2 PATH3
$ swiftgen --output OUTPATH PATH1 PATH2 PATH3
```

I expect to be able to write something probably like this:
 - `command(…) { verboseFlag, output, pathsArray in … }`
or maybe
 - `command(…) { verboseFlag, output, parser in … let strings: [String] = parser.parseRemaining() }`

Is this kind of stuff possible?

AliSoftware:
Implemented by @djbe in #213

PR #213, opened 2016-11-26
Multiple paths

djbe:
Covers issues:
- #207 regression with multiple catalogs
- #33 multiple paths

Aditionaly relates to:
- #32 strings directory input
- #39 localisable table name

I've now hit a wall in the easy implementation bit:
- [x] Fonts just adds entries to one big structure, no need to split up by path.
- [x] Same for storyboards, especially because value collisions of scene identifiers (and storyboard names) are not allowed (they generate errors/warnings).
- [x] Images (asset catalog) could be done easily. `Image(named: "...")` does not care in which catalog something is, and name collisions (I assume) generate warnings. But it might be interesting for the user to know which catalog he/she is accessing. We could just add another folder level, but this would depend on #206). We'd only add an extra level if there are multiple catalogs.
- Colors could be namespaced or not, depends on what might be preferred. Same situation as with Images, because I could have 2 files with the color "orange" but they might actually have different values. _Postponed to later PR_
- Strings need to be namespaced, namely we'd need to use the correct `tableName` (see #39 and #32). _Postponed to later PR_

We could just leave some of these open for later, especially the strings one because it first needs more work to support `tableName`s.


AliSoftware:
You're on fire! 👍
I'll take a look in the afternoon, thanks again!

Podfile
```diff
@@ -2,7 +2,7 @@ platform :osx, '10.9'
 use_frameworks!
 
 target 'swiftgen' do
-  pod 'Commander', '~> 0.5.0'
+  pod 'Commander', :git => 'https://github.com/kylef/Commander.git'
```
AliSoftware: 0.6.0 is out 🎉

swiftgen-cli/storyboards.swift
```diff
@@ -15,19 +15,21 @@ let storyboardsCommand = command(
     description: "The name of the enum to generate for Scenes"),
   Option<String>("segueEnumName", "StoryboardSegue", flag: "g",
     description: "The name of the enum to generate for Segues"),
-  Options<String>("import", [], count: 0,
+  VariadicOption<String>("import", [],
```
AliSoftware: 🤔 I guess that means that #175 wasn't actually working as described in the CHANGELOG entry until now… (« Multiple imports can be added by repeating this flag. ») and that actually fixes it then?

AliSoftware:
My thoughts on your questions in your original comment:

### Images

> We'd only add an extra level if there are multiple catalogs.

Seems good to me. But be sure to make it as a separate key in the `StencilContext`, so that people would have the chose to separate it or not in their own templates.

That logic of adding an extra level should be done at the template level imho, with the `StencilContext` probably having a root key `assetCatalogs` containing an array of Dictionaries (`(name, assets)` pairs), one for each Asset Catalog with their name and content. And then the template would iterate over them; using the `assetCatalogs[i].name` or not in the template being up to the template itself, and using it only if `assetCatalogs.count > 1` seems a good idea for the bundled templates.
(You might need to add an extra `hasMultipleCatalogs: Bool` root key to the StencilContext as I think it's not possible in Stencil to compare values using `> 1` just yet?)

### Colors

Colors being generated by a file entirely written manually (as opposed to other commands relying on parsing directories and Xcode/IB files), I'm not sure a lot of people have _multiple_ colours input files 🤔 

But since we'd probably already have the logic for the Images, it's probably worth doing the same logic for Colors indeed.

### Strings

As you said, name spacing Strings is probably worth being in its own separate PR to handle `tableNames`.

swiftgen-cli/storyboards.swift
djbe: Yeah, you can imagine how I felt when I tried the release 😱 

I'd completely misinterpreted the purpose of `Options<>` which needs a count for the number of arguments it should expect. This fixes that issue 😅

djbe:
- Images: agreed. I'd let it depend on #206 as it already has all the structure stuff, just add an extra enum case.
- Colors: that's what I though, but it has a bit more implications as right now, a color parser is a specific subclass for a specific file type. I'd need to create a more general color parser type, that internally creates the correct parsers for each file type.
- Strings: later.

swiftgen-cli/storyboards.swift
AliSoftware: Yeah I know that @plivesey had the same misunderstanding of this `Options<…>(count: )` stuff in `Commander` (and so did I when we talked about it)… not very intuitive ^^

AliSoftware:
Another thing that would be nice to add in this PR would be to generate an error in case the path(s) given to `swiftgen images` isn't an `.xcassets` directory. That way, people that used to do `swiftgen images $(PROJECT_DIR)` would now have an explicit error message rather than ending with #207

djbe:
I'd actually put that in a separate PR, coming up!

djbe:
Tried a rebase this time. Also tried it with #206 but that just gave me headaches 😐

AliSoftware:
Is this really done and ready to merge now? (I mean, for what I've reviewed in GitHub it seems OK to me, but given that the original comment of this PR still has unchecked boxes I don't want to merge on your behalf if you planned to add more to this PR or will do them in separate PRs.

But if you're good, I'll let you the pleasure of hitting the Merge button 👍

djbe:
No, I want to at least also do the images one, to fix #207. Just leave it for a minor release. 

The rest will have to go in other PRs.

AliSoftware:
Ok. Added the WIP label to mark it as unfinished, feel free to remove once ready 👍

djbe:
Right, so images was a bit more annoying than expected.

Both images and strings `dot-syntax` should be rewritten with some sort of recursion using the `set` tag. I've noticed that the modified context is stack dependent, so if you set a variable inside an `if` block, it won't be visible outside of that block. Which is good news for recursion.

The current templates contains code that depends on unreleased Stencil code (https://github.com/kylef/Stencil/pull/77), but compiles without issues for now. Multiple catalogs' entries will for now appear under one enum. Once we have the new Stencil, we can expect it to automatically work, and will just need an updated output file.

AliSoftware:
Does that mean that we should wait for the new version of Stencil containing kylef/Stencil#77 to be out to release a version of SwiftGen with this #213 PR? As otherwise we would do breaking changes twice (one with the images template containing the extra level even if there's only one assets catalog, then later the one removing it)?

kylef:
I'm happy to make a Stencil 0.8 (or maybe 1.0-rc.1 🤔) release soon to support this if needed.

djbe:
Yeah, I suggest waiting until we can merge a PR supporting the latest Stencil.

AliSoftware:
👍

PR #217, opened 2016-11-29
Throw an error when the provided path is not an asset catalog

djbe:
Title says it all.

### DIFF
--- a/swiftgen-cli/images.swift
+++ b/swiftgen-cli/images.swift
@@ -15,7 +15,13 @@ let imagesCommand = command(
   Argument<Path>("DIR", description: "Directory to scan for .imageset files.", validator: dirExists)
 ) { output, templateName, templatePath, enumName, path in
   let parser = AssetsCatalogParser()
-  parser.parseCatalog(at: path)
+
+  switch path.extension {
+  case "xcassets"?:
+    parser.parseCatalog(at: path)
+  default:
+    throw ArgumentError.invalidType(value: String(describing: path), type: "xcassets file", argument: nil)
+  }
 
   do {
     let templateRealPath = try findTemplate(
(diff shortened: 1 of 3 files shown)


swiftgen-cli/images.swift
```diff
@@ -15,7 +15,13 @@ let imagesCommand = command(
   Argument<Path>("DIR", description: "Directory to scan for .imageset files.", validator: dirExists)
 ) { output, templateName, templatePath, enumName, path in
   let parser = AssetsCatalogParser()
-  parser.parseCatalog(at: path)
+
+  switch path.extension {
+  case "xcassets"?:
```
AliSoftware: Wouldn't an if/else be more simple here instead of switch with only one case?

swiftgen-cli/images.swift
```diff
@@ -14,15 +14,13 @@ let imagesCommand = command(
   Option<String>("enumName", "Asset", flag: "e", description: "The name of the enum to generate"),
   Argument<Path>("DIR", description: "Directory to scan for .imageset files.", validator: dirExists)
 ) { output, templateName, templatePath, enumName, path in
-  let parser = AssetsCatalogParser()
-
-  switch path.extension {
-  case "xcassets"?:
-    parser.parseCatalog(at: path)
-  default:
+  guard let `extension` = path.extension, `extension` == "xcassets" else {
```
AliSoftware: Haha I was more thinking about `guard case "xcasstes" = path.extension` but that works too ^^

AliSoftware (approved):

swiftgen-cli/images.swift
djbe: Ha, no, you're right, for some reason I automatically write unwrap code 😅

AliSoftware:
Oh is Travis starting to *finally* waking up? (So frustrating to wait for the CI so long!)

djbe:
Yeah, they probably had an enormous backlog of jobs to get through.

PR #224, opened 2016-12-05
Clean up some warnings (xcode and lint)

djbe:
I don't think this warrants a changelog entry...


GenumKit/Parsers/ColorsFileParser.swift
```diff
@@ -187,7 +187,7 @@ public final class ColorsXMLFileParser: ColorsFileParser {
                       namespaceURI: String?, qualifiedName qName: String?) {
       guard elementName == ColorsXMLFileParser.colorTagName else { return }
       guard let colorName = currentColorName, let colorValue = currentColorValue else { return }
-      parsedColors[colorName] = try? parse(hex: colorValue) ?? UInt32(0)
+      parsedColors[colorName] = try? parse(hex: colorValue)
```
AliSoftware: Do we want `parsedColors` dictionary to contain a `nil` in case the parsing is invalid (like `colorValue` is `foobar`), or should it contain `0`?

If the latter (use `0`), we should still coalesce using `??`, but the on the whole `try?` result:
```
parsedColors[colorName] = (try? parse(hex: colorValue)) ?? UInt32(0)
```
This way if the parse fails and throw, we store 0.
Or that could be a good occasion to actually print a proper error instead (= use `do/try/catch` and not `try?`)

GenumKit/Parsers/StringsFileParser.swift
```diff
@@ -39,7 +39,7 @@ public final class StringsFileParser {
     let plist = try PropertyListSerialization
         .propertyList(from: data, format: nil)
 
-    guard let dict = plist as? Dictionary<String, String> else {
+	guard let dict = plist as? [String: String] else {
```
AliSoftware: Indentation

GenumKit/Parsers/ColorsFileParser.swift
```diff
@@ -187,7 +187,15 @@ public final class ColorsXMLFileParser: ColorsFileParser {
...
       guard elementName == ColorsXMLFileParser.colorTagName else { return }
       guard let colorName = currentColorName, let colorValue = currentColorValue else { return }
-      parsedColors[colorName] = try? parse(hex: colorValue) ?? UInt32(0)
+
+      do {
+        parsedColors[colorName] = try parse(hex: colorValue)
+      } catch let e as ColorsParserError {
+        fatalError(e.description)
+      } catch {
+        fatalError("error: unknown color parser error")
```
AliSoftware: 🚨 fatalError is gonna make SwiftGen crash. Not just print the error cleanly to the user and properly exit

AliSoftware: I see 3 possible solutions here:

* We use `(try? parse(hex: colorValue)) ?? UInt32(0)` and don't error. Seems like a bad solution as there won't be any error message to the user
* We do as you suggest, with `parser.abortParsing()` and store the error in some property to check it at the end outside of the parser delegate.
* We transform the `var parsedColors = [String: UInt32]()` in the delegate into a `var parsedColors = [String: ColorResult]()` with `enum ColorResult { case color(UInt32), error(ColorsParserError) }` so we can store either a valid color UInt32 or an error then handle that at the end of the `parseFile` function.

The first solution is easier a d quicker to implement. The second solution would allow us to continue parsing and report _all_ errors at the end. And still even allow the output file to be generated even (with partial output) if we want to (but we better match the behavior for all color parsers then)

AliSoftware: Given that, if I'm not mistaken, I believe that the other parser fail as soon as they encounter an invalid hex value in the input file, solution 2 is probably the way to go, right?

AliSoftware: BTW, we should add a unit test with an invalid XML file to catch and test that use case!!

djbe: Sure thing, I propose I create a new PR for this?

PR #225, opened 2016-12-06
Added Swift 3 storyboard templates for OSX

FelixII:
I've migrated the existing OSX storyboard templates to the new Swift 3 syntax.


djbe:
Thank you for the PR. Is there a reason that both `default` and `lowercase` exist for swift3? I think only the lowercase one is needed.

On another (related) topic:
Can't we just unify the `osx` and `ios` templates into one like the other templates? I think that'd be a more worthwhile endeavour instead of adding more templates.

Couple of missing things:
- Unit tests for each template.
- Changelog entry.

tbaranes:
> Can't we just unify the osx and ios templates into one like the other templates? I think that'd be a more worthwhile endeavour instead of adding more templates.

At the beginning, we searched a solution to have a common templates for all the gen files, but we concluded that storyboards will be an exception since the use case differ a bit from iOS whereas strings, colors, fonts are the same. You will find all the details to this in #101 and the PRs where the storyboards have been implemented.

FelixII:
> Thank you for the PR. Is there a reason that both default and lowercase exist for swift3? I think only the lowercase one is needed.

You are absolutely right, thank you! The `lowercase` template does only affect the enum value, which is lowercase in Swift 3 by default. I will remove the superfluous template and supplement the missing items shortly.

FelixII:
I've merged the two templates into one, added unit tests and a changelog entry. Please let me know if there is anything else missing or wrong.

djbe:
Looks good, just a small changelog issue. Unless someone else (@AliSoftware?) has any input, I think we can merge this.

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ _None_
 
 ### New Features
 
+* Added a `storyboards-osx-swift3` template.
```
djbe: Please add 2 spaces at the end of the line (after the `.`).

AliSoftware: I've pushed the changes for this, will merge as soon as Travis is OK 👍

PR #227, opened 2016-12-06
Better color parser error handling.

djbe:
Just a bunch of extra checks and corresponding unit tests.


GenumKit/Stencil/Contexts.swift
```diff
@@ -4,6 +4,8 @@
 // MIT Licence
 //
 
+// swiftlint:disable file_length
+
```
AliSoftware: :cry:

djbe:
Yeah, that'd be a better solution, that long file has been bothering me for a while 😄

PR #229, opened 2016-12-07
Fix compile error in swiftgen-cli

djbe:
Travis didn't catch this, because it doesn't build the swiftgen target, only the unit tests.

### DIFF
--- a/swiftgen-cli/colors.swift
+++ b/swiftgen-cli/colors.swift
@@ -20,7 +20,7 @@ let colorsCommand = command(
   switch path.extension {
   case "clr"?:
     let clrParser = ColorsCLRFileParser()
-    clrParser.parseFile(at: path)
+    try clrParser.parseFile(at: path)
     parser = clrParser
   case "txt"?:
     let textParser = ColorsTextFileParser()


AliSoftware:
This made me think of something (in addition to the fact that we should add `rake install` to the `.travis.yml` indeed).

Maybe we should change this `switch` selecting the right parser into a hash table / dictionary, would be easier to reason about I think, right? Given that every parser has the same signature anyway…

```swift
public protocol ColorsFileParser {
  var colors: [String: UInt32] { get }
  func parseFile(at path: Path) throws // Add this to the protocol
}
…

let parsersMap: [String: ColorsFileParser.Type] = [
  "clr": ColorsCLRFileParser.self,
  "txt": ColorsTextFileParser.self,
  "xml": ColorsXMLFileParser.self,
  "json": ColorsJSONFileParser.self,
]
if let parser = parsersMap[path.extension]?() {
     try parser.parseFile(at: path)
} else {
     throw ArgumentError.invalidType(value: path.description, type: "CLR, TXT, XML or JSON file", argument: nil)
}
```

djbe:
Related to #213, I think this should be moved to a generic color file parser that creates the correct sub-parser. Or something like that...

AliSoftware:
Yeah, like a Class Cluster, in short ;)

djbe:
Exactly!

Issue #231, opened 2016-12-07
[Internal] Review the templates to take advantage of the new features of Stencil 0.7


AliSoftware:
I think a lot of new features brought by Stencil 0.7+ could allow us to simplify our templates a lot, and get rid of some context variables that were introduced by GenumKit in the Context just because some constructs were not possible to be created in Stencil before, but now they are…

Examples include:

* in the StoryboardContext, there's a `isBaseViewController` value which indicate if the baseType is 'viewController' or anything else. Now that we can use `==` in templates, that's probably obsolete
* same for the Strings template. GenumKit exposes a context variable `typednames` which is an array containing typed declarations like `"let p0: String"`,`"let p1: Int"`. Had to add that because there was no easy way to construct that directly in the template from the other variables. But now that GenumKit has the `MapNode`, we should be able to construct those from the `types` and `names` context 
* Same for `declarations` which contains `let p0`, `let p1`, etc… that we can probably construct from the `names` with is an array of `p0`, `p1` and a `MapNode`
* Even the `names` variable in the String context might now maybe be constructible in the template itself using a `MapNode` and the `forloop.index` or something

djbe:
We could do another review-pass, but I think most of these have been covered in the 4.2 release?

AliSoftware:
Idk i think everything has been covered but better double check, e.g.

* did we properly get rid of our own `join` filter in templates and use Stencil's proper `join` filter, which is the same but with parameters and different default separator? (In other words did we specify the separator explicitly so that when we remove our own join filter it will continue to work)
* did we prepare everything to get rid of old tags and filters in StencilSwiftKit which were implemented when Stencil didn't support filters with parameters yet, so that we're ready in 5.0 to migrate them to cleaner implementations using parameters instead of having multiple similar filters without parameters?

djbe:
Item 1 is in https://github.com/SwiftGen/templates/pull/22.
Item 2 still needs to be done, you opened a separate issue for it: https://github.com/SwiftGen/StencilSwiftKit/issues/6

AliSoftware:
Right. Then this should be OK!

PR #233, opened 2016-12-07
Reworked camelCase to lowercase all-uppercased strings

### LINKED ISSUE #226: All leafs are complete-lowercased with dot-syntax-swift3 template
When using the `dot-syntax-swift3` template and having the key

```
WORD_POOL_ITEM_FORM.NOTIFICATION.QUICK_ADD_INFO
```
the usage in code currently is:
```
L10n.WordPoolItemForm.Notification.quickaddinfo
```

As you can see the last part is not camelCased anymore. I'd expect it to be `quickAddInfo` though.

Jeehut:
This attempts to fix #226 by lowercasing the tails on the `lowercase` method – without breaking already camel-cased strings. Please see the full commit messages for details on the rationale.

Note that strictly speaking this is a "breaking" change as it will change the structure of some cases, e.g. `acceptCGUScene` now becomes `acceptCguScene`. The latter was always what I was expecting from a camel-cased string though, so I think with the auto-completion help of Xcode which nowadays has automatic correction suggestions implemented for wrong cased names this should be acceptable for a minor version update and imho doesn't need to be changed long term.


UnitTests/expected/Fonts-Dir-CustomName-Swift3.swift.out
```diff
@@ -30,13 +30,13 @@ extension Font {
 }
 
 struct CustomFamily {
-  enum SFNSDisplay: String, FontConvertible {
+  enum SfNsDisplay: String, FontConvertible {
```
AliSoftware: Don't know about you, but I'm not a fan of this case…

AliSoftware (commented):
Given that with that changes, all uppercase acronyms are now TitleCased (`CGU` -> `Cgu`, `URL` -> `Url`), not sure if the current `lowerFirstWord`filter, that was the one taking care of being sure to keep the first word lowercased if it was such an acronym, still makes sense?

AliSoftware:
Thx for the PR!

Given that this PR is technically a breaking change, would probably only go in the next major release. 

_(As #231 would also probably be a major breaking change (changing the Stencil Context keys to simplify things a lot, but making existing custom templates break), both PRs will probably go in the same major release)_

UnitTests/expected/Fonts-Dir-CustomName-Swift3.swift.out
Dschee: Yeah, I agree on this one. Don't have the time to fix it, but maybe this is solvable by changing the template? Or the new camelcase method?

djbe:
Could you add extra Unit Tests for this? Go through each possible combination of uppercase/lowercase words and dividers, and what should come out the other end.

AliSoftware (commented):
:warning:

> note: This is quite similar to `capitalise` except that this filter doesn't lowercase
>          the rest of camel cased string but keeps it untouched.

* `"helloWorldHowAreYou".capitalized` returns `"Helloworldhowareyou"`
* `"helloWorldHowAreYou".titlecase` returns `"HelloWorldHowAreYou"`

So they're not strictly equivalent, so not sure that commit doesn't break some use cases…

Dschee:
If we can fix the multiple uppercased letters issue then I don't think this is a breaking change any more, other than that I agree that this should be in the next major release.

I won't have any more time to do it today though (it's evening here). I just added an entry to the changelog and refactored the code (by removing the titlecase method completely as it now reflects the same logic as `.capitalized`).

Dschee:
@AliSoftware That's exactly what I changed with this PR, I'm now making sure `helloWorldHowAreYou` is kept by going through the detour of snake-casing it first, then splitting by "_" and capitalizing it then. This is a pretty common strategy for camelCasing string. See the `titleize` method in Rails as an example [here](https://github.com/rails/rails/blob/f2489f493b794ee83a86e746b6240031acb8994e/activesupport/lib/active_support/inflector/methods.rb#L162) – they first `underscore` the string to keep camel cased structures.

Dschee:
@djbe There's already quite some cases in the tests:

```
let expectations = [
      "string": "String",
      "string_with_words": "StringWithWords",
      "someCapString": "SomeCapString",
      "": "",
      "x": "X",
      "a__b__c": "ABC",
      "__y_z": "__YZ",
      "PLEASE_STOP_SCREAMING!": "PleaseStopScreaming!"
    ]
```

They all pass (even with my change to `.capitalized`). But feel free to add more cases.

djbe:
Merged master, the tests will currently fail. We should check what we actually want as result, and work from there on how to implement it.

As it is, the tests both fail for both `titlecase` and `camelCase`, which definitely shouldn't happen.
Is `titlecase` meant to be capitalised? I think it should maintain uppercase letters inside the string, such as `SomeCapString` should remain `SomeCapString`.

Dschee:
@djbe I've had a look into your merged changes and I'm completely confused. I'm not sure we're understanding the same thing under camel-casing here. From a method called `snakeToCamelCase` I would expect that the input string would be treated as a snake-cased string and converted to a camel-cased one.

This also means that a string like "PLEASESTOPSCREAMING" would be expected to be a snake-cased string (a compound of only one word) even though there is no underscore. So after camel-casing it I would either expect it to be "Pleasestopscreaming" or "pleasestopscreaming" depending on if we use camel-caps style camel-casing or camel-back style camel casing. Also "PLEASE_STOP_SCREAMING" would then (using the camel-caps style) become either "PleaseStopScreaming". The downside of this method would be that something like "URLChooser" becomes either "Urlchooser" – but that's actually what I would expect the method to do.

Dschee:
As for titlecased I think you're right, "SomeCapString" should stay "SomeCapString" – thinking of it for a second time I think titlecase is meant to only capitalize the first letter and keep everything else as is.

Dschee:
Ahh, now I remember. The issue was that after implementing the snakeToCamelCase as I would expect it to behave I saw that it didn't keep already-camelcased strings as they were. That's why "string_wiTH_WOrds" becomes "StringWiThWOrds" – the string is first interpreted as already camelCased and converted afterwards. Let me see if I can find a solution for that ...

Dschee:
Okay, I think I found a satisfying solution now. Here's the trick: I'm checking the input string of the `snakeToCamelCase` method for lowercase characters. If it contains any then I'm simply using the old logic, if it is a full-uppercase string though I'm suggesting the user wants the string to be truly camel-cased and therefore lowercase the parts of it. Thanks to this I could also revert some of the expected output string files that I had changed before. Also note that I made titlecase work as before again.

I think this can be merged now! All problems should have been solved. 🎉

djbe:
I like it, this way we maintain existing behaviour and fix a bug for some specific cases. I think this is ready to be merged.

Note: I've re-added the titlecase function documentation that got lost somehow.

AliSoftware:
:+1: :tada:

PR #234, opened 2016-12-07
Update to Stencil development

kylef:
@djbe Here's a working version of SwiftGen on Stencil master. Probably some things we will want to change but here it is.


AliSoftware:
I think we talk about the concept of "Stencil contexts" in multiple places in the doc comments and in SwiftGen's documentation. We should be sure not to forget to update them as it seems this concept doesn't exist anymore (well at least doesn't have an explicit named for for it anymore)

djbe:
Can we squash & merge this? Or shall we wait until 0.8?

AliSoftware:
@djbe I think there are still references to Stencil contexts in the documentation/Templates.md 😉

Also, funny that since Stencil got rid of the concept of context, your change above got rid of the term Stencil… and not the term context 🤔

djbe:
Well the way I thought about it, it conceptually still is a context of variables, right? Just that we don't use a specific "Stencil Context" type anymore. You could just call it "dictionary", but that'd be really confusing.

AliSoftware:
Yeah i didn't check how @kylef calls them now in his documentation, everything is just called dictionary in the code and parameter names but not sure what name too give to the concept when explaining it in the docs s just using dictionary of variables indeed might not be clear enough and a dedicated term would probably be better for the explanations

kylef:
> @djbe I think there are still references to Stencil contexts in the documentation/Templates.md 😉

> Also, funny that since Stencil got rid of the concept of context, your change above got rid of the term Stencil… and not the term context

Stencil still has the Context, you still have access to from custom tags and use it for variable access. Just that now you don't manually create a `Context`, it is created during rendering because we store other information in there and it's unnecessary to add the complexity of a `Context` when rendering the template. A context life only spans across the template render, a user cannot re-use a context with multiple renders.

I think it's still fine to use the term "template context", but perhaps it would be clearer in documentation to use "template variables" as I think this term is more straight forward and clearer for template authors.

djbe:
Yeah, we discussed it for a while and thought it'd be better to revert the changes.

My way of thinking was that it's not a `StencilContext` anymore, but it's still a "context of variables", that's why I left the name of "context". Couldn't really choose "dictionary" as a name as that'd be too generic.

Maybe instead of `stencilContext` or `context`, we could go with `templateContext`?

AliSoftware:
I'm ok with keeping the _term_ "Stencil Context" (or parameter names like `stencilContext` maybe), we just have to be careful not to use the name of the old type `StencilContext` exactly as is (as this type doesn't really exist publicly anymore), but we can use the "Stencil Context" term instead.

But I agree using "template variables" is probably more clear, at least in the documentation, to explain the concept easier to newcomers and template authors indeed.

Issue #232, opened 2016-12-07
[Documentation] Add detailed description for each template


AliSoftware:
We should create a `templates/` subfolder in the `documentation/` folder (or provide Wiki entries for the same intent, but I prefer keeping my doc in the repo so they can be updated in the same scope as PRs) to describe each template.

One possible format could be to have one file for each template with the same name as the template name, with:
* an example output (or at least an output extract) of what is generated
* an example of how it would look like at call site
* a summary about when you might prefer to use this template / for which kind of use case this template is a better fit (e.g. Swift 2 vs Swift 3, or if you prefer structured keys, or if you organise your Localizable.strings in some way or another, …)

AliSoftware:
We should also document the whole SwiftGen infrastructure so that it's more clear for both contributors and template authors how the whole architecture works.


Maybe something like this os similar?
```
  Command Line
   Parameters
       |
       v
  +----------+
  | SwiftGen |
  |    CLI   |- - - - - - - - - - - - - - - - - - - - - - - -.
  +----------+           '                                   |
       |            subcommand                               v
                   selects the                         Template file
       |           right parser                              |
                         '                                   |
       |                 v                                   v
                  +-------------+      - - - - -       +------------+
       v          | SwiftGenKit |     ' Stencil '      |   Stencil  |
Input file(s) --->|    Input    |---->' Context '----> |  Template  | ---> Output file
                  |   Parser    |     '  (Dict) '      |   Engine   |
                  +-------------+      - - - - -       +------------+
                                                             ^
                                                             |
                                                     +-----------------+
                                                     | StencilSwiftKit |
                                                     +-----------------+

Legend:
* "SwiftGen CLI" interprets the Command Line arguments. the subcommand determines which parser to use, the other parameters include the path to the input file(s) and the template to use
* The "SwiftGen Input Parser" parses the input files (Localizable.strings, image catalog, fonts, etc…) and transform that input into a Stencil context, which is basically a Dictionary representation of the input files
* The "Stencil Template Engine" ("Stencil") injects that Stencil context (dictionary) into the chosen template to generate the output. The templates can use the variables provided by the context to customize the output, using the tags and filters provided both by Stencil itself and by the "StencilSwiftKit" framework which provides additional tags which are not part of Stencil but are useful additions to generate Swift code more easily
```

AliSoftware:
Suggestion for what each markdown file could look like to describe each template.

[...]

PR #237, opened 2016-12-11
Template recursion

djbe:
This defines 2 stencil tags:
 - `define`: This defines a `CallableBlock`, accepts the following arguments:
   - _name_: the block will be stored in a variable of this name (in the current context stack).
   - _parameter names_ (0 or more): will then be available inside the block using `args.parameterName`.
- `call`: This calls a `CallableBlock`, accepts the following arguments:
  - _name_: the name of the variable pointing to a callable block.
  -  _arguments_ (0 or more): will be resolved and passed along to the block. The number of arguments must match the number of parameters.

When a block is called, it pushes a context (same as the `for` and inheritance tags). This means that variables defined inside this context won't be available outside of it.


djbe:
Obviously, these new tags will need to be documented (see #232).

CHANGELOG.md
```diff
@@ -19,6 +19,9 @@ _None_
 * Images: new dot-syntax template, use `dot-syntax-swift3` or `dot-syntax` (for Swift 2.3).  
   [David Jennes](https://github.com/djbe)
   [#206](https://github.com/AliSoftware/SwiftGen/pull/206)
+* Stencil: added two new tags `define` and `call`, see the documentation for in depth explanations on how to use them. Reworked the "dot-syntax" and "structured" templates to use them, which removes the limitation of 5-level deep structures.  
+  [David Jennes](https://github.com/djbe)
+  [#232](https://github.com/AliSoftware/SwiftGen/pull/232)
```
AliSoftware: Wondering if it wouldn't be more readable to split that CHANGELOG entry in two, one in "Internal Changes" about the new tags define & call (well that's not totally internal, as it opens those tags for people doing custom templates, so maybe not such a good idea, but …), and another to tell that the 5-level limitation has been lifted?

Not really sure, just opening the debate.

GenumKit/Stencil/CallNode.swift
```diff
@@ -0,0 +1,92 @@
...
+
+    for (parameter, argument) in zip(parameters, arguments) {
+      result[parameter] = try argument.resolve(context)
+    }
+
+    return result
+  }
+}
+
+open class DefineNode: NodeType {
```
AliSoftware: Is it a convention in Stencil as well to make everything `open` like that (class, variables and the `parse` function) even for "leaf" (non-abstract) nodes?

CHANGELOG.md
djbe: Sure thing

GenumKit/Stencil/CallNode.swift
```diff
@@ -0,0 +1,92 @@
...
+
+  public init(variableName: String, parameters: [String], nodes: [NodeType]) {
+    self.variableName = variableName
+    self.parameters = parameters
+    self.nodes = nodes
+  }
+
+  open func render(_ context: Context) throws -> String {
+    let result = CallableBlock(parameters: parameters, nodes: nodes)
+    context[variableName] = result
```
AliSoftware: What does that mean if that `CallableBlock` isn't used using a `CallNode` later?

Like if I do this:
```
{% define printx %}Value of x is {{x}}.{% enddefine %}`
{% define add x y %}Adding {{x}} and {{y}} results in {{x + y}}.{% enddefine %}
```
But then instead of calling `{% call printx %}` or `{% call add 2 3 %}` I do `{{printx}}` or `{{add}}`?

What should we do in this case?

* For a `CallableBlock` which don't take arguments, making it `CustomStringConvertible` and rendering them in place could be a nice solution? Cause for now I'm not sure what this will render, but it will probably work but render something not pretty, like `<CallableBlock 0x12345678>` or something?
* For a `CallableBlock` which takes parameters, the `TemplateSyntaxError` thrown by the non-matching arity would probably be enough, making the rendering fail properly

djbe: Hmmm, good idea. Will have to test this.

GenumKit/Stencil/CallNode.swift
djbe: I just started with a `SetNode` copy 😆

AliSoftware:
> Obviously, these new tags will need to be documented (see #232).

I don't think the documentation for these tags will take place in #232 . I mean, #232 is more to document each templates (their name, what the generated code looks like, …), while the documentation for the Template Syntax itself already exists in [Templates.md](https://github.com/AliSoftware/SwiftGen/blob/master/documentation/Templates.md#templates-format-nodes-and-filters). Sure, when we tackle #232 this `Templates.md` will probably also be cleaned up and rewritten more nicely, but I think until then we should put the documentation of those new tags in the existing `Templates.md` documentation anyway so that we don't forget them, and they will be moved with the rest in case #232 refactor this documentation file in the process of documenting each template as well.

djbe:
@AliSoftware I was just doing that when you wrote that comment 😆

CHANGELOG.md
```diff
@@ -19,12 +19,18 @@ _None_
 * Images: new dot-syntax template, use `dot-syntax-swift3` or `dot-syntax` (for Swift 2.3).  
   [David Jennes](https://github.com/djbe)
   [#206](https://github.com/AliSoftware/SwiftGen/pull/206)
+* Reworked the "dot-syntax" and "structured" templates to use the new `define` and `call` tags, which greatly simplifies the templates, and also removes the limitation of 5-level deep structures.
```
AliSoftware: &lt;nitpicking mode ON&gt;
I know the 80-columns-limit thing is a bit outdated, but still, don't hesitate to line-wrap the CHANGELOG entries in the raw markdown (the CRLF will be ignored once rendered unless the previous line ends with two spaces, so…)
&lt;nitpicking mode OFF&gt;

GenumKit/Stencil/CallNode.swift
AliSoftware: 😇

templates/images-dot-syntax-swift3.stencil
```diff
@@ -32,67 +32,22 @@ struct {{enumName}}Type: ExpressibleByStringLiteral {
     self.init(stringLiteral: value)
   }
 }
+{% define recursiveBlock images tb %}
```
AliSoftware: `tb` is for "tabs"? Maybe use `sp` for space or `indent` instead? :wink:

djbe: Yeah, is from my other PR. indent was too long, so I'd go `sp`.

GenumKit/Stencil/CallNode.swift
AliSoftware: Do you think we should keep it that way or would it be better to fail for `enumBlock` and render the `enumBlock2`?

AliSoftware: Yeah, let's make it throw if it's not used via `{% call %}` then


🎶 Let it ~snow~ throw, let it throw, lt it throw… 🎶
https://www.youtube.com/watch?v=mN7LW0Y00kE

djbe:
Any more input on this? Otherwise I think it's ready to be merged.

AliSoftware:
I was actually currently reviewing that (but I'm on the train so connectivity isn't good…) but feels ok to me at a glance so 👍

djbe:
Take your time, I'd rather get it right ;)

GenumKit/Stencil/CallNode.swift
```diff
@@ -0,0 +1,96 @@
...
+
+    for (parameter, argument) in zip(parameters, arguments) {
+      result[parameter] = try argument.resolve(context)
+    }
+
+    return result
+  }
+
+  public var description: String {
+    fatalError("`func` blocks must be called using the `call` tag")
```
AliSoftware: We should find a way to throw properly instead of crashing with `fatalError` (maybe that can be done by implementing `render` rather than `description` or something?)

templates/strings-structured.stencil
```diff
@@ -10,218 +10,55 @@ import Foundation
...
+{{args.sp}}  {% endif %}
+{{args.sp}}  {% endfor %}
+{{args.sp}}  {% if args.strings.subenums %}
+{{args.sp}}  {% for subenum in args.strings.subenums %}
+{{args.sp}}  {% set subkey %}levelSub{{keySuffix}}Key{{subenum.name|swiftIdentifier|titlecase|snakeToCamelCase|escapeReservedKeywords}}{% endset %}
+{{args.sp}}  case .{{subenum.name|swiftIdentifier|titlecase|snakeToCamelCase|escapeReservedKeywords}}(let {{subkey}}):
+
+{{args.sp}}    switch {{subkey}} {
+{{args.sp}}    {% set sp %}{{args.sp}}    {% endset %}
+{{args.sp}}    {% set keySuffix %}{{keySuffix}}Sub{% endset %}
```
AliSoftware: We might want to change that to `%{ set keySuffix %}{{keySuffix + 1}}{% endset %}` but iirc that requires a change in Stencil so that the `+` operator tries to _convert_ its parameters to `Number` (instead of failing if they are of type `String` without trying to interpret them as numbers)?

AliSoftware: A workaround in the meantime is to use a filter that returns a number, like use `{{args.sp|length / 4}}` (or maybe a dedicated variable instead of reusing `args.sp` to avoid confusion is someone customises the template and change the indentation, so recurse with `{% set keySuffix %}{{keySuffix}} {% endset %}` then use `{{keySuffixSpaces|length}}`) to compose the subenum name with a numeric value instead of `SubSubSubSub…`?

djbe: Now that sounds really hacky... I'd rather do the extra work of some filter/operator.

TBH no one is going to see these long keys in autocomplete, they're scoped within the switch call. You'll only see it when looking at the generated code for the `structured` template, which we're fasing out anyway.

AliSoftware: Fair.

documentation/Templates.md
```diff
@@ -67,6 +67,8 @@ Additionally to the [tags and filters](https://stencil.fuller.li/en/latest/built
 | Node | Description |
 | ---- | ----------- |
 | `{% set my_capture_variable %}`<br>`{% endset %}` | node pair allowing you to capture the rendering of nodes inside it, so you can reuse it later by reusing `{{ my_capture_variable }}`.<br>This is useful if you intend to reuse the same content at multiple places — like the `sceneID` is for the `storyboards-default.stencil` template for example. |
+| `{% func my_block a b c %}`<br>`{{args.a}} - {{args.b}} - {{args.c}}`<br>`{% endfunc %}` | node pair allowing you to define a reusable block of nodes inside (like a function), so you can reuse it later.<br>In contrast to the `set` tag, the capturing for `func` is lazy, it will only be processed when actually called. When processed, the block will provide a variable `args` containing all the parameter values.<br>Similar to the `for` tag, this tag provides it's own context scope when called, so variables inside of it are not available outside of it. |
```
kylef: I wonder if it might instead make sense to make use of `{% include %}` tag. You could break component into separate files and make them more reusable/composable.

Short Example:

```html+django
{% for box in boxes %}
    {% include "render_box.html" %}
{% endfor %}
```

This has the benefit of allowing users to override default reusable components without having to overriding complete templates.

Example:

```swift
FileSystemLoader(paths: ["~/Projects/some/user/templates", "/usr/local/share/swiftgen/templates"])
```

Where these paths look as follows:

```
swiftgen/templates
├── article_detail.html
├── article_list.html
├── comment_detail.html

user/templates
└── comment_detail.html
```

Where the user has only overridden `comment_detail.html` to be the following:

##### `user/templates/comment_detail.html`

```html+django
<li>{{ comment.author }} said {{ comment }}</li>
```

##### `swiftenv/templates/comment_detail.html`

```html+django
<li>
  <h5>{{ comment.author }} at {{ comment.published }}</h5>
  {{ comment|markdown }}
</li>
```

##### `article_detail.html`

```html+django
<h3>{{ article.title }}</h3>

{{ article|markdown }}

<h4>Comments</h4>

<ul>
  {% for comment in article.comments %}
    {% include "comment_detail.html" %}
  {% endfor %}
</ul>
```

Now, when `article_detail.html` is rendered, the users overridden `comment_detail.html` will be used for each comment.

AliSoftware: I think I remember back in the day when I first implemented `SetNode` that I did that in order to avoid having secondary files and have stand-alone templates instead, especially for reusable parts that are short.

I think the same might apply here for func/call in the sense that sure `include` makes sense and can be really nice to allow people to override only parts of the templates, but:

1. sometimes the parts we reuse are so small that it feels not worth it to separate them in dedicated files just for a single node
2. but more importantly SwiftGen templates offer _alternatives_, unlike templates to generate a web site like with Jekyll for example. What I mean by that is that the files we have in `/usr/local/share/swiftgen/templates` are more "1 file = one template alternative to choose from, by using the `-t` flag during invocation", unlike a HTML website where the set of template files are more various parts to construct a single web site and you'll always use the same templates every time you regenerate your website.

Not sure that makes sense?

Anyway, that could still be interesting to see if using`{% include %}` in some of our cases is worth it (for the car you point out maybe it is indeed) but people should then be careful to use template parts that goes together when they select one over the other with `-t`. And if we do that we might end up with template names like `strings-dot-syntax-swift3[case].inc.stencil` or similar to disambiguate… would be hard not to be lost in all the template alternatives otherwise…

djbe: But then you'd need a partial for each template, because there are always subtle differences, sometimes even multiple partials. There are already quite a few templates.

More importantly, how would you achieve recursion using the template tag? I quite like the clean syntax the `func`/`call` tags provide.

djbe: @kylef We're considering moving these (`func`/`call`) upstream into Stencil. Would you consider this, and if so, are there any big changes you'd want? That way we can implement those now and preserve compatibility for our users.

kylef: @djbe I'm a little concerned about adding a lot of complexity for template authors, but I do see a use for it so providing it can be clearly documented and the names and use is very clear I think we can go ahead with it.

---

Just want to add that Jinja2 (which is very similar to Stencil for Python) has similar functionality called "macros" ([docs](http://jinja.pocoo.org/docs/dev/templates/#macros)). Just incase any of the ideas here may be useful.

```html+django
{% macro input(name, value='', type='text', size=20) -%}
    <input type="{{ type }}" name="{{ name }}" value="{{
        value|e }}" size="{{ size }}">
{%- endmacro %}
```

```html+django
<p>{{ input('username') }}</p>
<p>{{ input('password', type='password') }}</p>
```

---

They also have functionality to pass a block down: ([docs](http://jinja.pocoo.org/docs/dev/templates/#call))

```html+django
{% macro render_dialog(title, class='dialog') -%}
    <div class="{{ class }}">
        <h2>{{ title }}</h2>
        <div class="contents">
            {{ caller() }}
        </div>
    </div>
{%- endmacro %}

{% call render_dialog('Hello World') %}
    This is a simple dialog rendered by using a macro and
    a call block.
{% endcall %}
```

kylef: Is there any justification to why you have to access the args via `args.` instead of directly accessing them in the context:

```html+django
{% func my_block a b c %}`<br>`{{args.a}} - {{args.b}} - {{args.c}}`<br>`{% endfunc %}
```

vs:

```html+django
{% func my_block a b c %}
  `<br>`{{ a }} - {{ b }} - {{ c }}`<br>`
{% endfunc %}
```

kylef: @djbe I think in that case, it would make sense to call the `a` something else?

djbe: Sure, I can remove it, makes the syntax (to use vars) shorter.

By the way, with this you can pass blocks around. They are actually variables in the context, so you could bass a block A to another block B (as a parameter) and within B call A.

Would prefer the Jinja2 syntax? Switching to `macro` instead of `func` is easy. The parameter/argument syntax is more difficult, currently I'm just using the functionality already in Stencil. The passing jinja blocks as parameters can be done using a combination of the `Set` tag and `call` tag.

Note: yeah, we also want to merge `set` upstream 😆

AliSoftware: I have no preference of calling the node to define `macro` instead of `func`, but I'm not a fan of the Jinja2 syntax for the parameters, also because as @djbe pointed out, we'd have to parse them ourselves while with our current syntax for `func` (1) we use the already existing methods of Stencil to parse node arguments (2) it makes the syntax more consistent with all other Stencil nodes that use spaces to separate arguments of various Nodes.

kylef: I agree with @AliSoftware, just wanted to share theirs incase we wanted to provide some of their functionality such as default values.

```html+django
{% function input name, value='', type='text', size=20 %}
```

Although, a user could use this behaviour with the `default` filter.

```html+django
{{ type|default:"text" }}
```

kylef: But, adding default values can be done at any time without breaking the syntax so I'd suggest against adding it early on.

djbe: Right, so to summarise:
- Rename `func` to `macro`
- Later maybe add default value functionality

If that's all, I'll quickly adjust this PR and we can merge it. The PR to Stencil will come later when we'll split SwiftGen up in frameworks.

CHANGELOG.md
```diff
@@ -13,12 +13,19 @@ _None_
...
+* Added a `strings-no-comments-swift3` template that does not include the
+  default translation of each key.  
   [Loïs Di Qual](https://github.com/ldiqual)
   [#222](https://github.com/AliSoftware/SwiftGen/issues/222)
-* Images: new dot-syntax template, use `dot-syntax-swift3` or `dot-syntax` (for Swift 2.3).  
+* Images: new dot-syntax template, use `dot-syntax-swift3` or `dot-syntax` (for
+  Swift 2.3).  
   [David Jennes](https://github.com/djbe)
   [#206](https://github.com/AliSoftware/SwiftGen/pull/206)
+* Reworked the "dot-syntax" and "structured" templates to use the new `func`
```
AliSoftware: `macro`

CHANGELOG.md
```diff
@@ -31,6 +38,10 @@ _None_
 * Better error handling in the `colors` command.  
   [David Jennes](https://github.com/djbe)
   [#227](https://github.com/AliSoftware/SwiftGen/pull/227)
+* Stencil: added two new tags `macro` and `call`, see the documentation for in
```
AliSoftware: Link to documentation

AliSoftware:
🎉

PR #243, opened 2016-12-15
Parse storyboard customModule property

djbe:
This deprecates the `--import` option.

As a bonus, this fixes an issue where a user could have a class name collision with classes from modules.

### DIFF
--- a/GenumKit/Parsers/StoryboardParser.swift
+++ b/GenumKit/Parsers/StoryboardParser.swift
@@ -147,7 +147,7 @@ func == (lhs: StoryboardParser.Scene, rhs: StoryboardParser.Scene) -> Bool {
 
 extension StoryboardParser.Scene: Hashable {
   var hashValue: Int {
-    return "\(storyboardID);\(tag);\(customModule);\(customClass)".hashValue
+    return storyboardID.hashValue ^ tag.hashValue ^ (customModule?.hashValue ?? 0) ^ (customClass?.hashValue ?? 0)
   }
 }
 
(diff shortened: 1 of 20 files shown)


GenumKit/Parsers/StoryboardParser.swift
```diff
@@ -109,26 +123,43 @@ public final class StoryboardParser {
...
-  return lhs.storyboardID == rhs.storyboardID && lhs.tag == rhs.tag && lhs.customClass == rhs.customClass
+  return lhs.storyboardID == rhs.storyboardID &&
+    lhs.tag == rhs.tag &&
+    lhs.customClass == rhs.customClass &&
+    lhs.customModule == rhs.customModule
 }
 
 extension StoryboardParser.Scene: Hashable {
   var hashValue: Int {
-    return "\(storyboardID);\(tag);\(customClass)".hashValue
```
AliSoftware: btw I wonder if it wouldn't be better to `[storyboardID,tag,customModule,customClass].map({ $0.hashValue }).reduce(0, ^)`?

GenumKit/Parsers/StoryboardParser.swift
```diff
@@ -109,26 +123,43 @@ public final class StoryboardParser {
...
 func == (lhs: StoryboardParser.Segue, rhs: StoryboardParser.Segue) -> Bool {
-  return lhs.segueID == rhs.segueID && lhs.customClass == rhs.customClass
+  return lhs.segueID == rhs.segueID &&
+    lhs.customClass == rhs.customClass &&
+    lhs.customModule == rhs.customModule
 }
 
 extension StoryboardParser.Segue: Hashable {
   var hashValue: Int {
-    return "\(segueID);\(customClass)".hashValue
```
AliSoftware: In case we decide to do that change, apply here as well (note that I'm not sure if `^` is a better hash or not, I've read it is, but also depends on how `String.hash` is computed by Apple, so… And I think it's cheaper than string interpolation anyway

PR #245, opened 2016-12-18
More string filter test cases

djbe:
Currently written just to pass the unit-tests, this needs a thorough check of what's expected, and then fix it. For example #233.

A good idea would be to look how other languages provide such functionality, and compare our output to theirs.

### DIFF
--- a/UnitTests/Filters/StringFiltersTests.swift
+++ b/UnitTests/Filters/StringFiltersTests.swift
@@ -29,9 +29,9 @@ class StringFiltersTests: XCTestCase {
       "a__b__c": "a__b__c",
       "__y_z!": "__y_z!",
       "PLEASESTOPSCREAMING": "pleasestopscreaming",
-      "PLEASESTOPSCREAMING!": "pleasestopscreaminG!",
-      "PLEASE_STOP_SCREAMING": "pleasE_STOP_SCREAMING",
-      "PLEASE_STOP_SCREAMING!": "pleasE_STOP_SCREAMING!"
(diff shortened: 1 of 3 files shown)


UnitTests/Filters/StringFiltersTests.swift
```diff
@@ -12,13 +12,26 @@ class StringFiltersTests: XCTestCase {
...
+      "STRing_with_words": "stRing_with_words",
+      "string_wiTH_WOrds": "string_wiTH_WOrds",
       "": "",
-      "X": "x",
       "URLChooser": "urlChooser",
+      "a__b__c": "a__b__c",
       "__y_z!": "__y_z!",
-      "PLEASESTOPSCREAMING": "pleasestopscreaming"
+      "PLEASESTOPSCREAMING": "pleasestopscreaming",
+      "PLEASESTOPSCREAMING!": "pleasestopscreaminG!",
```
AliSoftware: Why would this case upper-case the last `G`?
Same for the next two expectations.

Maybe that's how the filter works now, so the unit tests were written to match them, but that's not what we want as an output in those cases, so unit tests should dictate what we want and implementation should be written so as to match the expectations

djbe: From the comments in the code:
- If the string starts with only one uppercase letter, lowercase that first letter
- If the string starts with multiple uppercase letters, lowercase those first letters up to the one before the last uppercase one

e.g. "PeoplePicker" gives "peoplePicker" but "URLChooser" gives "urlChooser"

We should modify the filter to ignore non-letters for testing.

Issue #244, opened 2016-12-15
Remove (to be) deprecated templates and variables in next major, new defaults


djbe:
### List of to be removed template variables:
See https://github.com/SwiftGen/SwiftGenKit/issues/5

### List of to be removed/renamed templates:
See https://github.com/SwiftGen/templates/issues/11

### Other things to remove

- `--enumName`: replaced by CLI parameter, template should provide default
- `--sceneEnumName`: same as above
- `--segueEnumName`: same as above
- `--import` no longer used, replaced by `modules` context variable

AliSoftware:
We should also remove deprecated templates (or maybe you planned to create a dedicated issue for them)?

* Make Swift 3 the new default template in the next major, moving the Swift 2 templates to opt-in
* Maybe remove "structured" templates as "dot-syntax" are way better anyway ?
* Probably make "dot-syntax" the default template for any subcommand that has a dot-syntax (images, strings), and move the non-dot-syntax ones as opt-in ?

djbe:
No, let's also do that here, we'll create dedicated PR's for each sub item

AliSoftware:
> * storyboards:
>   * lowercase: no idea
This one was originally created to address #111

Not sure if it's still relevant in Swift 3, haven't looked in details about the consequences regarding that.

djbe:
Seems like a really specific use case, better just rename the scene to something non-conflicting, or use a custom template.

AliSoftware:
That is actually not so uncommon, I've encountered that as well, and it's very common to chose the name of the class as the scene identifier (use `"DetailViewController"` for the scene identifier of a scene of class `DetailViewController` for example), like most people do for `UITableViewCells` and their `reuseIdentifier` as well.

The solution we used was to duplicate the bundled template to make a custom one, adding the name of the App (= the name of the _module_ that's associated with the app target) in front of every `customClass`.

Note that if we implement the `--param foo=bar` feature in the future to pass arbitrary parameters to the command line, we could imagine a `--param appModuleName=MyAwesomeApp` that would be used when `customModule` is empty, instead of using no prefix at all, and that would solve #111 altogether. So maybe we could get rid of `storyboards-lowercase` indeed but it would then be good to implement the `--param foo=bar` CLI flags soon after to give a way to overcome this #111 issue ;)

djbe:
True. There's no other way to avoid these collisions? The enum inside an extension works here (playgrounds):
```swift
struct T {
}

extension T {
[...]

PR #247, opened 2016-12-20
Strings dot-syntax-swift3 fix

### LINKED ISSUE #241: Fix naming issues in strings dot-syntax-swift3 template
Currently using the `-t dot-syntax-swift3` option leads to the following code in usage side:
```swift
L10n.Bananas.owner(p0: 2, p1: "John")
```

Instead it should be:
```swift
L10n.Bananas.owner(2, "John")
```

See https://github.com/AliSoftware/SwiftGen/pull/168#discussion_r92412861 on how this issue was introduced and where it lies. We should wait for #237 to be merged before fixing this issue though as suggested by @AliSoftware [here](https://github.com/AliSoftware/SwiftGen/pull/168#discussion_r92413697).

djbe:
Fixes #241, is also an example on how to get rid of the `p0`, `p1` variables.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 ### Bug Fixes
 
 * Strings: fix issue with `dot-syntax-swift3` where function definitions were
-  not swift3 guidelines compliant.  
+  not Swift 3 guidelines compliant.  
   [David Jennes](https://github.com/djbe)
   [#248](https://github.com/AliSoftware/SwiftGen/pull/248)
 
@@ -94,7 +94,7 @@
 * Fix swift 3 storyboard templates to be compliant with swift 3 api design guidelines.  
   [Afonso](https://github.com/afonsograca)
   [#194](https://github.com/AliSoftware/SwiftGen/pull/194)
-* Remove the `key` param label from the `tr` function for Localized String in the swift3 template.  
+* Remove the `key` param label from the `tr` function for Localized String in the Swift 3 template.  
   [AndrewSB](https://github.com/AndrewSB)
   [#190](https://github.com/AliSoftware/SwiftGen/pull/190)
 * The `swiftgen images` command now uses the `actool` utility to parse asset catalogs,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,10 @@
 
 ### Bug Fixes
 
-_None_
+* Strings: fix issue with `dot-syntax-swift3` where function definitions were
+  not swift3 guidelines compliant.  
+  [David Jennes](https://github.com/djbe)
+  [#248](https://github.com/AliSoftware/SwiftGen/pull/248)
 
 ### New Features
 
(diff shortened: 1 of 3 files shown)


templates/strings-dot-syntax-swift3.stencil
```diff
@@ -10,12 +10,14 @@ import Foundation
 // swiftlint:disable nesting
 // swiftlint:disable variable_name
 // swiftlint:disable valid_docs
+{% macro parametersBlock params %}{% for type in params.types %}_ p{{forloop.counter}}: {{type}}{% if not forloop.last %}, {% endif %}{% endfor %}{% endmacro %}
+{% macro argumentsBlock params %}{% for type in params.types %}p{{forloop.counter}}{% if not forloop.last %}, {% endif %}{% endfor %}{% endmacro %}
```
AliSoftware: Why not use Stencil's `|join(", ")` filter here instead of a `for` loop?
Would avoid the use of `if not forloop.last`…

templates/strings-dot-syntax-swift3.stencil
```diff
@@ -10,12 +10,14 @@ import Foundation
 // swiftlint:disable nesting
 // swiftlint:disable variable_name
 // swiftlint:disable valid_docs
+{% macro parametersBlock params %}{% for type in params.types %}_ p{{forloop.counter}}: {{type}}{% if not forloop.last %}, {% endif %}{% endfor %}{% endmacro %}
```
AliSoftware: Wondering what it would like (i.e. would it be more readable or less readable?) if you used a `MapNode` here instead?

templates/strings-dot-syntax-swift3.stencil
djbe: Because I don't have an array at that point? I'm trying to __not__ use the names/namedtypes variables.

templates/strings-dot-syntax-swift3.stencil
AliSoftware: Ah, right. Might be worth nitpicking it in a separate PR btw (that we could merge either before or after the Great Split and the next major version, depending on our agenda)

templates/strings-dot-syntax-swift3.stencil
AliSoftware: Ah, right, sorry I talked too fast, was thinking about `params.types|join(", ")` but that's not what you want, you don't want `Int, Float, String` but `p0, p1, p2` here, right. nvm

AliSoftware:
I've edited #244 to add `names` and `typednames` to your list btw

djbe:
Cool! Actually that whole `params` array of sub-items can just become the contents of `params.types`. The rest is not needed anymore.

AliSoftware:
Haha I see you've won the race and managed to validate your edit to add that new consideration… before I did validate my own edits and got rejected ^^

CHANGELOG.md
```diff
@@ -6,7 +6,10 @@
 
 ### Bug Fixes
 
-_None_
+* Strings: fix issue with `dot-syntax-swift3` where function definitions were
+  not swift3 guidelines compliant.
```
AliSoftware: `s/swift3/Swift 3/` ?

Issue #240, opened 2016-12-13
[Internal] Reorganize Frameworks to extract the Stencil nodes ("The Great Split")


AliSoftware:
There is multiple aims in this issue we want to discuss:

## Organization

We'll be reorganising the `GenumKit.framework` to split it into:
* `StencilSwiftKit.framework`, containing only the Stencil nodes specific to SwiftGen/Sourcery and don't have their place in Stencil proper
* `SwiftGenFramework.framework` for the rest of the code currently in GenumKit responsible for parsing the various input files and generating the contexts accordingly

We'll probably move each of those two frameworks in their own repos and define dependencies properly, so that both `SwiftGen` and `Sourcery` can depend on `StencilSwiftKit` and that each of them have their own `podspec` and versionning life cycle.

## Cleaning up the Stencil Nodes & Filters in GenumKit

* [ ] What are the ones that should now **be integrated into @kylef 's Stencil** proper directly?
* [ ] What are the ones that are **no longer needed** since Stencil evolved (I'm looking at you, `JoinNode`)
* [ ] What are the ones that should be **moved into a future `StencilSwiftKit.framework`**, so that wet can share them with @krzysztofzablocki 's [Sourcery](https://github.com/krzysztofzablocki/Sourcery)?

See associated GitHub Project here: ["The Great Split"](https://github.com/AliSoftware/SwiftGen/projects/1)

krzysztofzablocki:
What should we move into StencilSwiftKit:
- all filters, except `escapeReservedKeywords`
- SetNode

AliSoftware:
You mean move to **Stencil** itself, not to **StencilSwiftKit**… right?

AliSoftware:
Also, for the same reasons as for `escapeReservedKeywords`, I don't think the `swiftIdentifier` filter should be part of Stencil proper

djbe:
I suggest we (mis)use a project board for this:
[The Great Split](https://github.com/AliSoftware/SwiftGen/projects/1) 😆 

First add blocks for each item to the first column, except those related to pure command line stuff. Then we can move one by one each to the correct column, discuss if needed and also the impact of moving it, etc...

AliSoftware:
:+1: great idea

AliSoftware:
[...]

Issue #253, opened 2017-01-03
Allow to inject custom parameters to the Stencil context via command line


AliSoftware:
One thing that would add a lot of flexibility in SwiftGen is to be able to inject any custom parameters to the template from the command line:

```
swiftgen strings Localizable.strings --param MyParam Value1 --param OtherParam Value2
```

This would allow us to generalise the special case of `--enumName X`: the `--enumName` parameter would disappear in profit of the new `--param enumName X` way of injecting it.

For example one could imagine using `--param AppModule MyAwesomeApp` to inject a custom value for the `AppModule` variable in the Stencil context. Then in the storyboard templates, we'd be able to use `{% if AppModule %}{{AppModule}}.{% endif %}` to prefix all class names with the app's module name (to allow disambiguation) if its value has been provided via the command line.

krzysztofzablocki:
`arguments.AppModule` for consistency with Sourcery and avoiding name clash

AliSoftware:
@krzysztofzablocki Does Sourcery also named its CLI flag `--param`?

If so, why not use `param.AppModule` in the Stencil context for consistency between the CLI flag and the context?

AliSoftware:
Also, do you do KeyPath parsing?

Like if the user invokes Sourcery with `--param foo.bar 1 --param foo.baz 2`, does the dictionary created under the `"arguments"` entry has 2 entries named `"foo.bar"` and `"foo.baz"`, or do you create an entry `"foo"` which will itself be a dictionary with `"bar"` and `"baz"`?

(I don't think that last solution is worth implementing, but just asking in case you do that, for consistency's sake too)

krzysztofzablocki:
Right now our version is very simple, we'll need to figure out a way for parsing those extra args properly with commander, related discussion https://github.com/krzysztofzablocki/Sourcery/pull/80

Whatever we do here best we do it in both tools ;) I will prefer you mentioned for extra arguments but I also think we should scope them under a single context item

AliSoftware:
Yeah I totally agree with your suggestion in scoping them in a single item in the context, hence my question about the name of such key (`arguments` vs `param` vs `params`) 😉

djbe:
Why scope them though? It just ends up in more verbose templates. Could be useful if a user wants to override variables provided by the parser?

krzysztofzablocki:
well, that is exactly why I want to scope it to avoid potentially overriding one of the things we will add via parser, what use case can you see there? 
[...]

PR #255, opened 2017-01-09
Strings: added explicit bundle parameter to NSLocalizedString calls to support frameworks

NachoSoto:
`NSLocalizedString` defaults to `Bundle.main` when looking up strings.
With this change, we support the following setup:

- `Test.framework`:
    - `Localizable.strings`:
        - `"String" = "Test";`
    - Generated `Strings.swift`.
- App application target.
    - Link against `Test.framework`.
    - Reference `Test.L10n.String.string`.

Without this change, this setup would be looking for the `"String"` string in the main application bundle, which would result in it not being found.

Thanks for submittng a Pull Request to SwiftGen!
Don't forget to add an entry in the `CHANGELOG.md` file to explain your changes and credit yourself!

--- 

## Important: ⚠️ Refactoring in progress ⚠️ 

SwiftGen has arrived to a point where it's big enough to deserve [its own GitHub organization](https://github.com/SwiftGen/), and will also be split into multiple repositories for better structure and flexibility.

This means that SwiftGen's code will be subject to a big reorganizaton (see [#240](https://github.com/AliSoftware/SwiftGen/issues/240)) and that it may not be the best time to submit a new PR, as it will likely not be valid anymore after that refactoring.

We still welcome new PRs very much of course, and are really glad you wish to contribute! But if your change is likely to be affected by this big reorganization, it might be easier for you to wait a few weeks until that refactoring is finished (It's not expected to last long — we hope to finish that big migration ASAP) before submitting your changes, to avoid doing the work twice.

---

### DIFF
--- a/UnitTests/expected/Strings-Entries-Default.swift.out
+++ b/UnitTests/expected/Strings-Entries-Default.swift.out
@@ -27,7 +27,7 @@ extension L10n: CustomStringConvertible {
   }
 
   private static func tr(key: String, _ args: CVarArgType...) -> String {
-    let format = NSLocalizedString(key, comment: "")
+    let format = NSLocalizedString(key, bundle: Bundle(for: BundleToken.self), comment: "")
     return String(format: format, locale: NSLocale.currentLocale(), arguments: args)
   }
 }
@@ -35,3 +35,5 @@ extension L10n: CustomStringConvertible {
 func tr(key: L10n) -> String {
   return key.string
 }
+
+private final class BundleToken {}
--- a/UnitTests/expected/Strings-File-CustomName.swift.out
+++ b/UnitTests/expected/Strings-File-CustomName.swift.out
@@ -63,7 +63,7 @@ extension XCTLoc: CustomStringConvertible {
   }
 
   private static func tr(key: String, _ args: CVarArgType...) -> String {
-    let format = NSLocalizedString(key, comment: "")
+    let format = NSLocalizedString(key, bundle: Bundle(for: BundleToken.self), comment: "")
     return String(format: format, locale: NSLocale.currentLocale(), arguments: args)
   }
 }
@@ -71,3 +71,5 @@ extension XCTLoc: CustomStringConvertible {
 func tr(key: XCTLoc) -> String {
   return key.string
 }
+
+private final class BundleToken {}
--- a/UnitTests/expected/Strings-File-Default.swift.out
+++ b/UnitTests/expected/Strings-File-Default.swift.out
@@ -63,7 +63,7 @@ extension L10n: CustomStringConvertible {
   }
 
   private static func tr(key: String, _ args: CVarArgType...) -> String {
-    let format = NSLocalizedString(key, comment: "")
+    let format = NSLocalizedString(key, bundle: Bundle(for: BundleToken.self), comment: "")
     return String(format: format, locale: NSLocale.currentLocale(), arguments: args)
   }
 }
@@ -71,3 +71,5 @@ extension L10n: CustomStringConvertible {
 func tr(key: L10n) -> String {
   return key.string
 }
+
+private final class BundleToken {}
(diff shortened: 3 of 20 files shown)


djbe:
We appreciate the PR, and that BundleToken trick is really neat! But as you've seen, we're currently migrating a lot of code to different repositories, and won't be able to merge your PR after these code changes. I suggest you wait until after the migrations are done and create a new PR.

Quick question: is there a reason the `BundleToken` is an `NSObject`? And can we scope it inside the strings enum, so that we can avoid clashes if we add this to the other templates (images, storyboards, ....).

AliSoftware:
Thanks for submitting this PR!

I like the idea, but as — like the PR template currently says — we're in the middle of a refactoring, we might wait for the refactoring to be finished. This PR will likely have to be applied to github.com/SwiftGen/templates once the splitting of SwiftGen in multiple repo in its own org is complete (hopefully soon)

NachoSoto:
> I like the idea, but as — like the PR template currently says — we're in the middle of a refactoring, we might wait for the refactoring to be finished. This PR will likely have to be applied to github.com/SwiftGen/templates once the splitting of SwiftGen in multiple repo in its own org is complete (hopefully soon)
>We appreciate the PR, and that BundleToken trick is really neat! But as you've seen, we're currently migrating a lot of code to different repositories, and won't be able to merge your PR after these code changes. I suggest you wait until after the migrations are done and create a new PR.

I'm happy to do that whenever the split is finished 👍 

> Quick question: is there a reason the BundleToken is an NSObject

Oops, nope! I'll change that.

>  And can we scope it inside the strings enum, so that we can avoid clashes if we add this to the other templates (images, storyboards, ....).

The type is `private` so there can't be any conflict.

AliSoftware:
Ha, I see @djbe and I replied in sync 😄

I'm not sure if it's worth scoping in the enum. It's declare`private` so it can't possibly conflict.

AliSoftware:
Damn, I must reply quicker, will wait to be out of my train with better connectivity next time 😆

NachoSoto:
I've addressed the comments and tests pass now. But I understand if it can't be merged yet :)

djbe:
How would you solve the issue of (CocoaPods) resource bundles?

For example, I have a pod with the following spec:
```
Pod::Spec.new do |s|
	s.name = 'TestPod'
 	s.resource_bundle = { 'Resources' => [
		...
	] }
end
```

With the `BundleToken` trick, you'll correctly find the `TestPod.framework` bundle, but your resources will actually be inside the `Resources.bundle` (which is inside the `TestPod.framework` bundle).

AliSoftware:
@djbe For those cases I'm not sure we can automatically detect this…
But we could make the templates customisable using a `--param ImageBundle` once #253 is implemented

NachoSoto:
Yeah that makes sense, then the template could load it by bundle identifier.

NachoSoto:
Any idea how long until we can have this? It's pretty tedious to have to duplicate all strings in the app bundle...

AliSoftware:
@NachoSoto should be any time now. I only have a couple of Hrs to review and I can do the release. I though I would have time for that last week-end but have been more busy than I though, so I hope to find some time this week for that (maybe even tomorrow or Wednesday at work).

After finishing that Great Split and doing the release, we can merge your PR very quickly. We'll just have to report it in the new https://github.com/SwiftGen/templates repository now that templates have moved there.

---

In the meantime, nothing prevents you to use your custom template, even without waiting for the new release (that's the beauty of SwiftGen, because it's based on templates, you can customise everything you need, and just use `swiftgen strings -p <path/to/your/custom/template.stencil> …` (instead of `swiftgen strings -t name-of-bundled-template …`) to use your template right away without having to wait for a new release.

djbe:
@alisoftware: I already made a new PR on templates that incorporates this and applies it to the other templates/commands.

AliSoftware:
Ace! 👌
Will take a look tomorrow!

djbe:
I'm closing this in favor of the PR on the `templates` repository: https://github.com/SwiftGen/templates/pull/17

PR #265, opened 2017-02-15
Great split

### LINKED ISSUE #240: [Internal] Reorganize Frameworks to extract the Stencil nodes ("The Great Split")

djbe:
Closes #240


djbe:
Almost done, need to:
- [ ] update enrich call to match changes in https://github.com/SwiftGen/StencilSwiftKit/pull/19/files
- [ ] fix versions for kits to 1.0.0

SwiftGen/main.swift
```diff
@@ -45,6 +46,6 @@ let main = Group {
   $0.addCommand("fonts", "generate code for your UIFonts", fontsCommand)
 }
 
-let version = Bundle(for: GenumKit.GenumTemplate.self)
+let version = Bundle.main
   .infoDictionary?["CFBundleShortVersionString"] as? String ?? "0.0"
 main.run("SwiftGen v\(version)")
```
AliSoftware: Could be interesting to fetch versions for SwiftGenKit and StencilSwiftKit too so we list them all there?

djbe: Do we also list the stencil version?

AliSoftware: Good point

NachoSoto:
Awesome 🎉 looking forward to having this in Homebrew.

AliSoftware:
🎉 You can follow the status of https://github.com/Homebrew/homebrew-core/pull/10132 to check when 4.2 will be accepted in homebrew (should not take long, they're usually quite reactive once the CI has built the bottles)

PR #269, opened 2017-02-20
Configure CircleCI

djbe:
Keep track of https://github.com/SwiftGen/StencilSwiftKit/pull/20 for any changes that need to be copied here.


djbe:
Todo: podspec linting

circle.yml
```diff
@@ -0,0 +1,21 @@
...
+
+checkout:
+  post:
+    - git submodule update --init --recursive
+
+dependencies:
+  pre:
+    - bundle install
+    - gem install xcpretty --no-rdoc --no-ri --no-document --quiet
+    - Scripts/install_swiftlint.sh
```
AliSoftware: Shouldn't that be a rake task? Or is that in the other PR splitting tasks in rakelib? (If so how do we join-merge those?)

Rakefile
```diff
@@ -150,144 +171,15 @@ namespace :playground do
...
 
   desc "Regenerate all the Playground resources based on the test fixtures.\nThis compiles the needed fixtures and place them in SwiftGen.playground/Resources"
   task :resources => %w(clean images storyboard strings)
 end
-
-
-
-## [ Release a new version ] ##################################################
-
-namespace :release do
```
AliSoftware: Haven't checked yet but I guess you moved those in Eve?

circle.yml
djbe: This PR is a bit behind on the PRs in the other repos. Will update it here when the rest is done.

Rakefile
djbe: Yeah

AliSoftware: I see the `feature/rake` branch in `Eve` but no PR, you don't consider it ready yet?

djbe: Not really, but I'll open the pr for discussion purposes.

.gitignore
```diff
@@ -1,11 +1,73 @@
+# Xcode
+#
+# gitignore contributors: remember to update Global/Xcode.gitignore, Objective-C.gitignore & Swift.gitignore
```
AliSoftware: _Side-note: we should also share our .gitignore across repos later (same idea as the rest, a gitignore in Eve as the source of truth being sync'd using a rake task that we'd run from time to time)_

rakelib/lint.rake
```diff
@@ -0,0 +1,22 @@
+namespace :lint do
+  desc 'Install swiftlint'
+  task :install do |task|
+    swiftlint = `which swiftlint`
```
AliSoftware: I'd find it both more logical and less dangerous to do both the `which` and the `$?.success?` tests in one line.
Because:

* the purpose of `which swiftlint` isn't really to know the path where `swiftlint` is installed but rather if it's installed at all (we could use `swiftlint --version` for that test instead)
* If we happen to modify this script later and add another command execution between this `which swiftlint` and the test wether the last shell command was and error or success, then that test (`$?.success?`) would check the wrong command

I suggest we write it this way instead:

```sh
swiftlint_installed = `swiftlint --version` && $?.success?
if !swiftlint_installed
  …
```

(I'm open to options if you know a better way of testing from ruby if a command line tool exists an is installe btw)

rakelib/utils.rake
```diff
@@ -0,0 +1,58 @@
+# run a command, pipe output through 'xcpretty' and store the output in CI artifacts
+def xcpretty(cmd, task, subtask = '')
+  name = (task.name + (subtask.empty? ? '' : "_#{subtask}")).gsub(/[:-]/, "_")
+  command = [*cmd].join(' && ')
+  xcpretty = `which xcpretty`
+
+  if ENV['CI']
+    sh "set -o pipefail && (#{command}) | tee \"#{ENV['CIRCLE_ARTIFACTS']}/#{name}_raw.log\" | xcpretty --color --report junit --output \"#{ENV['CIRCLE_TEST_REPORTS']}/xcode/#{name}.xml\""
+  elsif xcpretty && $?.success?
```
AliSoftware: Same remark as above, if we happen to later insert a shell command on line 6 in the future, we might miss the point that `$?.success?` wont test the appropriate shell error code and break things. I'd prefer that th `$?.success?` test to be as close as possible as the shell command return code it tests.

rakelib/utils.rake
```diff
@@ -0,0 +1,58 @@
...
+    sh "set -o pipefail && (#{command}) | xcpretty -c"
+  else
+    sh command
+  end
+end
+
+# run a command and store the output in CI artifacts
+def plain(cmd, task, subtask = '')
```
AliSoftware: tbh I'm not a fan of that `plain` method name.

What about `run` or `run_cmd` instead?
Another alternative (which might actually be nicer now that I think about it, at least probably worth checking how it looks) would be to wrap both the `xcpretty` and `plain` methods in a single `run_cmd` function which would take an `xcpretty: Bool` argument

Ruby 2 accepts named arguments, so we could take advantage of this to write
```
def run_cmd(cmd, task, xcpretty: false)
…
end
```
and call it using something like `run_cmd('the command to run', task, xcpretty: true)` when we want to filter using xcpretty

AliSoftware: We could even imagine doing the same for `xcrun`, turning it as a parameter like `xcrun: false` by default and using `xcrun: true` at call site would allow to inject the version select argument + `xcrun`

It would make the call site more clear as to when we want unfiltered output, when we want xcrun/version select, and when we want xcpretty, all under the same function.

Thoughts?

rakelib/lint.rake
djbe: Weird, I actually changed this because it kept failing for me on undefined something 'success'

AliSoftware: Maybe you forgot the question mark at the end of the method name? (Yes in ruby the question mark at the end of a method name is part of the method name and making methods returning a boolean with a question mark at the end is a convention in ruby)

rakelib/utils.rake
djbe: Even better would be to auto detect this similarly to how I detect the need for xcpretty. If possible I'd like to keep the other commands but mark them as private or something? Better to have shorter/simpler functions instead of one big one prone to breaking

rakelib/lint.rake
djbe: No it was already failing before I even touched the rake file. I have a default ruby of 2.4 though, not 2.0 (or whatever the default is on macOS now)

rakelib/utils.rake
AliSoftware: Yes to keeping the other methods separate and make them private and just make the main method private. 

But I'm not entirely convinced in detecting the command and choosing of we should `xcpretty` or `xcrun` or nothing based on checking if the command contains some key words like `xcodebuild`, would make too much intelligence in the utility method imho and not really respecting responsibility principle. That's why I'd prefer the call site to decide that.

rakelib/lint.rake
AliSoftware: After googling it seems `system('which swiftlint')` is the easiest way and directly returns true if successful, false otherwise

djbe:
Right, so I've tried a few times to convert the `build` command in this repo to simply use `xcodebuild` with the `swiftgen` target, but didn't get it to work completely. Tried it with a separate cli target, but keep getting the issue that the swift libs are linked twice in the produced binary. 

Is this there a reason we're currently building it using manual `swiftc` calls, and if not, anyone feeling up to it tackle this? We can always leave it for a separate PR, but I think we've then reached a limit on how far we can unify the rake libs across repositories.

AliSoftware:
I think we initially used `swiftc` to compile the files… because it was before #204 and Cmd-B / Cmd-R in Xcode were not working due to the structure of the project being a CLI depending on frameworks.

Now that it's an app bundle (and extraction of CLI and frameworks separately is done during release only), we can probably migrate to `xcodebuild` now.

Sources/main.swift
```diff
@@ -12,7 +12,7 @@ import SwiftGenKit
 
 // MARK: Common
 
-let templatesRelativePath = (Bundle.main.resourcePath ?? ".") + "/templates"
+let templatesRelativePath = "../templates"
```
AliSoftware: Why this change? Doesn't this not have to work both in Xcode (who building a .app bundle) and in CLI (once split and the templates/resources are next to the binary outside an app bundle)?

AliSoftware:
Just pushed 2 commits that allow `rake install` to work. It simply compiles the workspace with `xcodebuild` the same way that a Cmd-B would do in Xcode (that's the `:build` task), then rearrange the built products (binary, frameworks, etc) properly and fix the `@rpath` (that's the `:install` task).

Be careful, with this change there's still one gotcha: I removed your trick to embed the Info.plist file into the executable, so `build/swiftgen/bin/swiftgen --version` prints a version `0.0` for now. We'll have to fix that in another commit. But I think we can revert the commit you did which hard-coded the values in the `Info.plist` and let the variables here instead… and embed the `Info.plist` file _from the inside .app built product_ (which is a compiled version once variables have been replaced by Xcode) after the fact?

AliSoftware:
Ok the Info.plist trick should be fixed with my latest commits now 👍

Sources/main.swift
djbe: Was comitted by mistake after running install a couple of times.

AliSoftware: Maybe at some point we should extract that constant into its own file, then make the Rakefile reset just that file at the end of the build? 
I don't want to `git checkout main.swift` because if we made local changes to main and forgot to stage and commit them before testing our changes with `rake install` locally we'd lose all our changes. But if there constant is isolated in a one-line file it should be ok.

djbe: I'm already making a fix that uses flags added by the rake build command, we can then simply `#if` from swift code.

AliSoftware:
Note for the future PR in Eve performing the release: we must ensure not to forget to update the `SWIFTGEN_VERSION` build setting so that the info plist get the right version string updated for every new release. 

Note that I created this user-defined build setting because I thought it would allow us to control it more easily but now that I think about it harboring it in the Info.plist and using `/usr/libexec/PlistBuddy` or `plutil` in Eve's Rakefile to update it on release might be easier so we might need to adapt that.

djbe:
True, we want to make sure we're syncing the version number between info.plist and the podspec.

SwiftGen.xcodeproj/project.pbxproj
```diff
@@ -556,7 +556,7 @@
 					"$(inherited)",
 					"$(SWIFTGEN_OTHER_LDFLAGS)",
 				);
-				OTHER_SWIFT_FLAGS = "$(inherited) $(SWIFTGEN_OTHER_SWIFT_FLAGS)";
+				OTHER_SWIFT_FLAGS = "$(inherited)";
```
AliSoftware: Probably better to completely remove the setting if it's gonna only inherit and add nothing (select the setting at the target level and hit delete key)

SwiftGen.xcodeproj/project.pbxproj
```diff
@@ -580,7 +580,7 @@
...
```
AliSoftware: ditto

djbe:
woops, forgot to change that 👍

AliSoftware:
I made Circle install swiftgen in explicitly-specified location, so that we are sure that we didn't accidentally hard-code paths (like `../templates`)  somewhere and that providing custom install locations still works even if we don't test it often ourselves.

I wonder if it would make sense to store the generated binary + fmks + templates in `CIRCLE_ARTIFACTS` though? Could make sense (they are generated files / artefacts after all) but could also take a lot of space (as they are kept accross builds) for nothing so not sure it's worth it?

djbe:
I don't think that's really worth it. You can always re-run a build with ssh access to fetch the data if we need something, or even configure fastlane to upload builds to github.

djbe:
I've modified https://github.com/SwiftGen/Eve/pull/6 to compare podspec and info.plist version numbers.

PR #271, opened 2017-02-20
Parse fonts.xml to enum like colors.xml

AFcgi:
I extend the font parser to convert fonts.xml file to 'FontName' enum.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "Resources"]
 	path = Resources
-	url = https://github.com/SwiftGen/templates.git
+url=https://github.com/AFcgi/templates.git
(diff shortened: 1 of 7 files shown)


djbe:
For reference purposes, this goes with https://github.com/SwiftGen/SwiftGenKit/pull/11 and https://github.com/SwiftGen/templates/pull/26

djbe:
I'll repeat my question from the other PR:

> Before I start a full review, could you quickly explain what the use case is here, and what it adds that's missing from the current fonts parser?

AFcgi:
Hey @djbe,
I will try to explain the use case: I think the idea and solution with the color class is awesome. So I tried to find a comparable solution for a pre definition fonts.xml structure. In this use case the fonts parser generates a type safe enum for fonts, their styles and their sizes (similar like the colors).

The fonts.xml example:
```
<?xml version="1.0" encoding="utf-8"?>
<resources>
    <font name="A01">
        <family>FranklinGothicFS</family>
        <style>Book</style>
        <size>58</size>
    </font>
</resources>
```

djbe:
Right, but the difference with fonts is: we have the actual font files themselves. Afaik, there's no need for an extra xml file, as the font files provide all the needed info. 

Or am I missing something?

PR #294, opened 2017-05-01
Update README links to the colors fixture files

juli1quere:
Hey, 

Some of the README links were broken (due to the split into multiple repo I think). Here is a fix :). 

PS: I did not have updated the CHANGELOG since tiny things like this one does not need to be mentioned :)


juli1quere:
I also added a tiny format improvement :)

README.md
```diff
@@ -141,7 +141,7 @@ swiftgen images /dir/to/search/for/imageset/assets
 This will generate an `enum Asset` with one `case` per image asset in your assets catalog, so that you can use them as constants.
 
 <details>
-<summary>Example of **code generated by the default template**</summary>
+<summary>Example of code generated by the default template</summary>
```
djbe: I think this was meant as markdown (bold). Seems github doesn't mix HTML tags with markdown. Looks ok to me without the `*`, and bold might be a bit much tbh.

djbe: I think this was meant as markdown (bold). Seems github doesn't mix HTML tags with markdown. Looks ok to me without the *, and bold might be a bit much tbh.

Issue #283, opened 2017-03-21
Show error when combining `-t` and `-p` cli options


djbe:
Right now, a user can invoke swiftgen using something like:
```swiftgen -p their-custom-template.stencil -t swift3 ...```

This might confuse the user as to the purpose of either option, or a behaviour they might not expect. We should throw an error if both options are present.

Liquidsoul:
I am not sure if this is really an _easy_ task. But maybe I am missing some information.

As I understand it, we are computing the resolved template in the [`findTemplate`](https://github.com/SwiftGen/SwiftGen/blob/d79f8a34c62cfe3c6a672a7e99948980d27172f3/Sources/ArgumentsUtils.swift#L128-L145).
Its arguments comes from `Commander`'s `Option` which means we do not have the information at this point if the value was set or if we are getting the default value.
Moreover, it is impossible for us to know if the user has set the `-t` option to its default value or not.

TL;DR:
- we cannot know it the user invoked swiftgen with `-t default` or without a `-t` option
- information about the `Option`s default values must be passed down to `findTemplate` in _every_ `command` call we have.

AliSoftware:
@Liquidsoul we plan to remove the"fallback to default template" behavior in 5.0. Both because it means changing the default when a new Swift version is released — but it's not reasonable to change the default without notifying the user and last old commands relying on one default template suddenly change the template in the next version because we decided to change the default — and because avoiding the fallback will force people to specify explicitly which template they want and make them realize they have three choice and that this template mechanism exists (we had too many issues in the past from people claiming SwiftGen didn't support Swift 3 because they didn't realize you could choose alternate templates) 

So that should solve the issue you're mentioning I guess.

AliSoftware:
Closed by #300

PR #300, opened 2017-05-08
Show error when combining `-t` and `-p` cli options #283

Liquidsoul:

---

This is a proposal to implement issue #283.
It includes disabling the default value for the `templateOption` argument which will break existing usage which does not provide any template when invoking swiftgen.

### DIFF
--- a/Sources/ArgumentsUtils.swift
+++ b/Sources/ArgumentsUtils.swift
@@ -141,15 +141,16 @@ let bundledTemplatesPath = Path(ProcessInfo.processInfo.arguments[0]).parent() +
  - returns: The Path matching the template to find
  */
 func findTemplate(prefix: String, templateShortName: String, templateFullPath: String) throws -> Path {
-  switch (templateFullPath, templateShortName) {
-  case ("", ""): throw TemplateError.noTemplateProvided
-  case (_, ""):
+  switch (!templateFullPath.isEmpty, !templateShortName.isEmpty) {
+  case (false, false):
+    throw TemplateError.noTemplateProvided
+  case (true, false):
     let fullPath = Path(templateFullPath)
     guard fullPath.isFile else {
       throw TemplateError.templatePathNotFound(path: fullPath)
     }
     return fullPath
-  case ("", _):
+  case (false, true):
     var path = appSupportTemplatesPath + "\(prefix)-\(templateShortName).stencil"
     if !path.isFile {
       path = bundledTemplatesPath + "\(prefix)-\(templateShortName).stencil"
@@ -158,7 +159,7 @@ func findTemplate(prefix: String, templateShortName: String, templateFullPath: S
       throw TemplateError.namedTemplateNotFound(name: templateShortName)
     }
     return path
-  case (_, _):
+  case (true, true):
     throw TemplateError.multipleTemplateOptions(path: templateFullPath, name: templateShortName)
   }
 }
--- a/Sources/ArgumentsUtils.swift
+++ b/Sources/ArgumentsUtils.swift
@@ -80,6 +80,8 @@ enum OutputDestination: ArgumentConvertible {
 enum TemplateError: Error, CustomStringConvertible {
   case namedTemplateNotFound(name: String)
   case templatePathNotFound(path: Path)
+  case noTemplateFound
+  case multipleTemplateOptions(path: String, name: String)
   case deprecated(option: String, replacement: String)
 
   var description: String {
@@ -89,6 +91,15 @@ enum TemplateError: Error, CustomStringConvertible {
       "or use --templatePath to specify a template by its full path."
     case .templatePathNotFound(let path):
       return "Template not found at path \(path.description)."
+    case .noTemplateFound:
+      return "A template must be chosen either via its name using the '-t' / '--template' option" +
+      " or via its path using the '-p' / '--templatePath' option.\n" +
+      "There's no 'default' template anymore: you can still access a bundled template but " +
+      "you have to explicitly specify its name using '-t'.\n" +
+      "To list all the available named templates, you can use the 'swiftgen templates list' command."
+    case let .multipleTemplateOptions(path, name):
+      return "You need to choose EITHER a named template (--template option) " +
+      "OR a template path (option --templatePath). Found name '\(name)' and path '\(path)'"
     case .deprecated(let option, let replacement):
       return "The option '--\(option)' has been deprecated. \(replacement)"
     }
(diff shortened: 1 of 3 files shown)


Sources/ArgumentsUtils.swift
```diff
@@ -88,6 +90,11 @@ enum TemplateError: Error, CustomStringConvertible {
       "or use --templatePath to specify a template by its full path."
     case .templatePathNotFound(let path):
       return "Template not found at path \(path.description)."
+    case .noTemplateFound:
+      return "A template must be chosen either using the named option --template or the path option --templatePath."
+    case let .multipleTemplateOptions(path, name):
+      return "You need to chose a named template (option --template) OR a template path (option --templatePath). " +
```
AliSoftware: s/chose/choose EITHER/

Sources/ArgumentsUtils.swift
```diff
@@ -88,6 +90,11 @@ enum TemplateError: Error, CustomStringConvertible {
       "or use --templatePath to specify a template by its full path."
     case .templatePathNotFound(let path):
       return "Template not found at path \(path.description)."
+    case .noTemplateFound:
+      return "A template must be chosen either using the named option --template or the path option --templatePath."
+    case let .multipleTemplateOptions(path, name):
+      return "You need to choose EITHER a named template (option --template) OR a template path (option --templatePath). " +
```
AliSoftware: Dammit I missed that on first review, but also: 
"--template option" not "option --template"

Sources/ArgumentsUtils.swift
```diff
@@ -126,20 +134,24 @@ let bundledTemplatesPath = Path(ProcessInfo.processInfo.arguments[0]).parent() +
  - returns: The Path matching the template to find
  */
 func findTemplate(prefix: String, templateShortName: String, templateFullPath: String) throws -> Path {
-  guard templateFullPath.isEmpty else {
+  switch (templateFullPath, templateShortName) {
+  case ("", ""): throw TemplateError.noTemplateFound
```
djbe: Maybe this case should generate a different error, telling the users that default behaviour has been deprecated, and what they should use next?

Liquidsoul: Maybe we should just change the error message [here](https://github.com/SwiftGen/SwiftGen/blob/aaa5dd3ff1f0fb166f24be5fddee00e8fe4761cd/Sources/ArgumentsUtils.swift#L93-L94) for this?

djbe: Oh, right, based on the case name I was assuming it didn't find the template file, but that's another error. Yeah, have the current error, but maybe add a description that the "default" behaviour has been removed?

Liquidsoul: What about adding this at the end of the error message?
`Default template support has been dropped but you can still access the bundled templates using --template option.`

djbe: Jup, that's what I meant 👍  As the full text would include the error you already have, I'd maybe change it to:

> A template must be chosen either using the named option --template or the path option --templatePath. Default template support has been dropped but you can still access the bundled templates using one of these options.

AliSoftware: I'm not sure the term "Default template support" is clear enough for users not familiar with SwiftGen enough. Maybe rather say "There's not anymore a template used by default, so you can still access the bundled templates but you have to explicitly specify its name"?

Btw I'm not sure the term "bundled" that we use throughout the app (in the `swiftgen templates list` command and here) is the best choice of vocabulary either. Maybe we should use "embedded" rather than "bundled" or something similar?

Sources/ArgumentsUtils.swift
```diff
@@ -89,6 +91,13 @@ enum TemplateError: Error, CustomStringConvertible {
       "or use --templatePath to specify a template by its full path."
     case .templatePathNotFound(let path):
       return "Template not found at path \(path.description)."
+    case .noTemplateFound:
+      return "A template must be chosen either using the named option --template or the path option --templatePath. " +
+        "Default template support has been dropped but you can still access the bundled templates using one of" +
```
AliSoftware: > A template must be chosen either via its name using the --template  option or via its path using the --templatePath option. 
> There's no 'default' template anymore: you can still access the bundled templates but you have to explicitly specify its name using -t.
> To list all the available named templates, you can use the 'swiftgen templates list' command.

Liquidsoul: Fine by me, I'll wait for @djbe's approval before making the change 😉

Sources/ArgumentsUtils.swift
djbe: Embedded would mean (to me) that the templates are embedded in the binary, which they aren't. I think we can keep "bundled" for now.

Sources/ArgumentsUtils.swift
djbe: Yeah, that looks better, especially the `templates list` bit.

Liquidsoul:
Wording changed (hopefully for the last time 😅)

djbe:
Wait maybe change ... Nah just kidding, LGTM.

Sources/ArgumentsUtils.swift
```diff
@@ -80,6 +80,8 @@ enum OutputDestination: ArgumentConvertible {
 enum TemplateError: Error, CustomStringConvertible {
   case namedTemplateNotFound(name: String)
   case templatePathNotFound(path: Path)
+  case noTemplateFound
```
AliSoftware: `case noTemplateProvided`?

AliSoftware: Because to me "noTemplateFound" looks like we provided a template option but couldn't find that template being asked

djbe: Oh, good point. That'd solve the first question I had about this PR.

Liquidsoul: You're right, it is a far better name 😉

Sources/ArgumentsUtils.swift
```diff
@@ -129,20 +141,24 @@ let bundledTemplatesPath = Path(ProcessInfo.processInfo.arguments[0]).parent() +
  - returns: The Path matching the template to find
  */
 func findTemplate(prefix: String, templateShortName: String, templateFullPath: String) throws -> Path {
-  guard templateFullPath.isEmpty else {
+  switch (templateFullPath, templateShortName) {
+  case ("", ""): throw TemplateError.noTemplateProvided
```
AliSoftware: Nitpicking on code style, but I'd rather do:

```
switch (!templateFullPath.isEmpty, !templateShortName.isEmpty) {
case (false, false)
  throw TemplateError.noTemplateProvided
…
```
That is:

 - Switch over `!.isEmpty` rather than comparing the strings with `""` and `_`, I think it's more performant and will make it (arguably) more readable?
 - Go to the next line after each `case …:` to be consistent, even for small case bodies.

Liquidsoul: I've made the change but I am wondering if the `!` is not making this harder to understand than going for just `isEmpty` 🤔

PR #301, opened 2017-05-08
Disable old CLI options

### LINKED ISSUE #244: Remove (to be) deprecated templates and variables in next major, new defaults

djbe:
Fixes #244.

Doesn't actually remove the options, but throw an error instead, and showing the user what they should use instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,9 @@ _None_
 
 ### Breaking Changes
 
-_None_
+* Removed deprecated CLI options. Should you still use them, SwiftGen will return an error and show the new recommended CLI option to use instead.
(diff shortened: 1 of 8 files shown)


djbe:
It's still a breaking change, we're just not getting rid of the CLI options yet for our users. We're nice like that 😆

PR #302, opened 2017-05-09
Update CHANGELOG format for linked CHANGELOGs

AliSoftware:
* [ ] ~I've added an entry in the `CHANGELOG.md` file to explain my changes and credit myself~ ➡️  How meta would that be 😄 (Probably not worth mentioning in the CHANGELOG)

---

## Motivation

* The scripts to automate the releases — and copy the content of the CHANGELOG entries for the latest version into the GitHub release text — currently fails because when sending the request to the GitHub API to update the release text, it seems that 🔗  and other characters were creating problems and made the API fail
* But more importantly, with this new format the links to the other CHANGELOGs will be more visible to users, inciting them to actually go and read those CHANGELOGs (the current format was probably _too_ discreet and I think a lot of people didn't read those changes in the other modules) as they are generally _more_ important than the changes done in this repo which is only the CLI part of SwiftGen

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,10 +24,12 @@ _None_
 
 ## 4.2.1
 
-• 🔗 [SwiftGenKit 1.1.0](https://github.com/SwiftGen/SwiftGenKit/blob/1.1.0/CHANGELOG.md)
-• 🔗 [StencilSwiftKit 1.0.2](https://github.com/SwiftGen/StencilSwiftKit/blob/1.0.2/CHANGELOG.md)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -20,7 +20,9 @@ _None_
 
 ### Internal Changes
 
-_None_
+* Improved Installation instructions in the README.
```
djbe: Err, wrong PR? 😆

AliSoftware: Whoops ^^

PR #313, opened 2017-06-07
Update SwiftGen to work with new SwiftGenKit parser interface

djbe:
This also refactors the separate commands into one (since all parsers now have the same interface). 

The only annoyance is that we have to add variations for deprecated options, because we want to show nice errors to the users.


djbe:
Removed the "breaking" tag because this PR doesn't change or remove anything, unless we remove the `enumName` (and variations) options.

djbe:
Quick feedback question:
- Do we keep the old options and show errors + what to use instead, thus leading to a bit of uglier code. 
- remove the old options, don't show an error with how to migrate and thus cleaner code. We can instead just document this in the migration guide.

AliSoftware:
I'd vote for making 5.0 clean. We have a Migration Guide now, so better remove old options right now and just document the migration.

---

I have been pushing changes to https://github.com/SwiftGen/SwiftGenKit/pull/46 with a recent commit and tried to adapt this PR to match the changes, but I fail to compile this branch now and don't understand why… not sure why Xcode is unhappy there while it seems to be OK on CI…

swiftgen-parser-compileerror

AliSoftware:
OK, nvm, I found the typo (the error message was totally unrelated and misleading) 😅

Sources/ParserCommand.swift
```diff
@@ -0,0 +1,132 @@
...
+      try ParserCommand.execute(parserType: parser,
+                                output: output,
+                                templateName: templateName,
+                                templatePath: templatePath,
+                                parameters: parameters,
+                                paths: paths)
+    }
+  }
+
+  init(parser: Parser.Type,
```
djbe: As we're dropping deprecated options, we can remove these `init` variations.

Sources/ParserCommand.swift
```diff
@@ -0,0 +1,132 @@
...
+                                output: output,
+                                templateName: templateName,
+                                templatePath: templatePath,
+                                parameters: parameters,
+                                paths: paths)
+    }
+  }
+
+  // swiftlint:disable:next function_parameter_count
+  private static func execute(parserType: Parser.Type,
```
djbe: The reason I made this static, is because `ParserCommand` is a value type and can't reference itself in the init (I think, it's been a while since I wrote this). Should we just make it a reference type to make this easier? There'll be no need to pass along the `parserType`.

djbe: Another thing is the number of parameters for this method. Maybe put all that "input" data in a struct?

AliSoftware: How come would it reference itself? 
(Switched to mobile while on the bus miss some code context…)

Sources/ParserCommand.swift
AliSoftware: :+1:

Sources/ParserCommand.swift
AliSoftware: We could at least group the templateName+templatePath+params together maybe yeah

AliSoftware: Oh ok I see it now, that's the private method used by the inits. Now that we'll probably only have one init as we drop deprecations, we'll probably not have this problem anymore.

djbe: It's been a while, I'll see if I can modify the code to call the function directly.

AliSoftware: working on it

AliSoftware:
@djbe you can review my last 2 commits which should address your 3 comments:

* no more support for deprecated params
* only one init, no more static func execute
* no more case of too many parameters, so no need for the intermediate struct
* and even no more `ParserCommand` intermediate type, as I now simply use an `extension` providing the `static func command() -> CommandType` on `Parser`…

The code is much simpler that way IMHO 🎉

Podfile
```diff
@@ -2,10 +2,10 @@ platform :osx, '10.9'
 use_frameworks!
 
 def common_pods
-  pod 'SwiftGenKit', '~> 1.1.0'
+  pod 'SwiftGenKit', :git => 'git@github.com:SwiftGen/SwiftGenKit.git', :branch => 'feature/parser-protocol'
```
AliSoftware: ⚠️  Reminder not to forget to change that back once we have merged https://github.com/SwiftGen/SwiftGenKit/pull/46

Podfile
```diff
@@ -2,10 +2,10 @@ platform :osx, '10.9'
...
   pod 'PathKit', '~> 0.8.0', :inhibit_warnings => true
   pod 'Stencil', '~> 0.9.0', :inhibit_warnings => true
-  pod 'StencilSwiftKit', '~> 1.0.2'
+  pod 'StencilSwiftKit', :git => 'git@github.com:SwiftGen/StencilSwiftKit.git'
```
AliSoftware: ⚠️ Reminder not to forget to change that back once we have released StencilSwiftKit 2.0

djbe:
Jup, changes look good 👍 

At some point, we might have deprecated options again (hopefully not), we'll then have to see how we can handle those. Maybe a generic "fail" that points to the migration guide, but we'll leave that up to our future selves 😆

AliSoftware:
Exactly. « never do tomorrow what you can do the day after tomorrow » 😆

AliSoftware:
TODO after I've got dinner, or maybe only tomorrow:

* Release SwiftGenKit 2.0
* Release StencilSwiftKit 2.0
* Modifiy this PR's Podfile to point to the version numbers instead of the git/branch
* Merge this PR

djbe:
Add the breaking tag again because we removed the deprecated options.

AliSoftware:
Submitted https://github.com/SwiftGen/SwiftGenKit/pull/48 & https://github.com/SwiftGen/StencilSwiftKit/pull/55 that have to be merged before we can amend SwiftGen's Podfile and merge this PR

AliSoftware:
@djbe I'll let you do a final check if you don't mind and hit merge? (I find it good practice in general that the person merging is someone different than the one who did the last non-trivial commit)

PR #333, opened 2017-09-30
Add a link between CONTRIBUTING and README #332

### LINKED ISSUE #332: Docs: How to run your code
I have contributions in mind for Swiftgen, but I have some issues to set up the project. In fact, it would be great to have a quick explanation on « How to test your modifications locally ». It will allow people with good will (like me) to contribute to the project in an easy way. The main goal would be to offer the shortest path from `I made a modification into the code` to `I ran it locally`. 

[We can inspire from what fastlane does](https://github.com/fastlane/fastlane/blob/master/ToolsAndDebugging.md)

I don’t think Swiftgen needs something as detailed as that. But just `How to run your code ?`.

juli1quere:
This PR is a really small update of CONTRIBUTING.md. I just added a link to the `Compile from source` installation method. It will help people who want to test their work on CLI.

This PR could close #332.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -111,3 +111,5 @@ Unless you're adding (or updating) dependencies, you won't have to run `pod inst
 ### Building & Running
 
 The Xcode workspace, by default, is configured to run SwiftGen built as an app bundle, with some default parameters. You can switch (and modify) these by going to `Product > Scheme > Edit Scheme... > Arguments`, and checking (and unchecking) the command you want to test.
+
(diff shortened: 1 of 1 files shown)


AliSoftware:
Tbf I'm not sure I understand the need for this addition, in the sense that "If you made some updates in the SwiftGen CLI" as you say, then you can already test it directly from Xcode, using the explanations already present in the previous paragraph: edit the scheme, provide the CLI arguments you want to test, and run from Xcode. 

Like you'd test any command-line target in Xcode actually (that's the recommended way by Apple). Which has the huge benefit of being debuggable in Xcode using breakpoints and all — while manually building the app using the Rakefile from the command line won't allow you to debug it

juli1quere:
Hum. That's a good point. 

In fact it looks that I took the issue from the wrong end. Since the build and run did not work for me, I though that I had to use `rake` manually. 

Sorry about that !

PR #337, opened 2017-10-06
Add support for an swiftgen.yml config file

### LINKED ISSUE #293: Support a `.swiftgen.yml` file to provide parameters
Similar to what Sourcery does with `.sourcery.yml`

Example:

```yaml
output_dir: Sources/Generated
images:
  template: swift3
  params:
    enumName: Images
  source: Sources/Assets.xcasset
  output: images.generated.swift
strings:
  template: swift3
  source: Sources/Base.lproj/Localizable.strings
  output: strings.generated.swift
```

Then the use can just call `swiftgen` in their Build Phase, and it will read the arguments from the `.swiftgen.yml` file found in the current directory, or in the directory specified by `--config <path>` if specified.

_Note: `output_dir` would be an optional key; if present, it will be used as a prefix for all `output` keys in the subdirectories to avoid repeating any common prefix path everywhere._

AliSoftware:
Closes #293

Work in Progress. In practice this PR is already working, only things missing are some cleanup tasks

* [x] Move the `listValues` private function in `ConfigCommands.swift` to `StencilSwiftKit.Parameters` (https://github.com/SwiftGen/StencilSwiftKit/pull/70)
* [x] Make `swiftgen` with no subcommand fallback to `swiftgen config run` (requires a new release of Commander featuring https://github.com/kylef/Commander/pull/52)
* [x] Add more unit tests for parsing the config file
* [x] Add documentation (Mention in README + Dedicated markdown for full config syntax/keys?)
* [x] CHANGELOG entry

---

I've added two entries for the possible arguments to the Xcode scheme, so that we can Cmd-R on the project to automatically run the following commands easily from Xcode and see the result in the console:
* `swiftgen config lint --config <path_to_fixture_swiftgen-bad.yml>`
* `swiftgen config run --config <path_to_fixture_swiftgen-2.yml> --verbose`

_Note: To make the `config run` command above work you'll have to manually create the output target directory (`mkdir /tmp/swiftgen`) beforehand; if all works well, you should have an error telling it doesn't exist anyway if you don't_ :wink:


AliSoftware:
PS: I haven't found a nice way with @kylef 's Commander to allow `swiftgen` with no subcommand to be interpreted as `swiftgen config run` by default (like `swiftlint` is actually a shortcut for `swiftlint lint` for example).
I think we could be able to achieve this using a custom parsing code (taking `init(parser: ArgumentParser)` or something?) but that would mean having to recreate all the logic ourselves for all the rest of the CLI handling (like triggering the help, detecting `--version`, and so on)?
So not sure if there's a better solution to have that?

kylef:
@AliSoftware Does https://github.com/kylef/Commander/pull/52 allow you to do what you are looking for?

AliSoftware:
@kylef totally, that's exactly what I was looking for indeed! 👌

kylef:
@AliSoftware Let me try rebase that and make a update with some of the other PRs tonight or over the weekend to unblock you.

AliSoftware:
Cool, thanks a lot! ❤️

AliSoftware:
@djbe Open question about the name of the YAML keys:

* I used `sources` for the key providing the input files to parse, probably without thinking about it much and mainly because that's what Sourcery and SwiftLint use.
* But in our case I now realize that this might not be the most appropriate name, since it could make people think about "source code", while what we expect here are resources (fonts, asset catalogs, …)

Which key name do you think would be better instead in that context? `input`? `inputs`? (as the key accepts either a single path or an array of paths so maybe the plural form is more appropriate)? `files`? `paths`? I'm open to suggestions :wink:

djbe:
Yeah `paths` or `inputs` are better options. You could make a case for either of those, so go with `paths` 😄

AliSoftware:
Ping: @djbe This is now ready for review!!

Sources/Commander/ConfigCommands.swift
```diff
@@ -0,0 +1,96 @@
...
+let configRunCommand = command(
+  Option<Path>("config", default: "swiftgen.yml", flag: "c",
+               description: "Path to the configuration file to use",
+               validator: checkPath(type: "config file") { $0.isFile }),
+  Flag("verbose", default: false, flag: "v",
+       description: "Print each command being executed")
+) { file, verbose in
+  let config = try Config(file: file)
+
+  if verbose {
```
djbe: Not really needed right now, but wouldn't it be nicer to have a logger system that accepts a logging level (info or above, warning or above, etc...) and the verbose flag modifies this (global) logger level? See comment below in the actual logger.

Sources/Commander/OutputDestination.swift
```diff
@@ -0,0 +1,46 @@
...
+//  swiftgen
+//
+//  Created by Olivier HALLIGON on 11/10/2017.
+//  Copyright © 2017 AliSoftware. All rights reserved.
+//
+
+import PathKit
+import Commander
+
+enum OutputDestination: ArgumentConvertible {
```
djbe: I never noticed this before, but it's a really cool idea! 👍

Sources/Logs.swift
```diff
@@ -0,0 +1,53 @@
...
+    if let termType = getenv("TERM"), String(cString: termType).lowercased() != "dumb" &&
+      isatty(fileno(stdout)) != 0 {
+      return "\(self)\(string)\(ANSIColor.reset)"
+    } else {
+      return string
+    }
+  }
+}
+
+func logMessage(_ level: LogLevel, _ string: CustomStringConvertible) {
```
djbe: It might be nicer to have a logger system that accepts a logging level (info or above, warning or above, etc...) and filters out (does not log) messages below the current log level. When we switch on "verbose" logging, we'd set the logger's level to `info` instead of `warning`.
- No longer need to check at the log point if it should be logged
- Introduce even more verbose logging (-vvv) at some point
- Introduce a silent mode at some point (disables all logging)
- Have the plugins (swiftgenkit) accept the logger to do their own logging if needed

Might be worth it checking some of the minimal existing logging libraries for ideas (and maybe dependency).

djbe: Also, if we decide to do this, the `_ string` message should be lazily evaluated to avoid processing expensing messages.

Sources/TemplateRef.swift
```diff
@@ -0,0 +1,84 @@
...
+//  TemplateRef.swift
+//  swiftgen
+//
+//  Created by Olivier HALLIGON on 11/10/2017.
+//  Copyright © 2017 AliSoftware. All rights reserved.
+//
+
+import PathKit
+
+enum TemplateRef {
```
djbe: Nice! (a bit similar to the output destination, I really like this 👍 )

Sources/Config.swift
```diff
@@ -0,0 +1,194 @@
+//
```
djbe: Might be worth splitting this file up right now, as it could get quite long once we add more features such as pipe-ing, etc..?

Tests/SwiftGenTests/ConfigTests.swift
```diff
@@ -0,0 +1,267 @@
...
+class ConfigTests: XCTestCase {
+
+  func testReadConfigWithParams() throws {
+    guard let path = Bundle(for: type(of: self)).path(forResource: "config-with-params", ofType: "yml") else {
+      fatalError("Fixture not found")
+    }
+    let file = Path(path)
+    do {
+      let config = try Config(file: file)
+      XCTAssertEqual(config.outputDir, "Common/Generated")
```
djbe: Noticed this really cool library by merowing in the slack recently:
https://github.com/krzysztofzablocki/Difference

Could be useful here, and if not, it's definitely useful for the context tests.

Tests/SwiftGenTests/ConfigTests.swift
```diff
@@ -0,0 +1,267 @@
...
+      XCTAssertEqual(stringsEntries.count, 1)
+      guard let entry = stringsEntries.first else {
+        return XCTFail("Can't get strings only config entry")
+      }
+      XCTAssertEqual(entry.paths, ["Sources1/Folder"])
+      if case TemplateRef.name("structured-swift3") = entry.template { /* OK */ } else {
+        XCTFail("Unexpected template")
+      }
+      XCTAssertEqual(entry.output, "strings.swift")
+      XCTAssertEqual(entry.parameters["foo"] as? Int, 5)
```
djbe: Anyway, these current tests seem a bit hard to read? Especially for params, where we could just test if the parameters are equal to a dictionary we provide (see code in SwiftGenKit for testing dictionaries).

Tests/SwiftGenTests/ConfigTests.swift
```diff
@@ -0,0 +1,267 @@
...
+        XCTAssertEqual(
+          String(describing: e), expectedError,
+          "Trying to parse config file \(configFile) should have thrown " +
+          "error \(expectedError) but threw \(e) instead."
+        )
+      }
+    }
+  }
+
+  // MARK: Linting
```
djbe: Maybe move the config linting to another test file

Podfile
```diff
@@ -5,14 +5,25 @@ def common_pods
...
 
 target 'UnitTests' do
   common_pods
 end
+
+post_install do |installer|
+  swift4_ready_pods = %w(PathKit Commander Yams)
+  installer.pods_project.targets.each do |target|
+    target.build_configurations.each do |config|
+      version = swift4_ready_pods.include?(target.name) ? '4.0' : '3.2'
```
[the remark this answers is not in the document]
AliSoftware: Yeah, thought the same, Might be one of the next step after the release, alongside with the big merge 😉
Tried to do it quickly during that current PR but was afraid to go too much off-track :P

Sources/Commander/ConfigCommands.swift
AliSoftware: Yeah, thought the same too. For now I didn't want to refactor _everything_, that PR was already the occasion of a little refactoring of other stuff (reorg files, finally making TemplateRef an enum as it was always meant to be exclusive choice, making it possible to provide the logging method for `Config.lint` so I can write proper tests on it to check if we logged the proper stuff…)

But for now I think it's not the most urgent thing to do. That reworking of `printError(string:)` to the new simplistic but level-based logging function was just a first step. Providing the log level on the CLI means that all commands should support it, meaning that we might need to tweak with Commander and the ArgumentParser even further, so that's be for later in a less urgent PR 😜

Sources/Config.swift
AliSoftware: Yeah, good idea, will try

Sources/Logs.swift
AliSoftware: Def a good idea for later, I agree (but I wanna do the big merge first, otherwise we'll forever postpone it)

Sources/TemplateRef.swift
AliSoftware: Yeah, it was about time we turn that exclusive choice of name vs path into an enum, I wonder why I didn't do it sooner 😄

AliSoftware: I even wondered if it wouldn't be worth it making it `ArgumentConvertible` like `OutputDestination`. But that would mean merging the `-t` and `-p` into one single option (and interpreting it as a `TemplateRef.path` if it contains a `/` and a `TemplateRef.name` otherwise, for example?). Would be too much breaking change for this, at least for now.

Tests/SwiftGenTests/ConfigTests.swift
AliSoftware: Yeah, I thought about it at some point. For comparing the generated files in SwiftGenKit it's no good, as we're comparing basically long texts there and not structures, but for here where we compare `Config` instances, it might be worth it 🤔

Tests/SwiftGenTests/ConfigTests.swift
AliSoftware: You means [this](https://github.com/SwiftGen/SwiftGenKit/blob/master/Tests/SwiftGenKitTests/TestsHelper.swift#L19-L72)? Yeah, would definitely help readability I guess!

Tests/SwiftGenTests/ConfigTests.swift
AliSoftware: Good point, will do.

AliSoftware:
@djbe I pushed some changes to help tests readability, thanks for the feedback.

I tried `Difference`, but it's no help in our case, because it compares Mirrors child by child and that doesn't work for unordered collections like dictionaries, and expects objects to be `Equatable`. So

* Either I had to make `Config` and `Config.Entry` both be `Equatable`, which would require to make `[String: Any]` comparisons between `lhs` and `rhs` in those implementations, leading to the same origin of the problem
* And I didn't want to introduce all that dependencies and complex comparisons of `[String: Any]` in the main implementation of `Config` in the main tard, while there's no need for it to be `Equatable` outside of tests
* And even trying to use `Difference.diff` on `[String: Any]` dictionaries to compare `Config.Entry.parameters` didn't work either, because again they are dictionaries so unordered, and thus I go errors like { "keys: expected foo, got bar" + "keys: expected bar, got foo" } as a result

In the end, `Difference` is (1) not really useful to check if 2 things are equal, but to generate a nice message when you already know that Equatable objects differ (2) is not reliable on unordered collections, so I finally didn't include the Pod at all.


Instead, I found a nicer and easiest way to compare `[String: Any]` dictionaries, which involves `Foundation`'s `NSDictionary`, but since we only manipulate `NSDictionary`-compatible/convertible values (`String`, `NSNumber`, arrays and dicts) in our `parameters`, that works out really nicely and is way easier and shorter than what we have in SwiftGenKit. We could even probably consider switching to that same trick in SwiftGenKit tests I guess.

AliSoftware:
btw, filed some issues in `Difference`'s repo to let Krzysztof know: https://github.com/krzysztofzablocki/Difference/issues/1

Sources/TemplateRef.swift
djbe: It would be too much of a breaking change, and a bit confusing for users. Also, paths might not always include a `/` character, so how would that even work?

AliSoftware: Yeah I also thought of other options (telling people to use `./foo` to disambiguate between name and local-path-in-same-dir but not ideal. Or on the contrary having to prefix template name with `:` or something to refer to a name instead of a path (eg `-t :swift3`) but that would be even more breaking (and using a name is the most common usage compared to using as path so path should be the exception not the other way around)… in the end none of those scenarios feels ideal 😕 so let's keep it as is for now 😉

djbe:
Good points on the `Difference` library.

The reason I wrote the (more convoluted) dictionary comparison in SwiftGenKit is because it works with much bigger dictionaries, and it can be hard to track down the difference if it just dumps both of them. The implementation actually writes the path to where it differs, and why. (The output could be better in some cases, but it works).

The full implementation of dictionary comparison isn't needed for these tests, so should be ok for now.

AliSoftware:
Good point.

We should share that dict diff helper in all those libs anyway. But it will be easier to do so once we merge everything back in a monorepo, so let's wait for that first before using the one in SwiftGenKit in both (as the error messages on non-equality is better in yours)

Issue #346, opened 2017-10-23
New version fails to generate sources but exists with code 0


NachoSoto:
_Version: `SwiftGen v5.2.0 (Stencil v0.9.0, StencilSwiftKit v2.3.0, SwiftGenKit v2.1.1)`_

I wasted a good 20 minutes trying to figure out why my build on Travis was failing. Turns out the latest `SwiftGen` stopped generating sources, but I didn't see this locally because the old sources worked, and new ones weren't being generated, despite the build phase **not failing**:

```
Error: You need to choose EITHER a named template OR a template path. Found name 'structured-swift3' and path ''
```

This is my script:
```bash
swiftgen fonts "$FRAMEWORK_DIR/../app/resources/fonts" --output "$OUTPUT_DIR/Fonts.swift" -t swift3
swiftgen colors "$FRAMEWORK_DIR/resources/Colors.clr" --output "$OUTPUT_DIR/Colors.swift" -t swift3
swiftgen strings "$FRAMEWORK_DIR/en.lproj/Localizable.strings" --output "$OUTPUT_DIR/Strings.swift" -t structured-swift3
swiftgen xcassets "$FRAMEWORK_DIR/Assets.xcassets" --output "$OUTPUT_DIR/Images.swift" -t swift3 --param noAllValues
```

I still don't really know why this broke, but either way this error should make `swiftgen` fail instead of continuing silently.

NachoSoto:
I'm really confused about what this error is trying to tell me. The docs in the `README` don't make sense either:

> Note: you should specify one and one template when invoking SwiftGen

djbe:
I think the word `only` is missing there: `one and only one template`.

AliSoftware:
Did you set `set -e` at the beginning of your script?

Because in a shell script, if one command fails, that doesn't make the whole sh script fail but the shell continues to execute the next line in the script. So the exit code of the whole script is only the exit code of the last command.
At least that's the default sh behavior. Setting `set -e` changes that behavior of your shell so it fails at soon as one command in the shell fails.

NachoSoto:
> Did you set set -e at the beginning of your script?

I always do, yes:

<img width="751" alt="screen shot 2017-10-23 at 12 02 58 pm" src="https://user-images.githubusercontent.com/685609/31907917-21658e14-b7ea-11e7-83c2-7d7c5373fb10.png">

djbe:
Could it be there are some spaces after one line?

[...]

PR #347, opened 2017-10-23
Fix TemplateRef to handle empty string paths

djbe:
Fixes one part of #346.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@
 * Fix SwiftGen no longer working using CLI parameters (instead of config file).  
   [David Jennes](https://github.com/djbe) 
   [#347](https://github.com/SwiftGen/SwiftGen/pull/347)
+* Errors now properly exit with a non-zero exit code.  
+  [Olivier Halligon](https://github.com/AliSoftware)
+  [#348](https://github.com/SwiftGen/SwiftGen/pull/348)
 
 ### Breaking Changes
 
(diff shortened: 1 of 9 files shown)


NachoSoto:
Thanks!

AliSoftware:
I'm not sure I agree with the direction of that change. I'm not a fan of defaulting to a `Path("")` then checking `path.string.isEmpty`, feels like a wrong usage of `Path`, building an invalid `Path` instance (with an empty path component) when we know that what we mean here is "no path provided"

That's why I conceptually used `Path?` here in the first place.

What about keeping that `Path?` in the init and that `nil` vs `let path?` in the `switch`… and defaulting the `templatePath` Commander option to `nil` instead (to reflect the concept of "no value provided", rather than "a value of empty string provided")? Wouldn't that be more suitable and logical?

AliSoftware:
Ok, I tried using `Path?` but it's complicated with Commander indeed, because `Optional<Path>` doesn't conform to `ArgumentConvertible`. Dammit.

djbe:
There's an (old) PR on Commander to add optional options (option-ception):
https://github.com/kylef/Commander/pull/48

AliSoftware:
Good find. We might want to ping Kyle on this one.

In the meantime, I'm gonna switch the argument parsing to consider the argument as a `String` for consistency, and convert it to a `Path` manually only after Commander parsing (basically in `TemplateRef.init`). Feels more consistent than to instantiate a `Path` with empty string. Commit pending.

PR #354, opened 2017-10-26
Fix CircleCI builds (for Xcode9 version search) + sync:all_files

AliSoftware:
See https://github.com/SwiftGen/templates/pull/86 — porting changes/fixes back here

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,9 +1,6 @@
 machine:
   xcode:
     version: 9.0
-  pre:
-    - sudo launchctl load -w /System/Library/LaunchDaemons/com.apple.metadata.mds.plist
-    - mdimport /Applications/Xcode*
 
 checkout:
   post:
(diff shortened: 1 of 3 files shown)


djbe:
If I understand this right, we don't use spotlight anymore if the version directly matches. If it doesn't, we fall back to spotlight to find one that does, right?

If so, this might probably still break when CI has Xcode 9 and 10, 9 is "selected", but we need 10.

djbe:
Note that my previous comment is just that, not a request for change, because I don't think there's much we can do about it.

AliSoftware:
Yeah, see the discussion with Danielle on the Slack (private messages), Circle doesn't have Spotlight enabled anymore, and even with `launchctl` this shouldn't even work (if it worked before, that's because of luck with race condition probably)

So on CI we'll always have the right Xcode version selected anyway (because `circle.yml` specified the machine to use, so this should always match and early-exit, and on other machines (like contributor's macs) at least it still uses spotlight to search for other versions as a fallback

AliSoftware:
I'll try not to forget to apply the same on SwiftGenKit and StencilSwiftKit this evening after work 😉

PR #356, opened 2017-11-05
Merge SwiftGenKit and templates back into this repository

### LINKED ISSUE #326: Merge templates and SwiftGenKit back into this repository
We initially split off the templates and SwiftGenKit repositories to have a clearer separation of concerns, and to force us to write better and more complete tests. Over the last few releases we have noticed that, while we achieved the previously stated goals, it has become quite burdensome to develop features that cross these repositories, such as new parsers for example.

We've decided that we'll leave the StencilSwiftKit repository separate, as it is used by other projects, and we seldomly have features that are both in that repository and other SwiftGen repositories. The `templates` and the `SwiftGenKit` repositores will be merged back into this one, but we'll try to keep code separate (like `GenumKit` before the split).

This issue is for discussing what files of each related repository should go where in this repository. The SwiftGenKit source code can be left in a separate folder, but we should organise the documentation and test code in a certain way to keep it clear that they belong with that code.

djbe:
Fixes #326. As mentioned before a few times (big merge) a while ago, we're merging the SwiftGenKit and templates repositories back into this repository.

We're doing this mainly because the 3 repositories are so interwoven with each other, the separation leads to a lot of maintenance work, and is probably quite confusing for new contributors. Especially when we have features that span across multiple repositories.

- [ ] Unify the `TestHelpers.swift` files into one.
- [ ] Figure out what to do with the changelogs from the other repositories.
- [ ] Adapt the release scripts now that the templates repo has been integrated.
- [ ] Adapt the release scripts now that the SwiftGenKit repo has been integrated.


djbe:
Note that most of the files have been merged, except for:
- SwiftGenKit changelog
- Templates changelog
- Templates readme

AliSoftware:
Below is what we want the `Package.swift` to probably look like once we finish the migration. I think the file structure should reflect this accordingly to make it more understandable.

See [SPM doc here](https://github.com/apple/swift-package-manager/blob/master/Documentation/PackageDescriptionV4.md)

> All targets should be declared in the Package.swift manifest file. Unless the relative path of the target is declared, the Package Manager will look for a directory matching the name of the target in these places:
>
> * Regular targets: package root, `Sources`, `Source`, `src`, `srcs`.
> * Test targets: `Tests`, package root, `Sources`, `Source`, `src`, `srcs`.

```swift
import PackageDescription

let package = Package(
    name: "SwiftGen",
    products: [
        .library(name: "SwiftGenKit", targets: ["SwiftGenKit"]),
    ],
    targets: [
        .target(name: "swiftgen", dependencies: ["SwiftGenKit"]),
        .target(name: "SwiftGenKit")
    ]
    dependencies: [
        .package(url: "https://github.com/SwiftGen/StencilSwiftKit"),
        .package(url: "https://github.com/kylef/Commander"),
        .package(url: "https://github.com/kylef/PathKit"),
        …
    ],
)
```

So even if SPM v4 allows sources for targets to be located elsewhere if we provide the explicit path where the sources are, I think it would be nice to respect the conventions, leading to the following structure:

```
(repo root)
 '-- Package.swift
 '-- Sources/
 |    '-- SwiftGenKit/
 |    '-- swiftgen/
 |         '-- main.swift
 |         '-- …
 '-- Tests/
 |    '-- SwiftGenKitTests/
 |    '-- SwiftGenTests/
 |    '-- TemplatesTests/
 '-- Templates/
 |    '-- xcassets
 |    '-- …
 '-- Documentation/
      '-- Configuration File.md
      '-- Using Custom Templates.md
      '-- Creating Your Templates.md
      '-- SwiftGenKit Contexts/
      |    '-- Assets.md
      |    '-- …
      '-- Templates/
           '-- xcassets/
                '-- swift3.md
```
This would also have the advantage of moving some stuff away from the repo root (especially `Pods/`)

We're almost at this structure, but I think:

1. migrating all the documentation to `Documentation/` at the root (instead of some in `Documentation` and some in `SwiftGenKit/Documentation` would make the documentation easier to find. The disadvantage of that is that SwiftGenKit's documentation would no longer be "attached" to the corresponding module and isolated (if SPM supports exporting documentation of a dependency one day…) but I think it's more important to make it easier to find the doc for our users.
2. Migrating the swiftgen & SwiftGenKit sources as subdirectories of a `/Sources/` root repo (instead of the current `Sources` directory directly containing the source only for swiftgen CLI)

AliSoftware:
I'm not sure I understand the first part: can't you `inhibit_warnings: true` (or even `!inhibit_all_warnings` at top level) in the main `Podfile` when you list Kanna and PathKit pods?

djbe:
No if I do that, I get a warning from pod install that there are different inhibit warning settings for the same pod in different targets.

djbe:
For example:
```
[!] The pod `PathKit` is linked to different targets (["Pods-swiftgen", "Pods-SwiftGenKit", "Pods-SwiftGen UnitTests", "Pods-SwiftGenKit UnitTests", "Pods-Templates UnitTests"]), which contain different settings to inhibit warnings. CocoaPods does not currently support different settings and will fall back to your preference set in the root target definition.
```

AliSoftware:
Oh, wait, they're not listed directly because you use `podspec :path => …` I see. I'm not sure if we can `inhibit_warnings: true` on a `podspec` directive? And what if we `inhibit_all_warnings!`, does it affect transitive dependencies brought by the `podspec` directive?

djbe:
We can't, I tried 😄

AliSoftware:
Ok, I guess it's one more argument to drop the podspec then :wink: you win ^^

djbe:
Hey I don't know, maybe leave the podspec and create PRs on the corresponding repos. (playing devil's advocate here)

AliSoftware:
I've started a GitHub Project to list all the small stuff and nitpickings we should do after merging this PR to clean up the last pieces: https://github.com/orgs/SwiftGen/projects/2

Podfile
```diff
@@ -1,24 +1,32 @@
...
 
-target 'swiftgen' do
-  common_pods
-  pod 'Commander', '~> 0.8'
+  target 'SwiftGen UnitTests' do
+    inherit! :complete
+  end
+
+  target 'Templates UnitTests' do
+    inherit! :complete
```
AliSoftware: Did you try with simply `inherit! :search_path` instead of `:complete`?

Podfile
```diff
@@ -1,24 +1,32 @@
...
+  end
 end
 
-target 'UnitTests' do
-  common_pods
+target 'SwiftGenKit' do
+  podspec :path => 'SwiftGenKit.podspec'
+
+  target 'SwiftGenKit UnitTests' do
+    inherit! :complete
```
AliSoftware: Same q

AliSoftware (commented):
Feedback on the structure:

* Better overall I think
* Seems strange to have `Tests/Fixtures` (resources for SGK to parse) and `Tests/SwiftGenTests/Fixtures` (with yml config files). We should either move `Tests/Fixtures` inside `Tests/SwiftGenKit` (if those fixtures are really indeed only used for SwiftGenKit and nothing else), or move everything to `Tests/Fixtures` (i.e. move the `*.yml` fixtures in `Tests/Fixtures/ConfigFiles` or something)
* Nitpicking: we should rename `Tests/Fixtures/Contexts/` to something more explicit (I forgot what it was until I looked inside), probably `StencilContexts` or something?

Didn't have much time trying to build the Xcode project and play with it (mostly reviewed it via GitHub UI and SublimeText), only thing I can say about the xcodeproj:

* After opening it, the "Generate SwiftGenKit Contexts" scheme was selected by default (which only works with Cmd-U not Cmd-B, so for newcomers can be even more confusing)
* After selecting the `swiftgen` scheme and hitting Cmd-B, I got a compile error, because `bundler` couldn't find the expected version of `rake`. I just went to the terminal and ran `bundle install` then re-ran the build and it passed

Both those are not that great for newcomers / new contributors, so we might try to improve that in some way (more explicit error messages, reordering schemes, improve CONTRIBUTING, …). Maybe for a separate PR, idk, will add that task to the GitHub project.

Podfile
djbe: Yeah, it's not enough it seems, because it starts complaining about `no such module StencilSwiftKit`.

Podfile
djbe: Same a.

djbe:
Fixtures
The `Fixtures` folder contains stuff only for SGKTests, but the `stub-env` belongs with TemplatesTests. To be more correct, the `stub-env` files are for the output compilation tests, which aren't part of Xcode tests (same for the `Expected` folder). The `Contexts` files is needed for both SGKTests and TemplatesTests, so those need to stay in a common folder.

I'll rename `Contexts` to `StencilContexts`. What do I do with the rest?

----

There's no way to control the order of schemes afaik, I tried moving the `xcschememanagement.plist` file (with the orders) to `xcshareddata/xcschemes` with no effect. We could rename the helper schemes to something that's alphabetically after `swiftgen`, for example: `SwiftGenKit - Generate Contexts`.

No idea what we can do about the bundler issue? Maybe use `bundle check` ([docs](https://bundler.io/man/bundle-check.1.html)) to verify, and if need be show an appropriate error?

AliSoftware:
I think for the rest (of Tests/Fixtures), putting everything in a common folder would still make sense?

djbe:
So to summarise:
- Move Tests/Fixtures (not stub-env) to Tests/SwiftGenKit/Fixtures
- Rename Tests/Contexts to Tests/StencilContexts
- Move Tests/Expected to Tests/TemplatesTests/Expected
- Move Tests/Fixtures/stub-env to Tests/stub-env (not sure about this one)

AliSoftware:
Mmmh I feel I'm lost here. Let me look at the folders in the Finder when I get to a computer to get a clearer idea

AliSoftware:
Ok, I'm not sure of anything rn. e.g. aren't`Tests/Fixtures` used both for SwiftGenKit (as input to test that the contexts generated from them match what's in StencilContexts) and by TemplateTests (don't we use them as input to check the generated swift code from our templates, instead of starting from the StencilContexts directly?)

AliSoftware:
So if I'm summarise all this correctly, here are all the fixtures used by tests as input and output:

| Sub-Folder of `/Tests`         | SwiftGenTests | SwiftGenKitTests | TemplatesTests | Compilation Tests |
|--------------------------------|---------------|------------------|----------------|-------------------|
| `Fixtures/<subcmd>/*` (rsrc)   | -             | Used as input    | -              | -                 |
| `Contexts/<subcmd>/*.plist`    | -             | Used as output   | Used as input  | -                 |
| `Expected/<subcmd>/*.swift`    | -             | -                | Used as output | Used as input     |
| `Fixtures/stub-env`            | -             | -                | -              | Used as input     |
| `SwiftGenTests/Fixtures/*.yml` | Used as input | -                | -              | -                 |

Did I fill this table correctly?

AliSoftware:
So as discussed on Slack, we'll probably do this org inside `Tests/`:

 - `Fixtures/`
   - `Resources/` (old `Tests/Fixtures/*`, except `stub-env`)
   - `StencilContexts/` (old `Tests/Contexts`)
   - `Expected/`
   - `CompilationEnvironment/` (old `Test/Fixtures/stub-env`)
   - `Configs/` (old `SwiftGenTests/Fixtures/*.yml`)
 - `SwiftGenKitTests/` (unchanged)
 - `SwiftGenTests/` (move `Fixtures` subfolder out to `Fixtures/Configurations/`)
 - `TemplatesTests/`
```

AliSoftware:
Or, explained differently:
conspiracy theory meme

AliSoftware:
I'll take a look at the CI tonight.

Once fixed, should we merge this and then tackle the merging of CHANGELOG + TestHelpers in a separate PR? Or finish that in this one?

(Reminder that there's https://github.com/orgs/SwiftGen/projects/2? to track what's left to do after this merge)

djbe:
Might be best to merge as is, and tackle remaining issues later. That way I can start migrating PRs from other repositories to this one.

AliSoftware:
Noted. I'll test on my work machine tomorrow morning and merge right after 👍 

I've created 2 cards in the GH Project (one for CHANGELOG one for TestHelpers merge) and reordered the cards in a somewhat priority order (not really strict but more like logical next course of actions)

PR #364, opened 2017-12-10
Interface Builder: Typed segues on a per scene basis

### LINKED ISSUE #362: Using same view controller with same segue identifier in multiple storyboards
Hi all.

Here's a scenario a colleague of mine has discovered just recently. About a week after we integrated `swift4` template for storyboards.

Let's say I have view controller `VC`.
Then I have 2 storyboards: `SB1` and `SB2`.
Now I want to use `VC` in both storyboards.
Then my view controller has a certain segue connected to it, the segue takes user to "next" page, whatever that next page is in each specific storyboard.
This is where it gets interesting, the segue identifier is something like `next` and it's **same** in both storyboards: `SB1` and `SB2`.

When it comes to autogenerated code, it's all good. There are no conflicts, what I get is something like:

```swift
enum StoryboardSegue {
  enum S1 {
    case next
  }
  enum S2 {
    case next
  }
}
```

Everything is OK until I need to perform that `next` segue in the view controller code:

```swift
func someAction() {
  perform(segue: StoryboardSegue.S1.next, sender: nil)
  // but when   perform(segue: StoryboardSegue.S2.next, sender: nil) ?
}
```

This code is now misleading, I kind of need to know from which storyboard the view controller was loaded from to know whether need to perform `StoryboardSegue.S1.next` or `StoryboardSegue.S2.next`.

Previously, with unsafe string literals, the code would read more clear:

```swift
performSegue(withIdentifier: "next", sender: nil)
```

Which still is prone to error, such as if segue ID is not the same in 2 storyboards, which is a developer error in the end.

The more generic problem here is how to deal with view controllers code which is being reused in more than 1 storyboard.

Probably this confusion is due to the fact that `StoryboardScene` and `StoryboardSegue` are two independent enums. While in runtime, all segues are properties of a concrete storyboard.
It almost feels as if it should be OK to be able to do something like `viewController.storyboard.segues.next` or `viewController.storyboard.nextSegue`or alike.

I'm not even sure if that could be achieved with code generation.
The same view controller can have different segues connected to it when used in different storyboards, which makes the whole setup very dynamic, something that could be difficult to cover up with statically generated code.

djbe:
Transfer PRs https://github.com/SwiftGen/SwiftGenKit/pull/54 and https://github.com/SwiftGen/templates/pull/79.
Fixes #362.

## SwiftgenKit Changes

This PR improves the storyboards parser in 2 ways:
- Know which segues start from which scenes
- Know the type information for a segue destination
The PR also refactors the Storyboards parser a bit (split up in files).

## Templates Changes

Uses the improvements added in SwiftGenKit to generate extensions for non-base classes (UIKit/AppKit), so that a user can:
- Perform a segue safely with an enum of segues that start from that scene (you can no longer incorrectly start a segue from a VC that doesn't support it).
- Prepare for a segue in a type safe way:
  - Have an exhaustive list of possible cases, including one for unnamed segues.
  - Get the segue's destination in the correct type (no need to cast it).
  - Get the segue itself as the correct subclass if you're using custom segues.

----------

For performing a segue, the user goes from:
```swift
perform(segue: StoryboardSegue.MyStoryboard.nextStep)
```
To:
```swift
perform(segue: .nextStep)
```

For preparing for a segue, the user goes from:
```swift
override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
  // autocomplete for cases won't work because of optionals
  switch segueEnum {
  case .nextStep?:
    guard let destination = segue.destination as? NextStepViewController else { return }
    // destination is casted version
  case .login?:
    guard let destination = segue.destination as? LoginViewController else { return }
    // destination is casted version
  default:
    // Handle segues that are for other scenes, or unnamed segues
  }
}
```
To:
```swift
override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
  switch TypedStoryboardSegue(segue: segue) {
  case .nextStep(let destination):
    // destination is casted version
  case .login(let destination):
    // destination is casted version
  case .unnamedSegue:
    // This case is added if you have unnamed segues (with identifier "")
  }
}
```

### DIFF
--- a/Sources/SwiftGenCLI/templates/ib/scenes-swift4.stencil
+++ b/Sources/SwiftGenCLI/templates/ib/scenes-swift4.stencil
@@ -42,7 +42,7 @@ import {{module}}
     {% set sceneClass %}{% call className storyboard.initialScene %}{% endset %}
     {{accessModifier}} static let initialScene = InitialSceneType<{{sceneClass}}>(storyboard: {{storyboardName}}.self)
     {% endif %}
-    {% for scene in storyboard.scenes %}
+    {% for scene in storyboard.scenes where scene.identifier %}
 
     {% set sceneID %}{{scene.identifier|swiftIdentifier:"pretty"|lowerFirstWord|escapeReservedKeywords}}{% endset %}
     {% set sceneClass %}{% call className scene %}{% endset %}
--- a/Sources/SwiftGenCLI/templates/ib/scenes-swift5.stencil
+++ b/Sources/SwiftGenCLI/templates/ib/scenes-swift5.stencil
@@ -42,7 +42,7 @@ import {{module}}
     {% set sceneClass %}{% call className storyboard.initialScene %}{% endset %}
     {{accessModifier}} static let initialScene = InitialSceneType<{{sceneClass}}>(storyboard: {{storyboardName}}.self)
     {% endif %}
-    {% for scene in storyboard.scenes %}
+    {% for scene in storyboard.scenes where scene.identifier %}
 
     {% set sceneID %}{{scene.identifier|swiftIdentifier:"pretty"|lowerFirstWord|escapeReservedKeywords}}{% endset %}
     {% set sceneClass %}{% call className scene %}{% endset %}
--- a/Sources/SwiftGenCLI/templates/ib/segues-swift4.stencil
+++ b/Sources/SwiftGenCLI/templates/ib/segues-swift4.stencil
@@ -16,10 +16,100 @@ import {{module}}
 // MARK: - Storyboard Segues
 
 // swiftlint:disable explicit_type_interface identifier_name line_length type_body_length type_name
-{{accessModifier}} enum {{param.enumName|default:"StoryboardSegue"}} {
+{% macro moduleName item %}{% filter removeNewlines %}
+  {% if item.moduleIsPlaceholder %}
+    {{ env.PRODUCT_MODULE_NAME|default:param.module }}
+  {% else %}
+    {{ item.module }}
+  {% endif %}
+{% endfilter %}{% endmacro %}
+{% macro className item %}{% filter removeNewlines %}
+  {% set module %}{% call moduleName item %}{% endset %}
+  {% if module and ( not param.ignoreTargetModule or module != env.PRODUCT_MODULE_NAME and module != param.module ) %}
+    {{module}}.
+  {% endif %}
+  {{item.type}}
+{% endfilter %}{% endmacro %}
+{% set enumName %}{{param.enumName|default:"StoryboardSegue"}}{% endset %}
+{% set unnamedSegue %}{{param.unnamedSegueCaseName|default:"unnamedSegue"}}{% endset %}
+{% for scene in customSceneTypes where scene.segues %}
+{% set sceneClass %}{% call className scene %}{% endset %}
+{{accessModifier}} extension {{sceneClass}} {
+  {{accessModifier}} enum {{enumName}}: String {
+    {% for segue in scene.segues where segue.identifier %}
+    {% set segueID %}{{segue.identifier|swiftIdentifier:"pretty"|lowerFirstWord}}{% endset %}
+    case {{segueID|escapeReservedKeywords}}{% if segueID != segue.identifier %} = "{{segue.identifier}}"{% endif %}
+    {% endfor %}
+  }
+
+  {{accessModifier}} func perform(segue: {{enumName}}, sender: Any? = nil) {
+    let identifier = {% if isAppKit %}NSStoryboardSegue.Identifier({% endif %}segue.rawValue{% if isAppKit %}){% endif %}
+    performSegue(withIdentifier: identifier, sender: sender)
+  }
+
+  {{accessModifier}} enum Typed{{enumName}} {
+    {% set hasUnnamedSegue %}{% for segue in scene.segues where not scene.identifier %}1{% endfor %}{% endset %}
+    {% for segue in scene.segues where segue.identifier %}
+    {% set segueID %}{{segue.identifier|swiftIdentifier:"pretty"|lowerFirstWord}}{% endset %}
+    case {{segueID|escapeReservedKeywords}}{% filter removeNewlines:"leading" %}
+      {% if segue.customClass or segue.destination %}(
+        {% if segue.destination %}destination: {% call className segue.destination %}{% endif %}
+        {% if segue.customClass and segue.destination %}, {% endif %}
+        {% if segue.customClass %}segue: {% call className segue %}{% endif %}
+      ){% endif %}
+    {% endfilter %}
+    {% endfor %}
+    {% if hasUnnamedSegue %}
+    case {{unnamedSegue}}
+    {% endif %}
+
+    // swiftlint:disable cyclomatic_complexity
+    init(segue: {{prefix}}StoryboardSegue) {
+      switch segue.identifier{% if isAppKit %}?.rawValue{% endif %} ?? "" {
+      {% for segue in scene.segues where segue.identifier %}
+      case "{{segue.identifier}}":
+        {% if segue.customClass %}
+        {% set segueClass %}{% call className segue %}{% endset %}
+        guard let segue = segue as? {{segueClass}} else {
+          fatalError("Segue '{{segue.identifier}}' is not of the expected type {{segueClass}}.")
+        }
+        {% endif %}
+        {% if segue.destination %}
+        {% set destinationClass %}{% call className segue.destination %}{% endset %}
+        {% if destinationClass != "UIKit.UIViewController" %}
+        guard let vc = segue.destination{% if isAppKit %}Controller{% endif %} as? {{destinationClass}} else {
+          fatalError("Destination of segue '{{segue.identifier}}' is not of the expected type {{destinationClass}}.")
+        }
+        {% else %}
+        let vc = segue.destination{% if isAppKit %}Controller{% endif %}
+        {% endif %}
+        {% endif %}
+        {% set segueID %}{{segue.identifier|swiftIdentifier:"pretty"|lowerFirstWord}}{% endset %}
+        self = .{{segueID|escapeReservedKeywords}}{% filter removeNewlines:"leading" %}
+          {% if segue.customClass or segue.destination %}(
+            {% if segue.destination %}destination: vc{% endif %}
+            {% if segue.customClass and segue.destination %}, {% endif %}
+            {% if segue.customClass %}segue: segue{% endif %}
+          ){%endif %}
+        {% endfilter %}
+      {% endfor %}
+      {% if hasUnnamedSegue %}
+      case "":
+        self = .{{unnamedSegue}}
+      {% endif %}
+      default:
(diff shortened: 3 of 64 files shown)


djbe:
@AliSoftware proposed another idea, that we could add to this template (or an alternative one):

```swift
protocol MyViewControllerTypedStoryboardSegue {
  func prepareSegueForEmbed(destination: EmbedDestinationVC)
  func prepareSegueForModal(destination: ModalDestionationVC)
  …
  func prepare(for segue: NSStoryboardSegue, sender: Any?)
}
extension MyViewControllerTypedStoryboardSegue {
  func prepare(for segue: NSStoryboardSegue, sender: Any?) {
    /* generate that switch/case implementation here which calls the proper prepareSegueForXXX(destionation:) method */
  }
}
```
There could be issues here because of `@objc` dispatch and stuff, but could be interesting.

Another proposal was:
```swift
/* Generated by SwiftGen */
protocol MyViewControllerTypedStoryboardSegue {
  func prepareSegueForEmbed(destination: EmbedDestinationVC)
  func prepareSegueForModal(destination: ModalDestionationVC)
  …
  func prepare(for segue: NSStoryboardSegue)
}
extension MyViewControllerTypedStoryboardSegue {
  func prepare(for segue: NSStoryboardSegue) {
    /* generate that switch/case implementation here which calls
        the proper prepareSegueForXXX(destionation:) method */
  }
}

/* Added by the user */
extension MyViewController: MyViewControlelrTypedStoryboardSegue {
  @objc func prepare(for segue: NSStoryboardSegue, sender: Any?) {
    self.prepare(for: segue)
  }
}
```
I'd probably rename the function here, because users want access to the sender argument so we'd need to pass it along.

parallaxe:
To know where the segues originate from would be a killer-feature for me to use SwiftGen. I have no opinion regarding the templates though.
Is it foreseeable that it will be merged soon? Is there something I could do to accelerate it?

djbe:
To begin with: any and all suggestions are welcome!

You could try mock-generating the code (manually) for one of your projects (or just 1 storyboard), and see if the generated code works for you? Are all features there that you might want? Does the functionality match with your project structure?

Maybe go through your project's current segues, their names and controller types, and whatever else you need when performing segues. See if there might be an issue (or not) with the generated code.

Sources/SwiftGenKit/Parsers/StoryboardParser/Storyboard+XML.swift
```diff
@@ -0,0 +1,107 @@
...
+
+private enum XML {
+  enum Scene {
+    static let initialVCXPath = "/*/@initialViewController"
+    static let targetRuntimeXPath = "/*/@targetRuntime"
+    static func initialSceneXPath(identifier: String) -> String {
+      return "/document/scenes/scene/objects/*[@sceneMemberID=\"viewController\" and @id=\"\(identifier)\"]"
+    }
+    static let sceneXPath = "/document/scenes/scene/objects/*[@sceneMemberID=\"viewController\"]"
+    static let sceneSegueXPath = "//connections/segue"
```
parallaxe: For my understanding, the StoryboardSegue-enum that is created inside the extensions of each custom Scene-class should only contain the segues that are started from that specific scene.
If this is the case, the xpath has to be "./connections/segue"

parallaxe:
I made a reduced-version of your storyboard-template which contains only the scene-extensions with an segue-enum containing all segues that are started from the scene. Essentially, that is all i would need for know. You can find it here https://gist.github.com/parallaxe/7815fc6a8ed98e2e66de20d65169d608
Though, to make it work as I expected it, I had to change one of the XPaths in the storyboard-parser - I added a comment to the specific line in the "File Changes"-part of the merge-request.

Sources/SwiftGenKit/Parsers/StoryboardParser/Storyboard+XML.swift
djbe: The reason for the `//` in `sceneSegueXPath` is to be able to match segues started from (for example):
- container views --> embed segues
- triggered by a button or cell

`sceneSegueXPath` is only used to match children of a scene object:
https://github.com/SwiftGen/SwiftGen/pull/364/files#diff-aab03c1cbf39dc54bf61cfae345dfc87R43

djbe:
This obviously needs a rebase, but before I continue dev. on this, a couple of issues:

- How do we handle subclasses? Let's say `ClassA` has segues `S1` and `S2`, and `SubclassB` has segues `S2` and `S3`. That the code generates a separate enum for each class isn't an issue, the segues correspond to each type. It's more the type/name collision (and autocompletion) that I'm worried about. 
- How do we handle "partial" types (with the `moduleIsPlaceholder` flag set to true, so no `module`)? The parser doesn't know what the eventual module will be, and currently `Stencil` doesn't have functionality to "join" arrays (or append). Nor do we have a "uniques" filter yet.

We could solve the latter by having the parser check what's in the environment/parameters list, but that's a less clean solution (no more separation of concerns), and we'd have to re-work our tests (again). If we want to solve this in the template, we're going to need some new array filters first (I think).

SwiftGen-Eve:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Error">
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please include a CHANGELOG entry to credit your work.  <br />
You can find it at <a href="https://github.com/SwiftGen/SwiftGen/blob/feature/scene-segues/CHANGELOG.md">CHANGELOG.md</a>.</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

---

📝 We use the following format for CHANGELOG entries:
```
* Interface Builder: Typed segues on a per scene basis.  
  [#364](https://github.com/SwiftGen/SwiftGen/pull/364)
  [@djbe](https://github.com/djbe)
```
:bulb: Don't forget to end the line describing your changes by a period and two spaces.


</p>

AliSoftware:
As for your two questions:

- Haven't really looked at what the generated code looks like yet (in bed from phone here), but not sure that will really be an issue? I mean first how often will both the parent and child classes have a segue named the same (`S2`), and then even if they do I think you'll always be able to disambiguate by prefixing the type explicitly. And if the need to be explicit bothers you as an end user then maybe you shouldn't have named your segues the same anyway? (Idk, I don't use segues so might be a bad judge about all that)
- I fully agree with the `unique` filter or some similar solution (like an additional keyword in the `{% for x in array uniquing x.name %}` tag or whatnot) in Stencil. There has been multiple occasions where I wanted it to avoid dupes and would have liked to have this (I think I even have an issue own about that from the early days of SwiftGen)

AliSoftware:
Ah that was #7 😉

djbe:
@AliSoftware This may better illustrate the situation:
segues

Would lead to this code:
```swift
class ClassA: NSViewController {
  enum TypedStoryboardSegue {
    case s1
    case s2
    init(_ segue: NSStoryboardSegue) { ... }
  }

  override func prepare(for segue: NSStoryboardSegue, sender: Any?) {
    switch TypedStoryboardSegue(segue) {
    case .s1:
      break
    case .s2:
      break
    }
  }
}

class ClassB: ClassA {
  enum TypedStoryboardSegue {
    case s1
    case s2
    case s3
    init(_ segue: NSStoryboardSegue) { ... }
  }

  override func prepare(for segue: NSStoryboardSegue, sender: Any?) {
    switch TypedStoryboardSegue(segue) { // ambiguous use of 'init'
    case .s1:
      break
    case .s2:
      break
    case .s3:
      break
    }
  }
}
```

The problem is that the "ambiguous use of 'init'" error.

AliSoftware:
Ok so the problem might be that `ClassB` has two `enum TypedStoryboardSegue` nested types (one coming from parent class, one from child). I might remember a radar in bugs.swift.org about an issue about this…

What if the user just calls `ClassB.TypedStoryboardSegue(segue)` instead to disambiguate? Does that work? If it does that's at least a solution until we can find a better one.

I think the problem doesn't come from the fact that both ClassA and SubClassB both have segues with the same name. I'm guessing that even if the Segues for SubClassB were named S11, S12 and 13 instead, the call would still be ambiguous, because that's not the content of the enum that makes it ambiguous, but the fact that there are two enums with the same name `TypedStoryboardSegue`, one from parent, one from child.

AliSoftware:
I think this is the bug report I was thinking about: https://bugs.swift.org/browse/SR-3492

And I agree, in fact that code we generate should probably not compile, as it should probably not be allowed to declare a `enum TypedStoryboardSegue` nested type both in the parent class and redeclaring a different one in the inherited class.

(And if we consider that this shouldn't even compile in the first place, we need to find an alternative solution for our SwiftGen Segue template indeed then)

djbe:
Yeah, it's the `init` that fails, not the cases in the `switch`. Changing it to `ClassB.TypedStoryboardSegue` doesn't help, still ambiguous.

We could prefix the `TypedStoryboardSegue` type with the class name, and only do this if some parameter is provided by the user (`prefixTypedSegueWithType` or something).

AliSoftware:
Yeah might be the best option for now.

How often do you think this kind of situation (both a class and its subclass in the same storyboard and both having segues) would happen in a typical project which use segues and not coordinator?

djbe:
Keep in mind that this PR aggregates all segues over all storyboards for each type.

Honestly? No idea. But as we have no information on what the class structure for a user may be, and people can do whatever they want, we better be prepared. We may even go so far as to define a parameter with all classes that may conflict, so that a user can fill this in if needed.

AliSoftware:
Well sure we collect all segues over all storyboards but how many of those use the parent class? I'm gonna hypothesise that if you have a parent class with multiple subclasses, you maybe don't use the parent class directly in your SB but mainly the subclasses…

But yeah ok whatever the odds that might still happen anyway.
Can we add user info (on the right panel in Xcode) on a scene when we select it in IB? That we could parse via SwiftGen (would be a nice thing to expose in the context anyway in case people want arbitrary customization) and document that if we define a custom key named e.g. `SwiftGenSegueType` we use that for the enum name?
Well could cause issues as well because if there are multiple scenes with class ClassAViewController we don't know which one should serve as reference… 🤔

Another idea would be to offer to suffix the enum name with the level/depth of inheritance so that the parent would have `StoryboardTypedSegue1` and child would have `StoryboardTypedSegue2`…?

AliSoftware:
Btw not sure if it's adjust the case in the PR but we need a test on a storyboard containing multiple scenes that all use the same UIVC parent class but have different segues starting from them. Expected output would probably be that the generated enum aggregates all the segues from all the scenes in the single enum generated in the class.

djbe:
The UserInfo thing's a good idea, as it's the only one we have access to. We actually have no idea what the level of inheritance is, that's in the swift code. (btw, that separation is really annoying for this PR 😄 )

Obviously we need more tests, to test the:
- class with the module placeholder or not
- aggregation of segues
- multiple destination types for a segue --> fallback to `UIViewController`?
- load UserInfo data --> if multiple, only use the first/last or throw an error?

domasn:
Hi all 👋 

@djbe thank you for this PR! I think it would be really valuable to a lot of users. I see that you add new commits from time to time and wonder what's the state of this feature?

It there anything I can do to help? 🙂

PR #371, opened 2017-12-20
Strings: Allow directory as input

### LINKED ISSUE #32: Allow `swiftgen strings` to use a directory as input

pinoy4:
I have added the option to specify a directory as an input for the strings command. It searches recursively in the directory for files.  
closes #32


AliSoftware:
Interesting feature, thanks!

Might be worth checking if the documentation doesn't need updating (`README.md` + `Documentation/*` which maybe at some point tell that you can only specify a file path and would thus need to be updated to tell you can now use a directory path?)

Also, I wonder if it's worth adding in the documentation to be careful not to use a parent directory containing multiple `*.lproj` when using this new feature. Like if the user has a `Resources/` directory containing `en.lproj`, `Base.lproj`, `fr.lproj`, `es.lproj`… then passing `Resources` as the directory path to parse will lead to SwiftGen parsing the same keys multiple times (because typically each `Localizable.strings` file inside each `*.lproj` directory will list the same keys) and would make it try to create conflicting constants in the generated code.

So maybe a reminder in the doc that the user is only supposed to pass a directory containing only one `*.lproj`, or more likely directly a path to one `*.lproj` directory only.
We could even think about adding a test in the code to print a warning if the directory passed as a parameter does not end with `*.lproj`, or if the `children()` of that directory contains multiple different `*.lproj`, to warn the user that it's likely not the directory they should pass as parameter…

Sources/SwiftGenKit/Parsers/StringsParser.swift
```diff
@@ -37,23 +37,38 @@ public final class StringsParser: Parser {
...
     }
+  }
 
-    tables[name] = try dict.map { key, translation in
-      try Entry(key: key, translation: translation)
+  private func isStringsFile(name: String) -> Bool {
+    guard let fileExtension = NSURL(fileURLWithPath: name).pathExtension else {
+      return false
     }
+    return fileExtension == "strings"
```
AliSoftware: I think that whole code can be simplified to `return NSURL(fileURLWithPath: name).pathExtension == "strings"`?

(as I think `==` works between a `String?` and a `String` and already returns `false` if the optional string is `nil`)

pinoy4:
Great feedback, thank you.
I agree with your suggestions. As soon as I have a bit more time I'll implement them.

Sources/SwiftGenKit/Parsers/StringsParser.swift
pinoy4: will do

CHANGELOG.md
```diff
@@ -32,6 +32,9 @@ _None_
   [SwiftGen/templates#84](https://github.com/SwiftGen/templates/pull/84)
   [Txai Wieser](https://github.com/txaiwieser)
   [SwiftGen/templates#81](https://github.com/SwiftGen/templates/pull/81)
+* Allow directory to be passed as input of strings parser.
```
AliSoftware: Please add 2 spaces after the `.`. This is a markdown formatting catch: adding two spaces at the end of a line allows a newline within the same paragraph, so that the next line in your `.md` here will indeed be rendered in a different line once the markdown is rendered, and not the continuation of the current line. That's a strange gotcha of markdown, but that's how it works ;-)

pinoy4: humm...I was sure I had done this...thanks

AliSoftware: Yeah some editors automatically trim trailing spaces so auto-revert those

pinoy4:
@AliSoftware I have pushed the changes.  
I don't know how (or if I should) somehow link this PR to the issue https://github.com/SwiftGen/SwiftGen/issues/32

AliSoftware:
Thanks, I'll take a look when I have some free time 👍

You can edit your PR's description to add a mention like "fixes #32" or "closes #32". GitHub will detect that keyword and automatically close the related Issue once this PR gets merged 👌

Documentation/SwiftGenKit Contexts/Strings.md
```diff
@@ -2,7 +2,9 @@
 
 ## Input
 
-The strings parser accepts a `strings` file, typically `Localizable.strings`. It will parse each string in this file, including the type information for formatting parameters. 
+The strings parser accepts a `strings` file, typically `Localizable.strings`, or a directory.  
+In case it is passed a file it will parse each string in this file, including the type information for formatting parameters.  
+In the case of a directory it will look recursively in the directory and parse each `strings` file it finds. Bear in mind that only one language can be used to generate the `enum` so do not pass a directory that includes multiple localizations of the same `strings` file.
```
AliSoftware: s/can be used to generate the enum/should be passed (to avoid parsing the same keys multiple times in different languages)/

s/multiple localizations of/multiple `.lproj` folders with/

pinoy4:
Great...I think I failed setting the closes #32 thing though ahah

README.md
```diff
@@ -433,13 +433,14 @@ override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
 ## Strings
 
 ```sh
-swiftgen strings -t structured-swift3 /path/to/Localizable.strings
+swiftgen strings -t structured-swift3 (/path/to/Localizable.strings | /path/to/dir)
```
AliSoftware: I think that notation might be confusing to people not familiar with it, risking to make some people actually type the pipe verbatim.

Instead I think it's clearer to split that in 2 different example command, like:

```sh
swiftgen strings -t structured-swift3 /path/to/Localizable.strings
# OR
swiftgen strings -t structured-swift4 /path/to/SomeLang.lproj
```

README.md
```diff
@@ -478,6 +479,8 @@ enum L10n {
 ```
 </details>
 
+</details>
+
```
AliSoftware: Duplicate closing `</ details>` tag here?

Sources/SwiftGenKit/Parsers/StringsParser.swift
```diff
@@ -37,23 +40,48 @@ public final class StringsParser: Parser {
...
+    if path.isDirectory {
 
-    guard tables[name] == nil else {
-      throw StringsParserError.duplicateTable(name: name)
-    }
-    guard let data = try? path.read() else {
-      throw StringsParserError.failureOnLoading(path: path.string)
-    }
+      if try hasMultipleLprojSubdirectories(path: path) {
+        throw StringsParserError.multipleLprojDirectories
```
AliSoftware: I'm wondering if this really needs to be a terminating error? Maybe in some special situations in some projects that could be valid (e.g. `cn.lproj` having all the keys for the strings used in the Chinese version of the app but `en.lproj` having only some keys that were not in `cn.lproj` for features only available to English speaking countries? Idk… that seems unlikely for sure but I've seen so many strange setups 😅

So maybe instead just use `log(.warning, "the message here")` (or whatever I called that global method I use to print warnings in the console I'm not sure if the name I used and not handy to check from phone 😉)

Sources/SwiftGenKit/Parsers/StringsParser.swift
```diff
@@ -37,23 +40,48 @@ public final class StringsParser: Parser {
...
     }
+  }
 
-    tables[name] = try dict.map { key, translation in
-      try Entry(key: key, translation: translation)
+  private func hasMultipleLprojSubdirectories(path: Path) throws -> Bool {
+    var numberOfLprojDirectories = 0
+    for childPath in try path.children() {
+      numberOfLprojDirectories += childPath.lastComponent.hasSuffix(".lproj") ? 1 : 0
     }
```
AliSoftware: If you really wanted to count the total number of `.lproj` dirs you could have used `path.children.count(where: $0.lastComponent.hasSuffiz(".lproj")` directly instead of a for loop.

But:
1) It's not worth to continue counting as soon as you see more than one, so maybe instead of counting then check that count>1, you could exit the for loop early
2) this code doesn't seem to check subdirectories recursively. I can totally see people passing a path like `$PROJECT_DIR` by mistake but the localization files be in `$PROJECT_DIR/MyTarget/Resources/*.lproj` for example

AliSoftware (changes requested):

README.md
pinoy4: sorry...it slipped in the commit

README.md
pinoy4: sure thing

Sources/SwiftGenKit/Parsers/StringsParser.swift
pinoy4: ok, I agree

AliSoftware:
I think the way you write it should do it. You'll only see it in action once we merge the current PR in master anyway.

https://help.github.com/articles/closing-issues-using-keywords/

Sources/SwiftGenKit/Parsers/StringsParser.swift
pinoy4: I did not intend to look for multiple `.lproj` subdirectories recursively, I intended to check for multiple `.lproj` subdirectories in one directory but I see your point.  
Just to be sure, you are suggesting to check recursively whether in the given `/path/to/directory` there are multiple `.lproj` subdirs so that the following example is "invalid"?  

`swiftgen strings -t structured-swift4 /path/to/directory`
```
- directory
  - pathA
    - en.lproj
  - pathB
    - es.lproj
```

AliSoftware: Mmmh that's a good question, as actually the use case you showed might be valid of pathA and pathB are directories corresponding to different targets (day different frameworks) in the same project…

(That's also why this would just be a warning and not an error, we want to prevent user mistakes but not block valid cases that we misinterpret as mistakes)

In your example though, even if pathA and pathB for example correspond to separate targets in the project then it's unlikely that the user will want to merge the keys of both targets into one giant constants output file, as some keys will be in one target but not the other anyway so they'll need two separate output files too…

So yeah I'd say in the case you presented we still might want to warn.

What would be even better / less error prone / more solid verification could be to check if we only have at most one language (e.g. having both `dir/pathA/en.lproj` and `dir/pathB/en.lproj` would be ok because overall we only have `en.lproj` so one language ; but your example mixing `en.lproj` and `es.lproj` is more likely to be a user mistake we could warn about

Documentation/SwiftGenKit Contexts/Strings.md
pinoy4: ok

Sources/SwiftGenKit/Parsers/StringsParser.swift
pinoy4: Sorry for the extreeeeemely long delay :) ...I finally got to implementing the changes we discussed about

Sources/SwiftGenKit/Parsers/StringsParser.swift
```diff
@@ -37,23 +37,58 @@ public final class StringsParser: Parser {
...
 
-    guard tables[name] == nil else {
-      throw StringsParserError.duplicateTable(name: name)
-    }
-    guard let data = try? path.read() else {
-      throw StringsParserError.failureOnLoading(path: path.string)
-    }
+      if try hasMultipleLprojSubdirsWithDifferentLanguages(path: path) {
+        warningHandler?("Multiple .lproj directories found with different languages." +
+          "Duplicate keys will be parsed multiple times.", #file, #line)
```
AliSoftware: `#file` and `#line` are not necessary here, they are already passed by the magic of default values for those parameters being resolved at call site 😉

Sources/SwiftGenKit/Parsers/StringsParser.swift
```diff
@@ -37,23 +37,58 @@ public final class StringsParser: Parser {
...
+      }
     }
+  }
 
-    tables[name] = try dict.map { key, translation in
-      try Entry(key: key, translation: translation)
+  private func hasMultipleLprojSubdirsWithDifferentLanguages(path: Path) throws -> Bool {
+    var lprojSubdir: String?
+    for childPath in try path.children() {
+      if childPath.lastComponent.hasSuffix(".lproj") {
```
AliSoftware: `childPath` is already a `Path` — which is a type coming from PathKit to ease manipulation of file paths.
So instead of using `lastComponent` to get a `String` and test for `hasSuffix`, it would be simpler and better to directly use `childPath.pathExtension` for the test :wink:

Also, I try to avoid magic strings. It would be a good idea to create constants for those path extensions we test, and check everywhere in the code if we use `"lproj"` and `"strings"` to replace them with those constants

```swift
enum FileExtensions {
  static let lproj = "lproj"
  static let strings = "strings"
}
```

Sources/SwiftGenKit/Parsers/StringsParser.swift
```diff
@@ -37,23 +37,58 @@ public final class StringsParser: Parser {
...
-    guard let data = try? path.read() else {
-      throw StringsParserError.failureOnLoading(path: path.string)
-    }
+      if try hasMultipleLprojSubdirsWithDifferentLanguages(path: path) {
+        warningHandler?("Multiple .lproj directories found with different languages." +
+          "Duplicate keys will be parsed multiple times.", #file, #line)
+      }
+
+      try _ = path.children().map { childPath in
+        if childPath.isDirectory || isStringsFile(name: childPath.lastComponent) {
```
AliSoftware: `childPath` is already a `Path` (and not a String) here too. So instead of extract a string (`.lastComponent`) from that path only to later (line 90) construct an `NSURL(fileURLWithPath: name).pathExtension`, it's probably simpler to just pass `isStringsFile(path: childPath)`, and in the `func isStringsFile` implementation directly test `path.pathExtension == FileExtensions.strings`

AliSoftware (changes requested):
No need to apologise for the late update, I myself haven't found time to work on SwiftGen in month, it's cool already that you contribute and come back to make things progress 👍 

Couple of notes from reading the code from GitHub. Will have to find time to test/run it on a Mac at some point before validation :wink:

djbe:
This has been implemented by #570, released in SwiftGen 6.1.0 🎉

PR #377, opened 2018-01-15
Add a new `custom` command

pbernery:
This patch adds a new `custom` in SwiftGen. It generates code given a template file. This can be used to generate code based on environment variables for example.

## More explanation
I work on apps that needs to be configured in many ways at build time. The common thing is to set the base URL for Web service calls.

Xcode Configurations are not the right answer in this case as there are multiple configuration variables. Xcode Configurations is not meant to combine several configurations, but only to select a fixed set of settings.

Code generation is one answer, and SwiftGen works well for this.

Here is an example of what `swiftgen custom` allows:
```swift
struct Config {
  let baseUrl = "{{ env.BASE_URL }}"
  let envCode = {{ env.ENV_CODE }}
  let analyticsAPIKey = "{{ env.ANALYTICS_API_KEY }}"
}
```

returns, with the appropriate env. variables set:

```swift
struct Config {
  let baseUrl = "https://www.my-api.org"
  let envCode = 123
  let analyticsAPIKey = "myAPiKeY"
}
```

Note that SwiftGen can already do this kind of code generation by simply calling any parser command with --templatePath.  For instance:

    swiftgen colors --templatePath <template>

This patch just tries to given more semantic of what is happening by adding the `custom` command (even though I may admit this name may not be precise enough). The previous command just becomes:

    swiftgen custom --templatePath <template>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,8 @@ _None_
 
 ### New Features
 
-_None_
+* Add a new `custom` command
+  [Philippe Bernery](https://github.com/pbernery)
 
 ### Internal Changes
 
--- a/Sources/SwiftGen/Commander/CustomCommands.swift
+++ b/Sources/SwiftGen/Commander/CustomCommands.swift
@@ -1,0 +1,41 @@
+//
+//  CustomCommands.swift
+//  swiftgen
+//
+//  Created by Philippe Bernery on 15/01/2018.
+//  Copyright © 2018 Backelite. All rights reserved.
+//
+
+import Commander
+import PathKit
+import StencilSwiftKit
+import SwiftGenKit
+
+private let templatePathOption = Option<String>(
+  "templatePath", default: "", flag: "p",
+  description: "The path of the template to use for code generation."
+)
+
+private let paramsOption = VariadicOption<String>(
+  "param", default: [],
+  description: "List of template parameters"
+)
+
+let customCommand = Commander.command(
+  outputOption,
+  templatePathOption,
+  paramsOption
+) { output, templatePath, parameters in
+  try ErrorPrettifier.execute {
+    let templateRef = try TemplateRef(templateShortName: "",
+                                      templateFullPath: templatePath)
+    let templateRealPath = try templateRef.resolvePath()
+
+    let template = try StencilSwiftTemplate(templateString: templateRealPath.read(),
+                                            environment: stencilSwiftEnvironment())
+
+    let enriched = try StencilContext.enrich(context: [:], parameters: parameters)
+    let rendered = try template.render(enriched)
+    try output.write(content: rendered, onlyIfChanged: true)
+  }
+}
--- a/Sources/SwiftGen/TemplateRef.swift
+++ b/Sources/SwiftGen/TemplateRef.swift
@@ -45,9 +45,12 @@ enum TemplateRef {
   /// - Returns: The Path matching the template found
   /// - Throws: TemplateRef.Error
   ///
-  func resolvePath(forSubcommand subCmd: String) throws -> Path {
+  func resolvePath(forSubcommand subCmd: String? = nil) throws -> Path {
     switch self {
     case .name(let templateShortName):
+      guard let subCmd = subCmd else {
+        throw TemplateRef.Error.noTemplateProvided
+      }
       var path = appSupportTemplatesPath + subCmd + "\(templateShortName).stencil"
       if !path.isFile {
         path = bundledTemplatesPath + subCmd + "\(templateShortName).stencil"
--- a/Sources/SwiftGen/main.swift
+++ b/Sources/SwiftGen/main.swift
@@ -42,6 +42,8 @@ let main = Group {
   for cmd in allParserCommands {
     $0.addCommand(cmd.name, cmd.description, cmd.command())
   }
+
+  $0.addCommand("custom", "generate code using a given template file", customCommand)
 }
 
 let version = Bundle.main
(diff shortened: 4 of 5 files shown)


CHANGELOG.md
```diff
@@ -14,7 +14,8 @@ _None_
 
 ### New Features
 
-_None_
+* Add a new `custom` command
```
djbe: Use the same format as the other entries, namely:
- End your sentence with a `.` and 2 spaces
- Author (you have this)
- PR

Sources/SwiftGen/Commander/CustomCommands.swift
```diff
@@ -0,0 +1,41 @@
...
+//  Created by Philippe Bernery on 15/01/2018.
+//  Copyright © 2018 Backelite. All rights reserved.
+//
+
+import Commander
+import PathKit
+import StencilSwiftKit
+import SwiftGenKit
+
+private let templatePathOption = Option<String>(
```
djbe: Don't repeat yourself, make the `templatePathOption` in `ParserCLICommands.swift` internal instead of private, and maybe move it to a separate file.

Sources/SwiftGen/Commander/CustomCommands.swift
```diff
@@ -0,0 +1,41 @@
...
+  "templatePath", default: "", flag: "p",
+  description: "The path of the template to use for code generation."
+)
+
+private let paramsOption = VariadicOption<String>(
```
djbe: Same thing, use the existing option.

Sources/SwiftGen/Commander/CustomCommands.swift
```diff
@@ -0,0 +1,41 @@
...
+  "param", default: [],
+  description: "List of template parameters"
+)
+
+let customCommand = Commander.command(
```
djbe: Maybe try to use the existing framework for parser commands? I know you don't have a resource "path" parameter, but maybe create a more generic protocol (or struct) for commands with no resources. @AliSoftware?

Sources/SwiftGen/TemplateRef.swift
```diff
@@ -45,9 +45,12 @@ enum TemplateRef {
   /// - Returns: The Path matching the template found
   /// - Throws: TemplateRef.Error
   ///
-  func resolvePath(forSubcommand subCmd: String) throws -> Path {
+  func resolvePath(forSubcommand subCmd: String? = nil) throws -> Path {
```
djbe: There's no need for this. You do have a subcommand ("custom" for now).

Sources/SwiftGen/Commander/CustomCommands.swift
```diff
@@ -0,0 +1,41 @@
...
+  outputOption,
+  templatePathOption,
+  paramsOption
+) { output, templatePath, parameters in
+  try ErrorPrettifier.execute {
+    let templateRef = try TemplateRef(templateShortName: "",
+                                      templateFullPath: templatePath)
+    let templateRealPath = try templateRef.resolvePath()
```
djbe: Just provide your subcommand, you have one anyway. And don't set the template name to an empty string, we might provide a built-in template at some point.

Sources/SwiftGen/Commander/CustomCommands.swift
AliSoftware: Never thought we could have subcommands not taking any path, but yeah would be nicer to have all subcommand code build on the same type of similar code.

AliSoftware:
Side note & shell trick: You can transform the output of any command to a pseudo-file using `<(command)`. For example when you use `diff` you're expected to provide two files as parameter to diff, but if instead of a real file you want to use the output of another command, say `swiftgen templates cat xcassets swift3`, you can wrap that command inside `<()`. So printing the diff of the two outputs `swiftgen templates cat xcassets swift3` and `swiftgen templates cat xcassets swift4` for example can be done using:

`diff <(swiftgen templates cat xcassets swift3) <(swiftgen templates cat xcassets swift4)`

This means that, if in the future we have a `swiftgen json` command expecting a JSON file as a parameter, but that in your case you don't have a real JSON file, I guess that you can dynamically simulate that you have one using something like `<(echo {})` as the path parameter, to make it like you have a real file whose content is the empty `{}` JSON… could be an acceptable solution to do what you want once we have the more generic subcommand for YAML / JSON

CHANGELOG.md
pbernery: Alright, thanks.

Sources/SwiftGen/Commander/CustomCommands.swift
pbernery: Indeed

Sources/SwiftGen/Commander/CustomCommands.swift
pbernery: Ok. I'll change it depending on how this PR may be accepted or not. See discussion follow-up below.

pbernery:
Agreed.

Moreover, this `custom` subcommand does not have a real added value. I just need a way to call the Stencil engine, through SwiftGen as it configures it to accept env. variables.

As I mentioned in my PR description, all SwiftGen subcommands are already capable of injecting env. variables. Example:

    swiftgen colors --templatePath <template>

works as well. I could just rely on this behavior, but calling the `colors` subcommand feels kind of strange for this purpose, although it technically works.
Also, this is a non-documented behavior, it means it may break in a future SwiftGen version as it is not an expected usage of the subcommand.

@AliSoftware, your proposal of using the future `json` subcommand could indeed be an acceptable solution, but again this would rely on an internal behavior of the subcommand. But documenting this behavior could guarantee its availability in future versions.

---

What about having only *action* subcommands instead of subcommands that mixes *actions* and *parser selection*? Instead of having a `xcassets`, `colors`, `strings`, `storyboards`, `fonts` and `json`, `yaml` subcommands, what about having a more generic `generate` subcommand, that take parameters.

The `generate` subcommand could take already existing parameters (`--template`, `--templatePath`) + a new `--parser` parameter that would take a parser name, responsible of injecting data in the Stencil context, usable in the template.

    swiftgen generate --parser xcassets --template xcassets-swift5 <PATH> 

That changes how SwiftGen is called, but I believe is more explicit on what the command is actually doing, meaning calling a parser that build a Stencil context, usable by a Stencil template.
Also, this may open new ways of using SwiftGen. What about letting users write their own private parsers, loaded and discovered dynamically by SwiftGen?

What are your thoughts?

djbe:
This is actually how we envision a future SwiftGen, where it could accept context data from for example another application via for example `pipes`. It would go even further and could have a flow such as `swiftgen parse something | my-command-that-modifies-context | swiftgen generate something`.

In retrospective, I'm not sure if there's any added value in adding a `custom` command now that we would have to obsolete in a not too remote future.

djbe:
Closing this for now, as #379 added JSON/Plist/YAML support, and you can use @AliSoftware's shell trick to simulate an empty file (with `<(echo {})`).

For now, we consider this use case too limited, but we can revisit this at some later point if need be.

Issue #376, opened 2018-01-12
Generate ObjC source for .strings files.


mr-fixit:
I wrote a couple of templates, and would like to contribute them back.  I'm opening this issue as suggested by `CONTRIBUTING.md`.

Given the following in `Localizable.strings`
```"cmd.delete" = "Delete";```

... the following is generated:

```
// Localizable.h
@import Foundation;
NS_ASSUME_NONNULL_BEGIN

@interface Localized : NSObject
  + (NSString*)cmdDelete; // cmd.delete --> "Delete"
@end
```

and 

```
// Localizable.m
#import "Localizable.h"

#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wformat-security"

NSString* tr(NSString* key, ...) {
    va_list args;
    va_start(args, key);

    NSString *format = [NSBundle.mainBundle localizedStringForKey:key value:@"" table:nil];
    NSString *result = [[NSString alloc] initWithFormat:format arguments:args];
    return result;
};

@implementation Localized : NSObject
+ (NSString*)cmdDelete {  
    return tr(@"cmd.delete"); // Delete
}
@end```

AliSoftware:
👍 looks like a nice interface too me 👌

Iirc I think there's a way to annotate functions like `tr` to tell the compiler they behave like `printf`, would be nice to use that annotation here, I'll try to find how it looks

AliSoftware:
Ha, here it is: [NSHipster article](http://nshipster.com/__attribute__/)

Which makes me realize that you can forget about what I said, that attribute won't be useful in the current case, because it's used by the compiler to check if the variadic parameters passed match the placeholders in the "format" parameter… but in the case of `tr` we don't pass a format parameter with placeholders, we pass a key… so that attribute wouldn't make sense here, nvm.

djbe:
What we can still do is generate functions with the correct signatures for parameter types. I have no idea what to name the parameters though? p1, p2, pN?
```objective-c
@interface Localized: NSObject
  + (NSString*)cmdDeleteWithP1: (NSObject *)a1 p2: (float)a2;
@end
```

As @AliSoftware said, you can just use varargs, but will have no way of checking the arguments at compile time.

AliSoftware:
Fun fact: in Objective-C, parameters are not _required_ to have a name.

You can totally do `+ (NSString*)cmdDelete: (id)p1, :(float)p2, :(int)p3;`
[...]

PR #378, opened 2018-01-17
Strings: Obj-C templates

mr-fixit:
This PR addresses https://github.com/SwiftGen/SwiftGen/issues/376


AliSoftware:
> not to the issue, cuz no one else did that

Tbh that's probably because, unlike you, no one else read the CONTRIBUTING.md 😜 (and fine by me if you wish to add the link to the issue in the CHANGELOG)

templates/strings/objc-h.stencil
```diff
@@ -0,0 +1,50 @@
...
+  {% for type in types %}
+    p{{forloop.counter}}{% if not forloop.last %}, {% endif %}
+  {% endfor %}
+{% endfilter %}{% endmacro %}
+{% macro paramTranslate swiftType %}
+  {% if swiftType == "Int" %}
+    NSInteger
+  {% elif swiftType == "String" %}
+    NSString*
+  {% else %}
```
AliSoftware: No case for `Float` (or `Double`, I don't remember which one we use)?

templates/strings/objc-h.stencil
```diff
@@ -0,0 +1,50 @@
...
+    p{{forloop.counter}}{% if not forloop.last %}, {% endif %}
+  {% endfor %}
+{% endfilter %}{% endmacro %}
+{% macro paramTranslate swiftType %}
+  {% if swiftType == "Int" %}
+    NSInteger
+  {% elif swiftType == "String" %}
+    NSString*
+  {% else %} 
+    NSObject*
```
AliSoftware: imho `id` would be more accurate

templates/strings/objc-h.stencil
```diff
@@ -0,0 +1,50 @@
...
+  {% endif %}
+{% endmacro %}
+{% macro recursiveBlock table item sp %}
+{% for string in item.strings %}
+{% if string.types %}
++ (NSString*){{string.key|swiftIdentifier:"pretty"|lowerFirstWord}}:{% call parametersBlock string.types %};{% if not param.noComments %}  // {{string.translation}}{% endif %}
+{% else %}
++ (NSString*){{string.key|swiftIdentifier:"pretty"|lowerFirstWord}};{% if not param.noComments %} // {{string.key}} --> "{{string.translation}}"{% endif %}
```
AliSoftware: Not sure why you have differently formatted comments for those two cases (one with `// key --> value` here but just `// value` above). That's just a style point and my OCD talking but ¯\\\_(ツ)\_/¯

templates/strings/objc-m.stencil
```diff
@@ -0,0 +1,64 @@
...
+
+NSString* tr(NSString* key, ...) {
+    va_list args;
+    va_start(args, key);
+
+    NSString *format = [NSBundle.mainBundle localizedStringForKey:key value:@"" table:nil];
+    NSString *result = [[NSString alloc] initWithFormat:format arguments:args];
+    return result;
+};
+
```
AliSoftware, line 28: You probably want a `#pragma clang diagnostic pop` around here?

templates/strings/objc-h.stencil
mr-fixit: I found the list of types in `StringParser.swift`, `public enum PlaceholderType`, they are String (%@), Float (%[a,e,f,g]), Int (%[d,i,o,u,x]), Character (%c), UnsafePointer<unichar> (%s), UnsafePointer<Void> (%p), and UnsafePointer<()> (not used)".

templates/strings/objc-m.stencil
mr-fixit: agreed

templates/strings/objc-h.stencil
mr-fixit: I'm sorely tempted to extend the code in StringParser.swift, creating a `typesObjC` attribute that would be an array of `ObjCPlaceholderType`.  But for now, I think I'll just extend the `paramTranslate` macro in the `.stencil` files.

amarcadet:
Hi, when are you hoping to merge this PR? Has it been dropped?

mr-fixit:
I'm pinged.  I got diverted and was satisfied using the work on my own branch.  I'll push it through this week.

templates/strings/objc-h.stencil
mr-fixit: I added a case for `Float`.

templates/strings/objc-h.stencil
mr-fixit: Are you referring to line 22 (`NSString*`) or line 24 (`NSObject*`)?

If the latter, then I changed the output for all the things I don't handle to `FIX_OBJC_STRING_DOT_STENCIL_PLEASE`.  That should make it easier for the next guy to find where the fix goes.

templates/strings/objc-h.stencil
mr-fixit: good catch!  I'll put `key --> value` in both places.

mr-fixit:
Dear @AliSoftware & @djbe, what's the next step here?  Is the ball in my court or someone else's?

atmamont:
It's quite a long time from the latest actions. Is there any chance this will be reviewed again and merged soon?

mr-fixit:
Concidentally, I asked the same question yesterday.  Oliver suggested I ping @djbe.

djbe:
Right, let's get this ball rolling:
- Needs a rebase on `develop`. A few things have changed:
  - location of generated output in subdirectories per template
  - compilation configuration using `compilation-configuration.yml` files, which doesn't support Obj-C yet. Not sure if you want to add this?
- After that rebase, change the header to `// Generated using SwiftGen — https://github.com/SwiftGen/SwiftGen` (all templates now have the same header)
- You'll need to write documentation for the template, with the right `params` it supports, etc...
- Might be best to provide a `config` example in that documentation, as users need to generate 2 outputs (`.h` and `.m`).

I'll review the templates themselves after the rebase on develop, otherwise they may get lost / point to the wrong lines (maybe?).

djbe:
Quick template feedback:
- Ensure the indentation is consistent everywhere.
- Why are the comments inline? It's probably easier to have them before each constant/function (and we can use doc comments syntax)
- Use `[NSBundle bundleForClass:]`
- Handle the `CChar`, `UnsafePointer<CChar>` and `UnsafeRawPointer` parameter types
- Why repeat the comments with the translation in the implementation file? The user will probably only ever check the header.

mr-fixit:
Thanks, @djbe, I'm getting started on your suggestions.  

Is there something I need to do vis-a-vis "location of generated output in subdirectories per template"?

I'd like to add support for `compilation-configuration.yml` in a different PR.  This one really really wants to be merged, and I don't want to delay that (further).

djbe:
> Is there something I need to do vis-a-vis "location of generated output in subdirectories per template"?

Easiest thing to do would be to run the "generate output" scheme (either in xcode or via rake). You'll have to remove your old generated files manually though.

> I'd like to add support for `compilation-configuration.yml` in a different PR. This one really really wants to be merged, and I don't want to delay that (further).

What do you mean? We can't merge this PR as all the points I mentioned above still need to be addressed. If you need these templates in your own project right now, you can always add them to your project as custom templates, and use the `templatePath` configuration option.

mr-fixit:
What I meant was in response to your question "Not sure if you want to add this?".  That made it sound optional.

djbe:
Right, I forgot to scoll up enough 🤦‍♂️ Yeah, obj-c compilation can be done in a separate PR.

CHANGELOG.md
```diff
@@ -331,6 +331,9 @@ Read the [SwiftGen 6.0 Migration Guide](Documentation/MigrationGuide.md#swiftgen
   [SwiftGen/templates#84](https://github.com/SwiftGen/templates/pull/84)
   [Txai Wieser](https://github.com/txaiwieser)
   [SwiftGen/templates#81](https://github.com/SwiftGen/templates/pull/81)
+* Templates to generate ObjC source for .strings
```
  <tbody>
    <tr>
      <td>:warning:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/SwiftGen/SwiftGen/blob/1ce32c1af15ffe7e28fb679a5978adff193b316d/CHANGELOG.md#L334"></span>Line describing your entry should end with a period and 2 spaces.</td>
    </tr>
  </tbody>
</table>

SwiftGen-Eve:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

---

</p>

mr-fixit:
@djbe @AliSoftware I've just pushed the last commit needed to address all of David's feedback.  The ball is in your court. Thanks again, and in advance!

Tests/TemplatesTests/TestsHelper.swift
```diff
@@ -182,7 +182,15 @@ extension XCTestCase {
       }
 
       for (index, (context: context, suffix: suffix)) in variations.enumerated() {
-        let outputFile = "\(contextName)\(suffix).swift"
+        let outputSuffix = { () -> String in
```
djbe: This isn't ideal. Maybe the function should accept an `extension` parameter? (with a default value of `swift`)

templates/strings/objc-m.stencil
```diff
@@ -0,0 +1,80 @@
+// Generated using SwiftGen — https://github.com/SwiftGen/SwiftGen
+
+{% if tables.count > 0 %}
+
+#import "Localizable.h"
+
+#pragma clang diagnostic push
```
djbe: I think it's best to have these push/pop only around the `tr` function.

templates/strings/objc-m.stencil
```diff
@@ -0,0 +1,80 @@
+// Generated using SwiftGen — https://github.com/SwiftGen/SwiftGen
+
+{% if tables.count > 0 %}
+
+#import "Localizable.h"
```
djbe: This should be parameterizable. For example: `{{ param.headerName|default:"Localizable.h" }}`
Not sure if it should contain the `.h` extension or not though 🤷‍♂️ 

Don't forget to add the documentation in the template docs for this, and a test variation to test the import.

templates/strings/objc-m.stencil
```diff
@@ -0,0 +1,80 @@
...
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wformat-security"
+
+@interface BundleToken : NSObject
+@end
+
+@implementation BundleToken
+@end
+
+NSString* tr(NSString* tableName, NSString* key, ...) {
```
djbe: Can a free standing function like this produce issues (conflicts)? We may want to embed it in a type (interface) or something.

templates/strings/objc-m.stencil
```diff
@@ -0,0 +1,80 @@
...
+  {% elif swiftType == "Int" %}
+    NSInteger
+  {% elif swiftType == "CChar" %}
+    char
+  {% elif swiftType == "String" %}
+    NSString*
+  {% elif swiftType == "UnsafePointer<CChar>" %}
+    char*
+  {% elif swiftType == "UnsafeRawPointer" %}
+    void*
```
djbe, line 53: `id` maybe?

templates/strings/objc-h.stencil
```diff
@@ -0,0 +1,60 @@
...
+  {% elif swiftType == "UnsafeRawPointer" %}
+    void*
+  {% else %}
+    objc-h.stencil is missing '{{swiftType}}'
+  {% endif %}
+{% endmacro %}
+{% macro recursiveBlock table item sp %}
+{% for string in item.strings %}
+{% if not param.noComments %}
+// {{string.key}} --> "{{string.translation}}"
```
djbe, line 40: I don't think we have the `key` in the other templates, maybe remove it?

templates/strings/objc-m.stencil
```diff
@@ -0,0 +1,80 @@
...
+  {% else %}
+    objc-m.stencil is missing '{{swiftType}}'
+  {% endif %}
+{% endmacro %}
+{% macro recursiveTableContents table item sp %}
+{{sp}}  {% for string in item.strings %}
```
djbe: You need the `sp` variable, you have a flat template, not a recursive structured one (see `flat-swift4` for example).

templates/strings/objc-h.stencil
```diff
@@ -0,0 +1,60 @@
...
+{% for table in tables %}
+@interface {{ table.name }} : NSObject
+    {% call recursiveBlock table.name table.levels "  " %}
+@end
+
+{% endfor %}
+
+NS_ASSUME_NONNULL_END
+{% else %}
+// No string found
```
djbe: strings

templates/strings/objc-m.stencil
```diff
@@ -0,0 +1,80 @@
...
+{% for table in tables %}
+        {% set tableName %}{{table.name|default:"Localized"}}{% endset %}
+@implementation {{ tableName }} : NSObject
+        {% call recursiveTableContents table.name table.levels "  " %}
+@end
+
+{% endfor %}
+#pragma clang diagnostic pop
+{% else %}
+// No string found
```
djbe: strings

templates/strings/objc-m.stencil
```diff
@@ -0,0 +1,80 @@
...
+@end
+
+@implementation BundleToken
+@end
+
+NSString* tr(NSString* tableName, NSString* key, ...) {
+    va_list args;
+    va_start(args, key);
+
+    NSBundle *bundle = [NSBundle bundleForClass:BundleToken.class];
```
djbe, line 17: Nitpick: add a space between the `:` and the argument (here and the lines below).

templates/strings/objc-m.stencil
djbe: Nitpick: have the `*` stick to the name, not the type:
`NSString *tr(NSString *tableName, NSString *key, ...) {`

Apply the same to the generated code.

templates/strings/objc-m.stencil
```diff
@@ -0,0 +1,80 @@
+// Generated using SwiftGen — https://github.com/SwiftGen/SwiftGen
+
+{% if tables.count > 0 %}
+
```
djbe: Remove this empty line to avoid the double empty lines in the generated code.

templates/strings/objc-h.stencil
```diff
@@ -0,0 +1,60 @@
+// Generated using SwiftGen — https://github.com/SwiftGen/SwiftGen
+
+{% if tables.count > 0 %}
+@import Foundation;
```
djbe: Might be safer to use `#import <Foundation/Foundation.h>` for users that have modules disabled (Obj-C++).

Documentation/templates/strings/objc.md
```diff
@@ -0,0 +1,53 @@
+## Template Information
+
+| Name      | Description       |
+| --------- | ----------------- |
+| File names | strings/objc-string-h.stencil & strings/objc-string-m.stencil |
+| Configuration example | <pre>strings:<br />  inputs: path/to/Localizable.strings<br />  outputs:<br />    templateName: objc-string-h<br />    output: Localizable.h</pre><br />    templateName: objc-string-m<br />    output: Localizable.m</pre> |
```
djbe: There's something wrong here, also `outpus` is an array, so there should be a `- ` before each `templateName` here:
```yaml
strings:
  inputs: path/to/Localizable.strings
  outputs:
    - templateName: objc-string-h
      output: Localizable.h
    - templateName: objc-string-m
      output: Localizable.m
```

CHANGELOG.md
```diff
@@ -331,6 +331,9 @@ Read the [SwiftGen 6.0 Migration Guide](Documentation/MigrationGuide.md#swiftgen
   [SwiftGen/templates#84](https://github.com/SwiftGen/templates/pull/84)
   [Txai Wieser](https://github.com/txaiwieser)
   [SwiftGen/templates#81](https://github.com/SwiftGen/templates/pull/81)
+* Templates to generate ObjC source for .strings.
```
djbe: Add "For more information, check the template's documentation." with a link, see other entries.

Documentation/templates/strings/objc.md
```diff
@@ -0,0 +1,53 @@
...
+| Language | ObjectiveC |
```
djbe: Objective-C

Documentation/templates/strings/objc.md
```diff
@@ -0,0 +1,53 @@
...
+| Author | Eric Slosser |
+
+## When to use it
+
+- When you need to generate *ObjectiveC* code.
```
djbe: Objective-C

Documentation/templates/strings/objc.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Customization
+
+You can customize some elements of this template by overriding the following parameters when invoking `swiftgen`. See the [dedicated documentation](../../ConfigFile.md).
+
+| Parameter Name | Default Value | Description |
+| -------------- | ------------- | ----------- |
+| `noComments`   | N/A           | Setting this parameter will disable the comments describing the translation of a key. |
+
+
```
djbe, line 22: Remove the extra newline.

Tests/TemplatesTests/StringsTests.swift
```diff
@@ -33,6 +33,19 @@ class StringsTests: XCTestCase {
     ]
   }
 
+  let variationsObjC: VariationGenerator = { name, context in
+    guard name == "localizable" else { return [(context: context, suffix: "")] }
+
+    return [
+      (context: context,
+       suffix: ""),
+      (context: try StencilContext.enrich(context: context, parameters: ["enumName=XCTLoc"]),
```
djbe: `enumName` isn't used as far as I can see in the template?

djbe (changes requested):
A few minor issues and some small code style questions, should be easy to fix.

templates/strings/objc-m.stencil
mr-fixit: I'll make `tr()` a `static` function, it won't be accessible from outside the file.

I agree with the nitpick!

so, coming soon: `static tr(NSString *tableName, NSString *key, ...)`

templates/strings/objc-m.stencil
mr-fixit: I don't think so!  This is "%p", which may not point to an ObjC object.

templates/strings/objc-h.stencil
mr-fixit: I think it's essential.  Multiple keys may have the same (English) translation, so just knowing the translation isn't enough to find the string in `Localizable.strings`

templates/strings/objc-m.stencil
mr-fixit: Do you mean I "**don't** need the `sp` variable"?

templates/strings/objc-h.stencil
mr-fixit: good point, thanks.

Documentation/templates/strings/objc.md
mr-fixit: I'll add '- ' before each 'templateName', and '  ' before each 'output'.  

Is there something else?

Documentation/templates/strings/objc.md
mr-fixit: ✔️

Documentation/templates/strings/objc.md
mr-fixit: ✔️

Documentation/templates/strings/objc.md
mr-fixit: done

CHANGELOG.md
mr-fixit: I think also I should be moving this out of 'Version 5.3.0' and into ... which version?  I'll put it under 'Develop' for now.

Tests/TemplatesTests/StringsTests.swift
mr-fixit: yeah, I figured that out too.  good catch.

CHANGELOG.md
```diff
@@ -33,6 +33,9 @@
   [426](https://github.com/SwiftGen/SwiftGen/issues/426)
   [468](https://github.com/SwiftGen/SwiftGen/issues/468)
   [573](https://github.com/SwiftGen/SwiftGen/pull/573)
+* Strings: templates to generate Objective-C. Please check [the template's documentation](Documentation/templates/string/objc.md) for more information.
```
  <tbody>
    <tr>
      <td>:warning:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/SwiftGen/SwiftGen/blob/2c514756ffc950c7f241f261ce41ccadacf7ae87/CHANGELOG.md#L36"></span>Line describing your entry should end with a period and 2 spaces.</td>
    </tr>
  </tbody>
</table>

templates/strings/objc-m.stencil
mr-fixit: really?  i haven't seen that style elsewhere.

templates/strings/objc-m.stencil
djbe: Uhm, yeah 😅

templates/strings/objc-h.stencil
djbe: Right, but the key is in the function name (kinda), so this isn't really necessary.

mr-fixit: It's the "kinda" that makes me want to keep the actual key.  I'd rather not ask the user to know that to find the value returned by `+[Localizable.fooBar]`, they need to look for `foo.bar`.

templates/strings/objc-m.stencil
djbe: From taking a look at most style guides, seems I have a weird style 😆 Keep it as is for this.

templates/strings/objc-h.stencil
mr-fixit: (Oh, BTW, process question: I 'resolved' the conversations to indicate I'd answered them, is that the right way to do things?)

CHANGELOG.md
djbe: Good catch, hadn't even noticed it was in the wrong section 👍

templates/strings/objc-m.stencil
mr-fixit: roger!  I'll remove it.

templates/strings/objc-m.stencil
mr-fixit: avec pleasure!

templates/strings/objc-m.stencil
djbe: Ah now I know what I was missing. There's actually an issue with the generated context (in SwiftGenKit): it parses `%@` as `.object` (good), which is converted for the context to `String` (bad). I'll need to fix that first.

templates/strings/objc-h.stencil
djbe: I usually resolve conversations when the outcome is clear (ok code changed). This was a bit more open-ended. I was going through the whole thing again anyway 🤷‍♂️

templates/strings/objc-m.stencil
djbe: See #620 for the necessary change. So for `String` you should generate `id`.

With SwiftGen 7.0 we'll correct the context type to something more generic (like `Object`), but for now check against `String`.

mr-fixit: I agree with your #620.  I'll change my templates so that when they see either 'String' or 'Any', they emit 'id'.

For the 17th time, I regret not creating an 'objcTypes' next to 'types' (see `./Documentation/SwiftGenKit Contexts/Strings.md#35`)

mr-fixit: okay, just pushed that commit.

djbe: Nah, we would never have added a separate `objcTypes` to the context. We're going to make the content of `types` more generic in some future version, so both the obj-c and swift templates will have to convert them to more concrete types.

djbe (approved):
Allright, I think that covers my feedback. @AliSoftware you wanna do a last review? (yours dates to a while back and some things have changed)

mr-fixit:
David, thanks for your thoughtful review, and the massive amount of other work you're doing on this repo.

templates/strings/objc-m.stencil
AliSoftware: Side note: it could be cool to have a nice Stencil tag or filter for such mappings, like `{{ type | map "String:NSString,Int:NSInteger,Any:id" }}` or something similar to what we'd do with a `switch` or `Dictionary` if it were in Swift

No rush and shouldn't block that PR, but would definitively make things easier in the future for such conversions rather than having to go the `{% if type == 'String' %}NSString{% …` route :wink:

AliSoftware:
+ 💯 on the amazing work @djbe is doing on the repo 🙌 
I'll do a review of the overall PR in a minute.

Documentation/templates/strings/objc.md
```diff
@@ -0,0 +1,53 @@
...
+| Name      | Description       |
+| --------- | ----------------- |
+| File names | strings/objc-string-h.stencil & strings/objc-string-m.stencil |
+| Configuration example | <pre>strings:<br />  inputs: path/to/Localizable.strings<br />  outputs:<br />    - templateName: objc-string-h<br />      output: Localizable.h</pre><br />    - templateName: objc-string-m<br />      output: Localizable.m</pre> |
+| Language | Objective-C |
+| Author | Eric Slosser |
+
+## When to use it
+
+- When you need to generate *Objective-C* code.
```
AliSoftware: Could someone remind me when this template is more useful compared to generate Swift code compatible with ObjC (for which we have a PR too I think?) — with `@objcMembers class L10n: NSObject` instead of `enum` for namespacing so that they are visible to ObjC?

Is it for projects that only have ObjC code and no Swift at all in them yet, so you don't want to have a `Strings.swift` generated by SwiftGen to be the only Swift file and triggering the inclusion of the Swift runtime just for it?

djbe: Essentially yeah.

Note that we still need to write some sort of (compilation) test for this template, but I'd leave that for another PR.

Documentation/templates/strings/objc.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Customization
+
+You can customize some elements of this template by overriding the following parameters when invoking `swiftgen`. See the [dedicated documentation](../../ConfigFile.md).
+
+| Parameter Name | Default Value | Description |
+| -------------- | ------------- | ----------- |
+| `noComments`   | N/A           | Setting this parameter will disable the comments describing the translation of a key. |
+| `headerName`   | "Localizable" | A generated .m file `#includes "{headerName}.h"` |
```
AliSoftware: That description of that parameter doesn't feel really clear to me.

1) Is that the name of the `.h` corresponding to the generated `.m`? If so, can't we assume it's gonna be the same name for the `.h` and  `.m` files? [EDIT] Oh wait, we don't have access to the name of the generated file from the context, that's why we have to provide it explicitly that way, right?
2) Wouldn't that description be clearer worded like "The name of the ObjC header generated by the `objc-string-h` template, that will be `#include`'d in the generated `.m` file — this parameter should have the same value as the `output` file name used for the `template: objc-string-h` entry in your config file"?

Documentation/templates/strings/objc.md
AliSoftware: Alright. We should mention that in the "When to use it" paragraph then, maybe?

Documentation/templates/strings/objc.md
```diff
@@ -0,0 +1,53 @@
...
+**Extract:**
+
+```swift
+@interface Localizable : NSObject
+// alert__message --> "Some alert body there"
++ (NSString*)alertMessage;
+// alert__title --> "Title of the alert"
++ (NSString*)alertTitle;
+// ObjectOwnership --> "These are %3$@'s %1$d %2$@."
++ (NSString*)objectOwnership:(NSInteger)p1 and:(NSString*)p2 and:(NSString*)p3;
```
AliSoftware: I know that this is some artefact of ObjC, but those `and:… and:… and:…` are not great. I forgot that ugly part of ObjC I guess 😓 
An alternative could be to not give any name to those parameters, which is technically allowed in ObjC… even if not common at all (it's only encountered on like 3 or 4 functions in Apple's SDK iirc);
```objc
+ (NSString*)objectOwnership:(NSInteger)p1 :(NSString*)p2 :(NSString*)p3;
```
And it's used like this in code:
```objc
NSString* fruit = @"bananas";
NSString* name = @"Olivier";
label.text = [Localizable objectOwnership:5:fruit:name];
```

Not saying that it's better. In fact it's so uncommon that it might feel uglier than the version with `and:… and:… and:…". But just thought I'd mention that this could be another possibility / another route we could choose too. Matter of taste :wink:

Tests/Fixtures/Generated/Strings/objc-m/localizable-headerName.m
```diff
@@ -0,0 +1,71 @@
...
+#pragma clang diagnostic ignored "-Wformat-security"
+
+static NSString* tr(NSString *tableName, NSString *key, ...) {
+    va_list args;
+    va_start(args, key);
+
+    NSBundle *bundle = [NSBundle bundleForClass:BundleToken.class];
+    NSString *format = [bundle localizedStringForKey:key value:@"" table:tableName];
+    NSString *result = [[NSString alloc] initWithFormat:format arguments:args];
+    return result;
```
AliSoftware, line 24: I think we are supposed to call `va_end(args)` once we are finished with the variadic args?

templates/strings/objc-m.stencil
```diff
@@ -0,0 +1,81 @@
...
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wformat-security"
+
+static NSString* tr(NSString *tableName, NSString *key, ...) {
+    va_list args;
+    va_start(args, key);
+
+    NSBundle *bundle = [NSBundle bundleForClass:BundleToken.class];
+    NSString *format = [bundle localizedStringForKey:key value:@"" table:tableName];
+    NSString *result = [[NSString alloc] initWithFormat:format arguments:args];
```
AliSoftware, line 22: `va_end(args);` here?

Documentation/templates/strings/objc.md
```diff
@@ -0,0 +1,53 @@
+## Template Information
+
+| Name      | Description       |
+| --------- | ----------------- |
+| File names | strings/objc-string-h.stencil & strings/objc-string-m.stencil |
+| Configuration example | <pre>strings:<br />  inputs: path/to/Localizable.strings<br />  outputs:<br />    - templateName: objc-string-h<br />      output: Localizable.h</pre><br />    - templateName: objc-string-m<br />      output: Localizable.m</pre> |
```
AliSoftware: Extra `</pre>` in the middle of that block makes the rendering bad after the first `templateName`+`output`: the second `templateName`+`output` for the `.m` isn't rendered in monospaced nor properly. Removing that extra `</pre>` before the `<br />` should solve it.

Documentation/templates/strings/objc.md
djbe: Forgot that was even possible, read it in some obscure article somewhere I think.

Would the compiler even know how to autocomplete that correctly? The `and...and` version'll work fine. This, not so sure?

templates/strings/objc-m.stencil
djbe: Right. And maybe move the `va_start` & stuff to right before this, to make it clear what it's for.

Documentation/templates/strings/objc.md
djbe: Depends. Do we expect to merge #490 as well as this PR? We can, but I don't think it's necessary.

Documentation/templates/strings/objc.md
AliSoftware: I think it will. It's used in the SDK at least by the `+[CAMediaTimingFunction functionWithControlPoints::::]` method, you could experiment with that if you want 😉

AliSoftware: Note: if we still go with named parameters I think `with:` would generally be a better fit than `and:`? (But I have to admit I don't remember the coding guidelines for ObjC that well 😅)

Tests/Fixtures/Generated/Strings/objc-m/localizable-headerName.m
mr-fixit: good catch! i'll fix that.

Documentation/templates/strings/objc.md
mr-fixit: I think `with:...` is better than `and:...`.   I think they both are references to existing (non-auto-generated) signatures that look like `-[doSomethingWith:and:and:]`. 

I originally had `p1:p2:p2:`, but that's not great either, and I was happy to switch away from it.

Documentation/templates/strings/objc.md
mr-fixit: Do the Swift strings templates generate Swift code that can be used from ObjC? I don't see that `enum L10n` getting into "<projectname>-Swift.h".

mr-fixit: I'll add the following to `Documentation/templates/strings/objc.md`

> When you need to generate *Objective-C* code.  Perhaps because you have a pure Objective-C project and don't want to add a Swift file to it (thus triggering the inclusion of the Swift runtime into your build).  Or perhaps because you can't find a template that generates Swift that can be accessed from Objective-C.

Documentation/templates/strings/objc.md
mr-fixit: I agree, I like your #2 better than mine, I've changed it slightly and it will now read:

>The name of the ObjC header generated by the `objc-string-h` template.  The generated `.m` file will `#include` "{{headerName}}".  This parameter should have the same value as the `output` file name used for the `template: objc-string-h` entry in your config file.

I also changed the .m template so it doesn't append ".h" to this parameter.

Documentation/templates/strings/objc.md
mr-fixit: Is there a consensus?

templates/strings/objc-m.stencil
mr-fixit: I'll do it!

Documentation/templates/strings/objc.md
mr-fixit: got it, thanks

Documentation/templates/strings/objc.md
```diff
@@ -0,0 +1,53 @@
...
+- When you need to generate *Objective-C* code.  Perhaps because you have a pure Objective-C project and don't want to add a Swift file to it (thus triggering the inclusion of the Swift runtime into your build).  Or perhaps because you can't find a template that generates Swift that can be accessed from Objective-C.
+
+## Customization
+
+You can customize some elements of this template by overriding the following parameters when invoking `swiftgen`. See the [dedicated documentation](../../ConfigFile.md).
+
+| Parameter Name | Default Value   | Description                                                  |
+| -------------- | --------------- | ------------------------------------------------------------ |
+| `noComments`   | N/A             | Setting this parameter will disable the comments describing the translation of a key. |
+| `headerName`   | "Localizable.h" | The name of the ObjC header generated by the `objc-string-h` template.  The generated `.m` file will `#include` "{{headerName}}".  This parameter should have the same value as the `output` file name used for the `template: objc-string-h` entry in your config file. |
```
djbe, line 21: The config example seems wrong. Should be something like:
```yaml
templateName: objc-h
```

Documentation/templates/strings/objc.md
```diff
@@ -0,0 +1,53 @@
+## Template Information
+
+| Name      | Description       |
+| --------- | ----------------- |
+| File names | strings/objc-string-h.stencil & strings/objc-string-m.stencil |
+| Configuration example | <pre>strings:<br />  inputs: path/to/Localizable.strings<br />  outputs:<br />    - templateName: objc-string-h<br />      output: Localizable.h<br />    - templateName: objc-string-m<br />      output: Localizable.m</pre> |
```
djbe: The config example seems wrong (wrong template names). Should be something like:
```yaml
strings:
  inputs: path/to/Localizable.strings
  outputs:
    - templateName: objc-h
      output: Localizable.h
    - templateName: objc-m
      output: Localizable.m
```

Documentation/templates/strings/objc.md
djbe: Looking at the media timing signature, it'd have to be `...WithValues: : : :`. (could be "values", "arguments", ...)

Tests/TemplatesTests/StringsTests.swift
```diff
@@ -33,6 +33,19 @@ class StringsTests: XCTestCase {
...
 
+  let variationsObjC: VariationGenerator = { name, context in
+    guard name == "localizable" else { return [(context: context, suffix: "")] }
+
+    return [
+      (context: context,
+       suffix: ""),
+      (context: try StencilContext.enrich(context: context, parameters: ["noComments"]),
+       suffix: "-no-comments"),
+      (context: try StencilContext.enrich(context: context, parameters: ["headerName=L10N"]),
```
djbe: Pretty sure this should become `"headerName=localizable-headerName.h"`? (if we're ever going to test this)

djbe (changes requested):

Documentation/templates/strings/objc.md
mr-fixit: that works for me.

Documentation/templates/strings/objc.md
mr-fixit: good catch, thanks

Documentation/templates/strings/objc.md
mr-fixit: (see above)

Documentation/templates/strings/objc.md
mr-fixit: I'll do the test for count==1, so we'll have `...WithValue:` and `...WithValues:`.  It makes for an uglier stencil, but prettier generated code.

Tests/TemplatesTests/StringsTests.swift
mr-fixit: 'L10N' is different from the default 'Localizable', so I think this exercises the feature, no?

Is there already a scaffolding for testing the content of a generated file?  If so, I would test that the generated .M file includes the right value in its `#include ...` line.

djbe: Well no that's for a later PR (hopefully). But just for us, for clarity of the tests (and maybe for our users), it probably should have a correct value.

Note: it should at least be `L10n.h`, as the template doesn't add the `.h` anymore.

mr-fixit: good catch (on lack of ".h")!  

and I'll switch to `headerName=headerName-from-parameter.h` to make it clear what's going on.

djbe (approved):
With the exception of @AliSoftware's last comment, looks good to merge to me!

mr-fixit:
"Why the extra space?"  Oh, man! Cue up the Typhora rant! BBEdit (or `vi` or `vim` or anything that doesn't get in the way) to the rescue.

AliSoftware (approved):
Thanks for the 🎁 Typhora 😄🤦‍♂️😛

AliSoftware:
It's time to Party

templates/strings/objc-h.stencil
```diff
@@ -0,0 +1,66 @@
...
+  {% endif %}
+{% endmacro %}
+{% macro emitOneMethod table item %}
+{% for string in item.strings %}
+{% if not param.noComments %}
+// {{string.key}} --> "{{string.translation}}"
+{% endif %}
+{% if string.types %}
+  {% if string.types.count == 1 %}
++ (NSString*){{string.key|swiftIdentifier:"pretty"|lowerFirstWord}}WithValue:{% call parametersBlock string.types %}
```
WingedDoom, line 44: Doesn't this (and the following one) method declaration stencil need a `;` at the end? I tried using this stencil in my project and resulting headers don't compile because of missing `;` after every method declaration. The error went gone when I added `;` at these two spots myself.

mr-fixit: i agree, it's missing the semi-colon. i probably fixed the stencil in my local project and forgot to update this repo.

thanks for catching this, @wingeddoom

mr-fixit: fixed by [PR 672](https://github.com/SwiftGen/SwiftGen/pull/672).

odnaks:
parameters doesn't work (headerName, noComments)

https://github.com/SwiftGen/SwiftGen/blob/stable/Documentation/templates/strings/objc.md#customization

milamina:
Hello,
Thank you for the template of strings using for Objective-C, it work for me!
Is there another one for Assets (Images) ?
Thanks

mr-fixit:
(pardonnez-moi mauvis Francais)

je vous en prie!

je n’a pas le temps pour faire un template pour vous, desolee.

peut etre si vous trouve quel qu'un a le SwiftGen repository site a Github?

bon channe!

eric

milamina:
@AliSoftware @djbe 
Hello,
Thank you for the template of strings using for Objective-C, it work for me!
Is there another one for Assets (Images) ?  
Note : (you can response in French/English ;) )
Thanks

milamina:
@mr-fixit 
You can response in English also ;)
You said you cannot write it for me and you've suggest some thing else but I don't understand what is-it? 

Thanks

mr-fixit:
Ha! I was trying to say “see if someone who follows Swiftgen repo might do it”.

You can also do it yourself, the syntax for template language is a bit hard to wrap your head around, but it’s doable.

djbe:
There currently isn't an obj-c template for asset catalogs. You're free to write your own, you can easily base yourself on the existing swift templates. You'll need to modify it to generate a "flat" output (instead of the structured output), similar to the flat-swift strings template, which this obj-c template is based on.

You can then create a PR to add the template to SwiftGen itself. I'm honestly not sure if it's worth the effort to add such templates to swiftgen itself... As this obj-c strings template kinda "unmaintained", it isn't really tested like the other templates. Might be worth having a separate "unofficial" templates repository, where templates such as these can be added, with the necessary warnings.

As this discussion isn't related to the pull request, I'm going to lock the conversation. Feel free to open a dedicated issue if you'd like to discuss this further.

PR #392, opened 2018-03-17
Release/5.3.0

AliSoftware:
TODO:

* [x] Migrate to Xcode 9.2 (downloading it rn…) — don't forget to update `circleci.yml`
* [x] Update Kanna to 4.0
* [ ] Run tests after pods being updated
* [x] Update SwiftLInt to 0.25
* [x] Update Version in podspec and CHANGELOG


djbe:
Maybe update SwiftLint to 0.25, although that can wait until later. Especially because we're still on 0.23, and 0.24 added some rules we may want to enable.

It's annoying that Kanna doesn't really have a changelog, makes it hard to check beforehand what the impact'll be, and/or if there are new features to use.

AliSoftware:
Yeah, that's why I decided to try to update it separately :wink:
Kanna 4.0 doesn't compile with Xcode 9.1, hence me downloading 9.2

AliSoftware:
@djbe I think I broke all the tests… not sure why they're mixing up stuff between iOS/macOS contexts or whatnot… and I need a break from all those small downloads & updates I had to do for this release and have to get back to my house work 😓 So if you feel so, could you take a look?

AliSoftware:
In particular:
* XCAssets seems broken and don't seem to detect any asset anymore
* Storyboards-macOS seems to believe they are iOS (import UIKit instead of Cocoa, etc)
* Colors does `internal enum UIColor` even when the customName is the special case `UIColor` which shouldn't generate it

Maybe we should revert to a previous version of Kanna and look at that in a separate PR if it's too hard to find what's the root cause (and maybe migrating to Kanna 4 was too dangerous without further testing), idk

djbe:
I'll take a look, just got back.

PR #396, opened 2018-03-19
Feature/fix fonts

### LINKED ISSUE #394: Fix small memory leak in Fonts template
[here](https://github.com/SwiftGen/SwiftGen/blob/master/templates/fonts/swift4.stencil#L27-L28) in the templates, we create an `Unmanaged<CFError>` but never release it.

We should instead:

* Either if we really don't care about the error, don't even use an `Unmanaged<CFError>` but pass `nil` as the third parameter
* Or if we do care about the error, use `let e = errorRef.takeRetainedValue()` to gain back ownership of the `CFError` and avoid the leak, and `throw e` if not `nil` (that would mean that `register()` would now be `throws`, and we could always call `try?` to ignore the error at call site later in the template — note though that it would mean that the template would be breaking the API generated by the template if some people did call `register()` themselves

AliSoftware:
Fixes #394 

* Fixes memory leak in generated code for Fonts
* Add ability to list all custom fonts and register all them using `FontFamily.registerAllCustomFonts()`

### DIFF
--- a/templates/fonts/swift4.stencil
+++ b/templates/fonts/swift4.stencil
@@ -24,8 +24,7 @@
 
   {{accessModifier}} func register() {
     guard let url = url else { return }
-    var errorRef: Unmanaged<CFError>?
-    CTFontManagerRegisterFontsForURL(url as CFURL, .process, &errorRef)
+    CTFontManagerRegisterFontsForURL(url as CFURL, .process, nil)
   }
 
   fileprivate var url: URL? {
--- a/templates/fonts/swift4.stencil
+++ b/templates/fonts/swift4.stencil
@@ -64,8 +64,13 @@
     {% for font in family.fonts %}
     {{accessModifier}} static let {{font.style|swiftIdentifier:"pretty"|lowerFirstWord|escapeReservedKeywords}} = FontConvertible(name: "{{font.name}}", family: "{{family.name}}", path: "{% call transformPath font.path %}")
     {% endfor %}
+    {{accessModifier}} static let all: [FontConvertible] = [{% for font in family.fonts %}{{font.style|swiftIdentifier:"pretty"|lowerFirstWord|escapeReservedKeywords}}{% if not forloop.last %}, {% endif %}{% endfor %}]
   }
   {% endfor %}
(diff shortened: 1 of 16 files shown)


templates/fonts/swift4.stencil
```diff
@@ -64,8 +63,13 @@
     {% for font in family.fonts %}
     {{accessModifier}} static let {{font.style|swiftIdentifier:"pretty"|lowerFirstWord|escapeReservedKeywords}} = FontConvertible(name: "{{font.name}}", family: "{{family.name}}", path: "{% call transformPath font.path %}")
     {% endfor %}
+    {{accessModifier}} static let all: [FontConvertible] = [{% for font in family.fonts %}{{font.style|swiftIdentifier:"pretty"|lowerFirstWord|escapeReservedKeywords}}{% if not forloop.last %}, {% endif %}{% endfor %}]
```
djbe: I'd make these multiline? It's what we do in (for example) the assets templates. Same comment for the other `all` constant.

templates/fonts/swift4.stencil
```diff
@@ -64,8 +63,13 @@
...
   }
   {% endfor %}
+  {{accessModifier}} static let allCustomFonts: [FontConvertible] = [{% for family in families %}{{family.name|swiftIdentifier:"pretty"|escapeReservedKeywords}}.all{% if not forloop.last %}, {% endif %}{% endfor %}].flatMap { $0 }
+  {{accessModifier}} static func registerAllCustomFonts() {
+    allCustomFonts.forEach { $0.register() }
```
djbe: I almost never use `forEach`, always `for ... in ...` for some reason.

djbe: Probably a remnant from my Objective-C days, where `for ... in` was more performant than the block based `forEach`.

templates/fonts/swift4.stencil
AliSoftware: I wondered about that. But I figured we rarely have hundreds of fonts in a given font family, at most 5-10, and we rarely have hundreds of custom font families either (unlike assets), so…

djbe: True, let's keep it as a oneliner.

templates/fonts/swift4.stencil
AliSoftware: Yeah I think in Swift that's the opposite, `forEach` might be optimised while `for … in` might not (because `for … in` can break? tomato tomato though

AliSoftware:
@djbe Ok to merge?

djbe:
yep :D

PR #398, opened 2018-03-23
Stop manipulating strings

### LINKED ISSUE #257: Structured strings template and camel cased Localizable.strings
My Localizable.strings have a bunch of keys similar to this one, all correctly camelCased. Some keys also includes uppercased acronyms.

`"myViewController.userABC.nameText" = "My localization";`

The problem is that using the dot-syntax-swift3 template I get something similar to this:

```
enum Myviewcontroller {
    enum Userabc {
        static let nameText = ...
    }
}
```

As you can see all the enum names lose the camel case of the original Localizable key. 

I think that the right behavior should be to preserve the camel case, just uppercasing the first letter of the enum's name.

The change is simple (I already did that and can make a PR), but it will break a bunch of tests because it seems that this behavior has been done by purpose to merge together keys with same name but different capitalization.

What is your opinion on this? I already know that I could use snake case, but I don't like that and I will lose anyway the acronyms capitalization.

dchohfi:
Hey guys, this is a try to fix #257.
I don't know if I did what @AliSoftware is thinking of, but I would love to help fixing this issue.

As this is a break change, I don't know if I need to follow a specific process to create the PR.

More than happy to contribute to the library.

---


dchohfi:
Ok, this is not that easy. Do I need to change the filters?

djbe:
TBH, before you continue fixing tests & we do a full review, it's still worth finishing the discussion in the issue. We can discuss concrete steps here if you want.

From a quick check, I'm not sure if this is a good solution, this eliminates *your* issue by disabling the normalisation of keys. As you can see from the test results, this can lead to failing compilation because of 2 types with the same name (but with other casing).

I totally agree that in a normal (decent) code base, this shouldn't happen. The question is, do we accept this as failure as normal and expected, or do we prefer that SwiftGen fixes this issue for us? A third option is for SwiftGen to generate an error for duplicate keys (same text if normalized, different casing).

dchohfi:
Hey @djbe! Thank you for your reply, I knew from beginning that my solution was too simple and I would probably need to work on a more robust one, how can we continue debating about the solution?

I agree with you that we don't normally face good code base when implementing SwiftGen in something that is already running, but It think that SwiftGen should also be responsible for pointing that out, make SwiftGen not only a library that generates code, but also help developers to keep a better code base (even on the static files, like strings and images). So, if something is duplicated, the script should fail, what you think?

AliSoftware:
The more I think about it the more I feel like this shouldn't be the responsibility of SwiftGen to normalise keys.

We still don't have a way to fold those similar-keys-where-only-case-differ into one on the template side (which would solve all the problems, because that way people would be able to choose to normalise keys or not in their template), but even though, in 99% of cases it should be the developer's responsibility to ensure their Localizable.strings keys are consistent in casing. If we chose to continue to go the route of "SwiftGen should fix the input case" then why would we stop at this and not also fix if people sometimes use diacritics and sometimes not, or sometimes uses plural and sometimes not… that shouldn't be our job in the end, but the developer's responsibility all the same.

Sure, stopping to normalise the key at parsing, and only normalise them via the template, could break some code bases — if people use same keys with different cases, now the templates would generate two enums with the same name and fail to compile. But that was already a case that could happen if people did a custom template and chose to normalise the name of the generated enums in a different way (say, force all titlecase or all lowercase) that the one used but the parser. So the current solution of normalising the keys at parsing time is not really a solution anyway, it's just a workaround which happens to work just because the most common templates we provide happen to use the same normalisation as the one used at parsing time.

At least if we stop normalising the keys at parsing time like this PR proposes, we could choose the normalisation that we want at template site (or none at all to keep the same case as the one mentioned in `Localizable.strings`), and sure it could generate duplicate enums and stop compiling, but just like it already can with the current solution now. And the rare case where that would happen feels super-rare to me, and if it happens would be the responsibility of the developer to either fix the Localizable.strings keys or use a custom template to normalise them

_The only thing we could think of to add to that new solution is to maybe later add some `reduce` filter/tag or something to Stencil to recursively turn a dictionary into another dictionary with keys recursively folded according to a normalisation operator (like mapping all the keys to normalise them using a user-provided filter and merge values ending up having the same keys after normalisation), but 
even that feels like a bonus that we don't absolutely need right away._

@djbe thoughts about moving that forward?

djbe:
Good for me! I was ok with the change, just wanted to discuss the repercussions.

To fix the tests, we'll have to remove the cases that were meant to test this, and now generate duplicate definitions.

AliSoftware:
Indeed.

@dchohfi could you clean up the tests as suggested by @djbe ?
Also it would be a good occasion to rebase your changes on top of master (especially since we migrated to CircleCI 2.0 since, so the checks on GitHub to make the PR valid won't work until you do)

.ruby-version
```diff
@@ -1 +1 @@
-ruby-2.4.2
+2.4.2
```
djbe: Why is this change needed? It worked fine before on both CI and here locally.

dchohfi: hey @djbe, my rbenv failed to find this version, but could find `2.4.2`. it wasn't listed on rbenv list of available versions. don't really know why, but I normally see ruby version declared with only the desired version, without the `ruby-`. this is a problem? want me to revert that?

dchohfi:
Hey guys, still looking around to fix the integration tests, give me some hours to investigate by myself and will send it ASAP.

.ruby-version
djbe: Funny, I also use `rbenv`. I remember spending quite a while getting it to work both locally as on CI. The change without `ruby-` works for me locally, so that doesn't seem to be an issue (anymore). If this works for everyone, then I definitely won't complain 🤷‍♂️ 

Note: what might help is manually specifying what version to install with `rbenv install 2.4.2`.

djbe:
Tip: I'm pretty sure you have to remove the corresponding "wrong" keys from the test fixtures (strings files), or correct them. Then regenerate the contexts, and then the output.

https://github.com/SwiftGen/SwiftGen/blob/master/Tests/Fixtures/Resources/Strings/Localizable.strings

AliSoftware:
@dchohfi in case you're not aware, there are dedicated schemes in the SwiftGen xcodeproj to auto-regenerate contexts and output.

Of course if you use those schemes the expected outputs will be overridden by the output of the tests, and the tests will then always pass because the expected output will be the generated output by definition in that special case, so if you do use that be sure to control by hand that what is generated looks ok and as expected. But at least it allows you to avoid editing each context and expected output manually by hand when you change the tests like here.

dchohfi:
Hey @AliSoftware, I’ve just realized that, this is a really nice way for doing integration tests. I fixed the generated code and ran the rake task to compile the code. They are passing now, but did I did everything right? Is the solution enough to fix it? Thank you.

dchohfi:
Any thoughts? @alisoftware @djbe

CHANGELOG.md
```diff
@@ -12,7 +12,9 @@
 
 ### Breaking Changes
 
-_None_
+* Change `StringsContext.swift` to stop manipulating strings.
```
djbe: Nitpick: add two spaces after the `.`

dchohfi:
`Don't normalize string keys while parsing, let all transformation be done on template side. This makes the developer responsible to keep the language file organized, duplications won't be removed.`

What you think @AliSoftware @djbe?

Really happy to know that this is approved, was a nice experience to be able to contribute.

What is the process to submit a new version to CocoaPods? Could I help with that too?

Thank you.

djbe:
There are still some more items we want to merge before releasing this version, especially because this is a major update (due to the breaking changes). For now though, you can always create a local version with your changes using the `rake cli:install` command.

PR #401, opened 2018-03-27
Use SwiftLint via CocoaPods

### LINKED ISSUE #393: [Internal] Integrate SwiftLint via CocoaPods instead of bundle
SwiftLint is now also distributed as a pod (similarly to SwiftGen, actually), so we could integrate it `pod 'SwiftLint', '~> 0.25'` in our `Podfile`.

This would let us stop relying on `bundler` to run it in the build phase (which doesn't really work on all people's configurations, depending on their GEM_HOME config and all) and instead just change the Build Phase to `"$PODS_ROOT/SwiftLint/swiftlint --path …"` to lint code & tests

djbe:
Fixes #393.

Note: similar StencilSwiftKit PR is https://github.com/SwiftGen/StencilSwiftKit/pull/78


Podfile
```diff
@@ -19,6 +24,7 @@ target 'swiftgen' do
 end
 
 target 'SwiftGenKit' do
+  common_pods
   podspec :path => 'SwiftGenKit.podspec'
```
AliSoftware: Now that we have `common_pods`, `SwiftGenKit` should be moved into `common` too, right?

SwiftGen.xcodeproj/project.pbxproj
```diff
@@ -1119,6 +1124,76 @@
...
+			);
+			inputPaths = (
+			);
+			name = "⚠️ SwiftLint";
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "if [[ ! $CI && ! $NO_CODE_LINT ]]; then\n  \"$PODS_ROOT/SwiftLint/swiftlint\" lint --strict --config \"${SRCROOT}/.swiftlint.yml\" --path \"${SRCROOT}/Tests/TemplatesTests\"\nfi";
+		};
```
AliSoftware: Not sure why we have four of those in the pbxproj here? (although, reading that from GitHub, maybe that's a glitch of how Xcode serializes the project?)

SwiftGen.xcodeproj/project.pbxproj
```diff
@@ -1119,6 +1124,76 @@
...
+			files = (
+			);
+			inputPaths = (
+			);
+			name = "⚠️ SwiftLint (generated output)";
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "if [[ ! $CI && ! $NO_CODE_LINT ]]; then\n  \"$PODS_ROOT/SwiftLint/swiftlint\" lint --strict --config \"${SRCROOT}/.swiftlint.yml\" --path \"${SRCROOT}/Tests/Fixtures/Generated\"\nfi";
```
AliSoftware: Now that we have that kind of integration, is it really worth splitting the linting of the code and of the generated output in separate Build Phases?

circle.yml
```diff
@@ -8,7 +8,6 @@ checkout:
 
 dependencies:
   post:
-    - bundle exec rake lint:install
```
AliSoftware: Are the invocations of `bundle exec rake lint:*` below still necessary too, if they now happen to be fully integrated as a Build Phase in the Xcode project, and thus will already be run by `bundle exec rake xcode:test` (I think)?

```
 test:
   override:
     - bundle exec rake lint:code
     - bundle exec rake lint:tests
     - bundle exec rake lint:output
```

rakelib/lint.rake
```diff
@@ -1,64 +1,31 @@
...
-      Utils.print_header 'Linting the code'
-      config = Pathname.getwd + '.swiftlint.yml'
-      config = config.exist? ? "--config #{config}" : ''
-      Utils.run(%(swiftlint lint --no-cache --strict --path Sources #{config}), task)
-    end
+  desc 'Lint the code'
+  task :code do |task|
+    Utils.print_header 'Linting the code'
+    config = Pathname.getwd + '.swiftlint.yml'
+    Utils.run(%(#{SWIFTLINT} lint --strict --path Sources --config "#{config}"), task)
```
AliSoftware: Same question, worth wondering if we should really keep those tasks in the Rakefile, especially since, now that they're integrated as Build Phases, I fear that we could miss keeping them in sync if we change those one day (like change the paths in one place but not in the other)…?

circle.yml
djbe: The xcode build phases are wrapped in a
```bash
if [[ ! $CI && ! $NO_CODE_LINT ]]; then
```

The CI steps are separate so that they are separate "steps", and can happen in parallel, have separate logs, etc...

SwiftGen.xcodeproj/project.pbxproj
djbe: So that when you're modifying SwiftGenKit, it'll lint those files when building that framework (before doing anything else), and trigger errors/warnings for only those files.

SwiftGen.xcodeproj/project.pbxproj
djbe: Because it's linting different files in each build step.

circle.yml
AliSoftware: Right. We should just be careful to be sure to maintain both (Build Phases vs Rake tasks) in sync. Alternatively, we could make those tasks simple bash scripts in a `scripts/` folder, then invoking those bash scripts both in the Build Phases (wrapped in `if !$CI`) and in `circle.yml`

djbe: Maybe keep the rake tasks as wrappers for the bash scripts? That way we can keep having a unified place for triggering "stuff" from CI and CLI (locally).

AliSoftware: Totally!

Scripts/SwiftLint.sh
```diff
@@ -1,7 +1,12 @@
 #!/bin/bash
 
-SWIFTLINT="$(PWD)/Pods/SwiftLint/swiftlint"
-CONFIG="$(PWD)/.swiftlint.yml"
+if [ -z "$PROJECT_DIR" ]; then
+	pushd "$(dirname $0)/.." > /dev/null
+	PROJECT_DIR=`pwd`
+	popd > /dev/null
```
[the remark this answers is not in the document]
AliSoftware: If you wish to use some bash foo, and still keep the `PROJECT_DIR` variable, you might want to use something like: `PROJECT_DIR=${PROJECT_DIR:-$(dirname $0)/..}`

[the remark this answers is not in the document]
djbe: I actually initially tried `$(dirname $0)/..`, but swiftlint crashes on that (can't read the config file). This is the first thing I found that normalizes the path, that isn't some massive function or depends on python (or something else).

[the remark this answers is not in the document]
AliSoftware: 1. If that's an issue, there's always the ability to explicitly pass to SwiftLint the exact path to the `.swiftlint.yml` file
2. Otherwise, if SwiftLint can't read the config file when you use `$(dirname $0)/..` directly, that means that it will likely crash in other situations depending on how you invoke that script, because after all if `PROJECT_DIR` isn't initially set, then your script will set it to the same value as `$(dirname $0)/..` anyway, probably leading to the same behavior?

The fact that `popd` exits with error if you pass to `pushd` a path that points exactly to the current directory (which means it won't push anything so the directory stack will still be empty and `popd` will error) makes me think that's not the best solution either anyway.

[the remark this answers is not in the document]
AliSoftware: Another simpler option would be to `cd $(dirname $0)/..` at the beginning of the script and `cd -` back at the end of the script. If we do that though, we'd just have to be careful to put that `cd -` not just at the end of the script, but also on interruption or error during shell (`trap … EXIT`)

```bash
cd $(dirname $0)/..
function restore_dir {
  cd -
}
trap restore_dir EXIT

… rest of the script …
```

Scripts/SwiftLint.sh
```diff
@@ -1,7 +1,8 @@
 #!/bin/bash
 
-SWIFTLINT="$(PWD)/Pods/SwiftLint/swiftlint"
-CONFIG="$(PWD)/.swiftlint.yml"
+PROJECT_DIR="${PROJECT_DIR:-`cd "$(dirname $0)/..";pwd`}"
```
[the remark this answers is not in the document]
djbe: Note: I missed your last comment about `trap` & stuff 😅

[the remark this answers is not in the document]
AliSoftware: It not necessary if you do like you did, because using backticks like you did here spawns a new, independent shell to run the `cd "$(dirname $0)/..";pwd` whole command at once in a separate shell and returns the result. So because that full command is in a separate shell it shouldn't affect the current working directory of the current shell (well… I _think_).

djbe:
Good to merge?

AliSoftware:
I'll let you do the same changes on the StencilSwiftKit one 👍

PR #402, opened 2018-03-27
Enable extra SwiftLint rules

djbe:
Been wanting to do this for a while, this enables a bunch of optional SwiftLint rules that ensures the codebase is a bit more consistent.

Note: built on top of #401, wait until that is merged.
Note: similar StencilSwiftKit PR is https://github.com/SwiftGen/StencilSwiftKit/pull/79

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,6 +1,3 @@
-disabled_rules:
- - type_name
-
 opt_in_rules:
  - array_init
  - attributes
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -21,7 +21,6 @@ opt_in_rules:
  - operator_usage_whitespace
  - overridden_super_call
  - override_in_extension
- - pattern_matching_keywords
  - private_action
  - private_outlet
  - prohibited_super_call
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -2,7 +2,36 @@ disabled_rules:
  - type_name
 
 opt_in_rules:
+ - array_init
+ - attributes
+ - closure_end_indentation
+ - closure_spacing
+ - contains_over_first_not_nil
+ - discouraged_optional_boolean
+ - empty_count
+ - fatal_error_message
+ - first_where
  - force_unwrapping
+ - implicit_return
+ - implicitly_unwrapped_optional
+ - joined_default_parameter
+ - literal_expression_end_indentation
+ - multiline_arguments
+ - multiline_parameters
+ - operator_usage_whitespace
+ - overridden_super_call
+ - override_in_extension
+ - pattern_matching_keywords
+ - private_action
+ - private_outlet
+ - prohibited_super_call
+ - redundant_nil_coalescing
+ - sorted_first_last
+ - sorted_imports
+ - trailing_closure
+ - unneeded_parentheses_in_closure_argument
+ - vertical_parameter_alignment_on_call
+ - yoda_condition
 
 # Rules customization
 line_length:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,9 @@ _None_
 * Switched to using SwiftLint via CocoaPods instead of our own install scripts.  
   [David Jennes](https://github.com/djbe) 
   [#401](https://github.com/SwiftGen/SwiftGen/pull/401)
+* Enabled some extra SwiftLint rules for better code consistency.  
+  [David Jennes](https://github.com/djbe) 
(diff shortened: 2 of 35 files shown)


.swiftlint.yml
```diff
@@ -2,7 +2,36 @@ disabled_rules:
  - type_name
 
 opt_in_rules:
+ - array_init
```
AliSoftware: I didn't know that rule, seems to be for preventing to do `seq.map { $* }` instead of `Array(seq)`, I didn't know some people would do the former 😄 :+1:

.swiftlint.yml
```diff
@@ -2,7 +2,36 @@ disabled_rules:
...
+ - attributes
+ - closure_end_indentation
+ - closure_spacing
+ - contains_over_first_not_nil
+ - discouraged_optional_boolean
```
AliSoftware: Hell yeah!

.swiftlint.yml
```diff
@@ -2,7 +2,36 @@ disabled_rules:
...
+ - implicit_return
+ - implicitly_unwrapped_optional
+ - joined_default_parameter
+ - literal_expression_end_indentation
+ - multiline_arguments
+ - multiline_parameters
+ - operator_usage_whitespace
+ - overridden_super_call
+ - override_in_extension
+ - pattern_matching_keywords
```
AliSoftware: > Combine multiple pattern matching bindings by moving keywords out of tuples.

I'm not fan of this one. I actually personally prefer `case (let x, let y)` over `case let (x, y)` generally. Because if there's a variable named `y` before that line, the intent of `case let (x, y)` is unclear whereas if `x` is binded or just has to match the existing, previously-defined `y`…

.swiftlint.yml
```diff
@@ -2,7 +2,36 @@ disabled_rules:
...
+ - private_action
+ - private_outlet
```
AliSoftware: Not sure we have a lot of `@IBAction` and `@IBOutlet` in our command-line based tool 😆  But /shrug

.swiftlint.yml
```diff
@@ -2,7 +2,36 @@ disabled_rules:
...
+ - prohibited_super_call
+ - redundant_nil_coalescing
+ - sorted_first_last
+ - sorted_imports
+ - trailing_closure
+ - unneeded_parentheses_in_closure_argument
+ - vertical_parameter_alignment_on_call
+ - yoda_condition
```
AliSoftware: Given that in Swift, contrary to most C-based languages, the affectation of a variable doesn't return the affected value but `Void`, it's not possible to write `if x = 42` instead of `if x == 42` because `x = 42` is an expression returning `Void`. That's one great thing about Swift.

So that being said, given that we already have safeguards provided by Swift here for those kind of mistakes, I think it's more important to make the code readable (`if foo == 42` reads better imho) rather than… what? not ever more safe, as it's already protected against `if foo = 42` mistakes, so…

Podfile
```diff
@@ -3,7 +3,12 @@ use_frameworks!
...
 
+def common_pods
+  pod 'SwiftLint', '~> 0.25'
+end
+
 target 'swiftgen' do
+  common_pods
   pod 'Commander', '~> 0.8'
   pod 'StencilSwiftKit', '~> 2.3'
   podspec :path => 'SwiftGenKit.podspec'
```
AliSoftware: Same notes as #401: maybe that could be moved to `common_pods` now?

.swiftlint.yml
djbe: Just enabling almost all extra rules that didn't have a big impact 😄

.swiftlint.yml
djbe: Uhm, `yoda_condition` enforces that the variable comes first, the constant later, which is what you want?

.swiftlint.yml
djbe: Good point, I'll remove it.

.swiftlint.yml
AliSoftware: Oh. Right. I misread the rule description, especially given that in C-based languages, what we usually call writing "Yoda Condition" is writing `1 == foo` (yoda style) instead of `foo == 1` to avoid the risk of `=` instead of `==`… So I thought given the rule name that it expected us to write in yoda style…

But in fact you're right (I guess the rule is misleading), so let's keep it 👍

Podfile
djbe: Note: this is a change from #401, so it'll disappear from this PR once that one is merged.

PR #403, opened 2018-03-30
Migrate to CircleCI 2.0

### LINKED ISSUE #397: Migrate to CircleCI 2.0
CircleCI 1.0 is sunsetting in august. We'll have to migrate to CircleCI 2.0 before that, by updating our config file.

* Docs: [Migrating from 1.0 to 2.0](https://circleci.com/docs/2.0/migrating-from-1-2/), [Migrating iOS Projects from 1.0 to 2.0](https://circleci.com/docs/2.0/ios-migrating-from-1-2/)
* Starting point, by [using the config-translation endpoint](https://circleci.com/api/v1.1/project/github/SwiftGen/SwiftGen/config-translation)

djbe:
Fixes #397.

I've attempted to split up the workflow in a few jobs, each grouping similar commands, such as moving all lint checks into one job. For now this boils down to:
- linting
- building & testing
- checking the output code
- checking deployment (only the podspec for now)

At some point we can add more jobs (such as danger), or even disable some jobs for certain branches. For example, we could disable the `check-deploy` job for all branches except the master branch.


.circleci/config.yml
```diff
@@ -0,0 +1,128 @@
...
+        BUNDLE_RETRY: 3
+  - &store-gems
+    save_cache:
+      key: v1-gems-{{ checksum "Gemfile.lock" }}
+      paths:
+        - vendor/bundle
+  - &fetch-xcode-logs
+    run:
+      name: Getting Xcode activity logs
+      command: find $HOME/Library/Developer/Xcode/DerivedData -name '*.xcactivitylog' -exec cp {} $CIRCLE_ARTIFACTS/xcactivitylog \; || true
```
AliSoftware: This command doesn't seem to be referenced anywhere else in the config file?

djbe: Huh, missed that. For now it's only needed in the `build-and-test` job, but could be used in more jobs if needed.

PR #417, opened 2018-04-29
Refactor parsers

djbe:
As it is right now, quite a few parsers are one long file containing both the (internal) data model and the actual parsing code. Some of these parsers will only get more complex as times goes on, so we might as well reorganise them a bit.


djbe:
I think the easiest way to compare is to check each individual commit in this case. Each commit is for a specific parser, the last commit is just a small rename to avoid swift compiler issues.

There shouldn't be any functional changes, only moving types (and related code) to their own files. Some small things of note:
- Asset catalog: `Entry` related parsing code is in the `Entry` file. The small general function to parse a directory is still in the parser.
- Storyboards: instead of 1 large `XML` enum, each type (scene, segue, ...) now has a file with the related XML parsing code.

Sources/SwiftGenKit/Parsers/AssetsCatalogParser/CatalogEntry.swift
```diff
@@ -0,0 +1,104 @@
...
+    if let properties = contents[AssetCatalog.properties] as? [String: Any],
+      let providesNamespace = properties[AssetCatalog.providesNamespace] as? Bool {
+      return providesNamespace
+    } else {
+      return false
+    }
+  }
+
+  private var contents: [String: Any] {
+    let contents = self + Path(AssetCatalog.path)
```
AliSoftware: I didn't realize we had that on an extension on `Path`, not sure it's a good idea since we extend a class that is not our own (but PathKit's) to add a property whose name could potentially be added in the futur, and   a  conflict with ours. Even if it's private, if an official one gets added I'm not sure about the behavior?

At least we should rename this `jsonContent`?

Sources/SwiftGenKit/Parsers/AssetsCatalogParser/CatalogEntry.swift
```diff
@@ -0,0 +1,104 @@
...
+        items: Catalog.process(folder: path, withPrefix: subPrefix)
+      )
+    }
+  }
+}
+
+// MARK: - Private Helpers
+
+private extension Path {
+  var isNamespaced: Bool {
```
AliSoftware: Same, even if it's private, feels strange to make all `Path` instances respond to this while it's super specific to one kind of file…

Sources/SwiftGenKit/Parsers/FontsParser/CTFont.swift
```diff
@@ -0,0 +1,46 @@
...
+import PathKit
+
+extension PathKit.Path {
+  /// Returns the Path relative to a parent directory.
+  /// If the argument passed as parent isn't a prefix of `self`, returns `nil`.
+  ///
+  /// - Parameter parent: The parent Path to get the relative path against
+  /// - Returns: The relative Path, or nil if parent was not a parent dir of self
+  func relative(to parent: Path) -> Path? {
+    let parentComponents = parent.absolute().components
```
AliSoftware: Why is this method, which is quite applicable to any path and could even one day be moved in PathKit directly, is only located in CTFont?

Sources/SwiftGenKit/Parsers/StoryboardParser/StoryboardScene.swift
```diff
@@ -0,0 +1,92 @@
...
+
+    private static let tagTypeMap = [
+      "avPlayerViewController": "AVPlayerViewController",
+      "glkViewController": "GLKViewController",
+      "pagecontroller": "NSPageController"
+    ]
+
+    private static let tagModuleMap = [
+      "avPlayerViewController": "AVKit",
+      "glkViewController": "GLKit"
```
AliSoftware: Maybe we could find a way to avoid repeating magic strings like this in those two maps. Probably not worth a dedicated enum or constants but maybe with a single map with tuple values? (Could be for separate PR, sorry not to have caught that in the previous one)

Sources/SwiftGenKit/Parsers/AssetsCatalogParser/CatalogEntry.swift
djbe: Oh it wasn't an extension, I think it was a free function? Thought it was better to namespace it a bit (`private`, in `PathKit`). I'll move it into the `Entry` extension.

Sources/SwiftGenKit/Parsers/FontsParser/CTFont.swift
djbe: Same as above, it was a free function and I wanted to compartmentalise it a bit. Move it into CTFont?

Sources/SwiftGenKit/Parsers/StoryboardParser/StoryboardScene.swift
djbe: Will do. This is a "refactor" PR, so might as well do it here 🤷‍♂️

PR #418, opened 2018-04-29
Convert context testing format from PLIST to YAML

### LINKED ISSUE #399: Store contexts in JSON (or YAML)
Instead of storing as plists, output the contexts as JSON files, which will make it easer for interop such as #310.

djbe:
Fixes #399.

Old note:
While YAML might be more powerful and allowing us to have much more compact contexts (using anchors and aliases), such contexts would have to be made by hand, they can't be generated automatically (as they are right now).


AliSoftware:
Well we could generate YAML, by using YAMS which we already use for config file 😉

djbe:
Which is better though in this case? I don't have a preference either way.

djbe:
Keep in mind that we may want to open up this context interface at some point, and JSON is much easier to use in general than YAML (which often needs an extra library to support it).

djbe:
And I know that YAML is a superset of JSON 😉

djbe:
Actually, I just proved your point I think? 🤦‍♂️ 
I'll convert it to YAML files.

Issue #384, opened 2018-02-15
Rename `storyboards` command to `ib`


djbe:
See https://github.com/orgs/SwiftGen/teams/corecontributors/discussions/3 discussion.

We want to evolve SwiftGen to a system of parsers with multiple outputs for each parser. To eventually support xib files (which have quite a lot of common with storyboard files), we'll have to rename `storyboards` to `ib`.

Note: we could keep the `storyboards` command for now, wich would be an alias of `ib --filter '*.storyboards'`, see #383.

PR #419, opened 2018-04-29
Storyboards: split scenes and segues into separate templates

djbe:
With #364 and XIB file support incoming, we may want to split the templates up a little bit.

Todo:
- [ ] Document the new templates.
- [ ] Provide a migration guide.


djbe:
We don't? (honestly haven't tried it yet)
I'll look into it.

AliSoftware:
Nope iirc we support the same as the command line so one template and output for one input. What we do support is multiple invocations (with one {input, template, output}) for a given subcommand

djbe:
@AliSoftware seeing as I eventually want to add per-scene segue information (#364), should I rename the segue template in this PR to `simple-segues-swiftX`, and that PR adds a `complex-segues-swiftX`, or should that PR replace the existing (simple) segue functionality?

djbe:
On another note, multiple outputs are supported now, so that isn't a blocker for this anymore.

djbe:
An in-depth doc for what #364 does will be necessary, but that belongs in that PR.

I'm finishing up this PR now (docs, simplify the template a bit more, etc... As it's quite a simple change we should consider merging it for 6.0. The advanced segues in #384, which are not a breaking change, but additive, could be merged in 6.1 (or 7.0).

djbe:
@AliSoftware Is this GTM for you? It's one of the last PRs for 6.0 ([see here](https://github.com/SwiftGen/SwiftGen/pulls?q=is%3Aopen+is%3Apr+milestone%3A%22Swiftgen+6.0%22)) and #485 depends on it.

Documentation/templates/MigrationGuide.md
```diff
@@ -6,6 +6,7 @@
 * XCAssets #480: removed `noAllValues` parameter, replaced by `allValues` parameter.
 * XCAssets #482: removed deprecated `allValues` constant
 * XCAssets #482: changed x-platform image alias default from `Image` to `AssetImageTypeAlias`.
+* IB #419: Split up the templates into one for scenes and one for segues (in preparation for future templates for accessibility labels, ...).
```
AliSoftware: I guess that for now that part in the `MigrationGuide.md` is just a scratchpad, but in the final version we should clearly document the impact, especially for people using segues how they would need to modify their config files to have 2 { template, output } pairs instead of one (especially since in that breaking change version we also changed the config file format anyway)

Documentation/templates/ib/scenes-swift3.md
```diff
@@ -2,24 +2,23 @@
...
+| Invocation example | `swiftgen ib -t scenes-swift3 …` |
 | Language | Swift 3 |
 | Author | Olivier Halligon |
 
 ## When to use it
 
-- When you need to generate *Swift 3* code
-- The generated code supports both UIKit platforms (iOS, tvOS and watchOS) and AppKit platform (macOS)
+- When you need to generate *Swift 3* code for your storyboard *scenes*.
+- The generated code supports both UIKit platforms (iOS, tvOS and watchOS) and AppKit platform (macOS).
```
AliSoftware: Might be interesting to add a 3rd bullet point just informing that: "Note: if you also need to generate code for your storyboard segues, you can use [that other template] in addition to this one" ?

Documentation/templates/ib/scenes-swift4.md
```diff
@@ -2,24 +2,23 @@
...
+| Invocation example | `swiftgen ib -t scenes-swift4 …` |
 | Language | Swift 4 |
 | Author | Olivier Halligon |
 
 ## When to use it
 
-- When you need to generate *Swift 4* code
-- The generated code supports both UIKit platforms (iOS, tvOS and watchOS) and AppKit platform (macOS)
+- When you need to generate *Swift 4* code for your storyboard *scenes*.
+- The generated code supports both UIKit platforms (iOS, tvOS and watchOS) and AppKit platform (macOS).
```
AliSoftware: same suggestion for linking to the segue template for info

Documentation/templates/ib/segues-swift3.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Usage example
+
+```swift
+// You can perform segues using:
+vc.perform(segue: StoryboardSegue.Message.embed)
+
+// or match them (in prepareForSegue):
+override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
+  switch StoryboardSegue.Message(rawValue: segue.identifier!)! {
```
AliSoftware: I know there's your PR on improved segues pending, but wondering if it would be complicated to just add:

```swift
extension SegueType {
  init(_ segue: UIStoryboardSegue) {
    guard let identifier = segue.identifier else { fatalError("Segue has no identifier") }
    guard let value = self.init(rawValue: identifier) else { fatalError("Segue with identifier \(identifier) wasn't of the expected type \(Self)") }
  self = value
  }
}
```
So that at least the call site / usage example would have less force-unwraps, like this:
```
override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
  switch StoryboardSegue.Message(segue) {
  case .embed: …
  case .nonCustom: …
  }
}
```
After all, SwiftGen is all about avoiding string-based identifiers… but also about avoiding force-casts and force-unwraps at call site, so… :wink:

While waiting for your improved Segues PRs that would at least easily improve the situation in the meantime, without too much code to add in the current templates?

Tests/Fixtures/Generated/IB-iOS/segues-swift3-context-all-ignore-module-need-extra-definitions.swift
```diff
@@ -0,0 +1,44 @@
...
+// swiftlint:enable explicit_type_interface identifier_name line_length type_body_length type_name
+
+// MARK: - Implementation Details
+
+internal protocol SegueType: RawRepresentable { }
+
+internal extension UIViewController {
+  func perform<S: SegueType>(segue: S, sender: Any? = nil) where S.RawValue == String {
+    performSegue(withIdentifier: segue.rawValue, sender: sender)
+  }
```
AliSoftware: Side note for another PR: I just realized that with our new possibility of generating multiple outputs from one single input — actually even now with the ability to parse multiple inputs, the problem might arise too — there's a possibility that a user would prefer generating multiple outputs declaring the same `protocol SegueType` and `extension UIViewController` etc… leading to compilation errors for multiple declarations.

So maybe for the next release we should think about a mechanism in our templates to ensure there's no repetition of our "Implementation Details" of the same template used multiple times. Should probably not be automatic (because depending on how the user will use the multiple output files they generate, maybe on the same target maybe not, they might want to have that implementation repeated on each output or not), but would probably be either via a `param` like `omitImplementationDeclarations` or `omitBoilerplate` or something, or having the Implementation details be declared in a separate, dedicated template (might be cleaner at implementation site with clearly separating that from the reusable templates, but that would require a lot of configuration in the `swiftgen.yml` file from the end user, so not cool. Or maybe using a `{% include %}` directive we could keep the boilerplate in a separate template for maintenance but conditionally auto-include it in the generated code depending on a `param`? Anyway, food for thoughts.

Documentation/templates/MigrationGuide.md
djbe: Yeah, exactly. I'm trying to just collect all the necessary info to write nice docs later (this WE?)

Documentation/templates/ib/scenes-swift3.md
djbe: Good idea, will do.

Documentation/templates/ib/segues-swift3.md
djbe: Oooh, I would not do that, that's a bad idea. This would break SwiftGen segues for a lot of people. You can't assume that all your segues have names, that's exactly why I have the extra `unnamedSegue` in the other PR.

I know that this would break quite a few of the apps we have here, because sometimes we name all the segues, but sometimes we just try to cast the destination to a type and set properties (for example). In those cases, segues don't always have names. I'm not discussing if all segues **must** have names, just saying that they don't always have a name. 

What we could do, but TBH that would belong in a separate PR (because this is just the split, no template modifications), is cherry-pick the `unnamedSegue` improvement from the other PR, and then we can do your idea (without throwing `fatalError`s for a whole bunch of users).

Tests/Fixtures/Generated/IB-iOS/segues-swift3-context-all-ignore-module-need-extra-definitions.swift
djbe: Good idea, too bad we can't use the old ObjC trick to check if a macro is defined 😢 

I particularly like the `omitImplementationDeclarations`, but it's probably something for 6.1.

Documentation/templates/ib/segues-swift3.md
djbe: Yes actually, this is what I currently use:
```swift
switch StoryboardSegue.ProductDetails(rawValue: segue.identifier ?? "") {
case .infoHeader?:
	... stuff ...
default:
	break
}
```

If we add the `unnamedSegue` case, and your first `init` (lightly modified), it becomes:
```swift
extension SegueType {
  init(_ segue: UIStoryboardSegue) {
    guard let identifier = segue.identifier else { return .unnamedSegue }
    guard let value = self.init(rawValue: identifier) else { fatalError("Segue with identifier \(identifier) wasn't of the expected type \(Self)") }
    return value
  }
}

switch StoryboardSegue.ProductDetails(segue) {
case .infoHeader:
	... stuff ...
case .unnamedSegue: // or default
	break
}
```

AliSoftware: We then should only generate the `unnamedSegue` case only of there exist such a segue in the storyboard. If all segues have an identifier then that extra case wouldn't make sense.

Maybe that's already what you do in your other PR idk. In the meantime, returning `nil` instead of `.unnamedSegue` for now (maybe until we re-discuss this in your PR for specific cases?) in the guard would feel more logical and simple for this PR until we can be sure, probably via your PR i the future, that unnamedSegue is only generated if such an unnamed segue exists.

djbe: Yeah #364 makes the necessary modifications so that unnamed segues are also in the context instead of being filtered out. The templates are also modified to take this into account. In that sense #364 is a breaking change BTW.

To clarify, given what this PR has now, the only thing we can provide is the following `init`:

AliSoftware: Cool. Let's do that then 😉
(Can't wait for #364 btw)

djbe: Yeah, I discovered yesterday that there are ...issues... when we combine #364 with #485, specifically, the parser can't build up the correct `custom types` set because some classes/modules aren't known at parse time. But that discussion should go in the other PR.

@AliSoftware Quick question though: should I cherry-pick the "don't filter out unnamed segues" out of #364 into a separate PR, that can be merged in 6.0, so we don't have a breaking change for #485?

AliSoftware: Yeah, probably, especially if that `init` will then "support" them by returning `nil`.

As of today, people do use them as such anyway: if they know they have unnamed segues, they'll `guard let id = sender.identifier` manually, and if they don't (or are careless) they'll `StoryboardSegue.Message(sender.identifier!)`, but that's their choice of crashing and not accounting for unnamed segues then, and with not filtering the unnamed segues anymore that `init` will at least give them a nicer solution to avoid a crash.

Documentation/templates/ib/segues-swift3.md
```diff
@@ -48,10 +49,10 @@ vc.perform(segue: StoryboardSegue.Message.embed)
...
 override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
-  switch StoryboardSegue.Message(rawValue: segue.identifier!)! {
-  case .embed:
+  switch StoryboardSegue.Message(segue) {
+  case .embed?:
     // Prepare for your custom segue transition
-  case .nonCustom:
+  case .nonCustom?:
     // Pass in information to the destination View Controller
   }
```
AliSoftware: That won't compile without adding the last missing `case nil` ("switch must be exhaustive" error)

Documentation/templates/ib/segues-swift4.md
```diff
@@ -48,10 +49,10 @@ vc.perform(segue: StoryboardSegue.Message.embed)
...
 // or match them (in prepareForSegue):
 override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
-  switch StoryboardSegue.Message(rawValue: segue.identifier!)! {
-  case .embed:
+  switch StoryboardSegue.Message(segue) {
+  case .embed?:
     // Prepare for your custom segue transition
-  case .nonCustom:
+  case .nonCustom?:
     // Pass in information to the destination View Controller
```
AliSoftware: Same

README.md
```diff
@@ -427,10 +432,10 @@ vc.perform(segue: StoryboardSegue.Message.embed)
...
 // or match them (in prepareForSegue):
 override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
-  switch StoryboardSegue.Message(rawValue: segue.identifier!)! {
-  case .embed:
+  switch StoryboardSegue.Message(segue) {
+  case .embed?:
     // Prepare for your custom segue transition
-  case .nonCustom:
+  case .nonCustom?:
     // Pass in information to the destination View Controller
```
AliSoftware: same

Tests/Fixtures/Generated/IB-iOS/segues-swift3-context-all-customname.swift
```diff
@@ -34,12 +34,19 @@ internal enum XCTStoryboardCustom {
...
 internal extension UIViewController {
   func perform<S: SegueType>(segue: S, sender: Any? = nil) where S.RawValue == String {
     performSegue(withIdentifier: segue.rawValue, sender: sender)
   }
 }
 
+internal extension SegueType {
+  init?(_ segue: UIStoryboardSegue) {
+    guard let id = segue.identifier else { return nil }
+    return self(rawValue: id)
```
AliSoftware: Did you check if this code compiled? I never remember if we can `return` anything other than `nil` from a `init?`, aren't we forced to do `self = self(rawValue: id)` instead?

(I guess our tests will catch that on CI if it doesn't compile…)

Documentation/templates/ib/segues-swift3.md
djbe: Maybe better a `default` case, because our current segues implementation has all the segues in a storyboard, not just the ones starting from 1 segue.

AliSoftware: yeah, but if that method is called with a segue from another scene, then `StoryboardSegue.Message(segueFromAnotherScene)` will also return `nil`, not being able to construct a `StoryboardSegue.Message` instance with `rawValue = segueFromAnotherScene.identifier`, so… `nil` should work here too

tbh I prefer avoiding `default` when possible, so that if later on the user adds another `segue` from that `Message` scene, the `switch` will warn them that they forgot a `.some(StoryboardScene.Message)` case in there, while a `default` would absorb both segue from another scene but also a missing segue starting from `Message` and forgotten in the `switch`.

djbe: I think you're mixing up PRs? This is the PR to split scenes and segues templates, not the scene-specific segues.

This PR's is our old's way of generating segue enums, which is just one big enum for all the segues in a storyboard. Our current implementation can not help a user if they try to use a case for an enum that can't happen for this scene, or init an enum from a storyboard where some segues happen to have the same name.

With our current segues, the only thing a user can do is use `default`, because there will be a bunch of segues that won't matter for this scene, as they can never happen.

AliSoftware: Mmmh maybe I am (tired), but I still don't think `default` would be the right choice?

`default` in your case will both cover `.some(_)` — meaning the `segue` passed will be a valid and existing `StoryboardSegue.Message`, so one of all the segues staring from that `Message` scene but that you have forgotten to cover in your `switch`, right? — and `.none` — meaning a `segue` coming from another scene, or unnamed.

All named segues starting from the scene named `Message` should be handled in the `prepare(for segue: UIStoryboardSegue)` method and thus covered as a dedicated `case` in the `switch` in my opinion in a classic app, right? And if you have a `case` in your `switch StoryboardScene.Message(segue)` for all `cases` listed in the `StoryboardScene.Message` enum, then the only remaining case — that will indeed make the `switch` exhaustive then — would be `case nil`, which will happen either if `segue.identifier` is `nil`, or if the `segue` isn't actually one of the cases of the `StoryboardScene.Message` enum.

So my point is: 

1. Using `case nil:`  here is enough to make the `switch` exhaustive, as both a segue from another scene (and thus an `identifier` not listed in the `cases` of the `enum`)  AND an unnamed segue will make `StoryboardScene.Message(segue)` return `nil`
2. If later on the user adds a named segue starting from the `Message` scene, then `switch StoryboardScene.Message(segue)` will have to explicitly add that `case .theNewSegueIdentifier` as `case nil` won't cover it — and that's good, as if enforces not to forget it — while `case default:` wouldn't enforce anything and let the user forget about handling the case for that new segue in their `switch`.

djbe: I completely agree that, when possible, exhaustive enums are the way to go. But:
> a segue from another scene (and thus an identifier not listed in the cases of the enum) ... return nil

Again, this is in the other PR, #364. Our current segue enums, **don't** have this information. Our current segue information is for a **whole** storyboard. My frustration with the current situation is what led to me writing #364.

With our current segue info, let's say you have a scene A with 5 segues. But in that storyboard, there are another 15 segues unrelated to that scene (about average case, not too unthinkable). You can't expect users to, for every scene in that storyboard, write a completely exhaustive `switch` statement. Most users just write switches with the cases for that particular scene, and ignore the rest --> `default`.

Does this prevent errors if the user adds more segues to a scene? No. But exhaustively handling every case of a whole storyboard in every scene isn't better. I'd argue it's worse.

djbe:
Man, that took way too many rebases at the end. Note to self: run the output compile tests locally 😅

PR #420, opened 2018-04-29
Remove Swift 2 support

djbe:
About 'damn time! 😄 

I don't know to what extent we should further document the removal of this. Maybe something extra for the release notes?


djbe:
That was in the docs for those templates:
https://github.com/SwiftGen/SwiftGen/pull/420/files#diff-829706d01268395950998d2821737035L14

The only remaining mention of Swift 2 is in our migration guide (where we'll need a new entry for this). But I'm pretty sure there will be more changes that'll warrant an entry there, so might be better to bundle that in a separate PR.

PR #422, opened 2018-04-30
Internal: namespace parsers

djbe:
Didn't think to do this in the refactor PR (#417), but this PR goes along with it.

I'll make sure to rebase/udpate other existing PRs where possible.


djbe:
From what I'm seeing in #411, everything still works (in as far as SPM can handle our project). SPM breaks in other parts of our code, but that's already covered in that PR.

PR #423, opened 2018-04-30
Rename `storyboards` to `ib`

### LINKED ISSUE #384: Rename `storyboards` command to `ib`

djbe:
Fixes #384.

Note: this breaks the homebrew formula, so we shouldn't forget to update it in a separate PR.


CHANGELOG.md
```diff
@@ -22,6 +22,9 @@
 * Remove Swift 2 support.  
   [David Jennes](https://github.com/djbe) 
   [#420](https://github.com/SwiftGen/SwiftGen/pull/420)
+* Renamed the `storyboards` command to `ib`, to better reflect it's purpose. An alias for `storyboards` still exists, but it will be removed at some point.
```
AliSoftware: Idk about that, we might keep `storyboard` as an alias for `swiftgen ib --filter '*.storyboard'` if/when we implement filtering files later?

djbe:
TIL. Good idea!

CHANGELOG.md
djbe: Eh, maybe. The way I see SwiftGen 6 developing right now, even the templates will be split up, so I'm not sure how feasible it'll be to actually keep the `storyboards` command compatible.

The alias you mention might be a good idea though, something with a predefined filter. Continuing that train of thought: `xib` for `ib --filter '*.xib'`, and others I can't think of right now.

PR #424, opened 2018-05-05
Config: add support for multiple outputs

djbe:
This adds support for multiple outputs for an invocation of a command, with the same input files; as described in https://github.com/SwiftGen/SwiftGen/pull/419#issuecomment-385282924. 

For example:
```yaml
input_dir: Fixtures/
output_dir: Generated/
ib:
  paths: IB-iOS
  outputs:
    - templateName: scenes-swift4
      output: ib-scenes.swift
    - templateName: segues-swift4
      output: ib-segues.swift
```

Note: this PR is fully backwards compatible. If no `outputs` is found, the config parser will fall back to parsing the old `output` (with `templateName` and `templatePath`).


Documentation/ConfigFile.md
```diff
@@ -15,19 +15,22 @@ input_dir: Sources/Resources
...
-  templateName: structured-swift3
-  output: L10n-Constants.swift
+  outputs:
+    - templateName: structured-swift3
+      output: L10n-Constants.swift
 xcassets:
   - paths: Logos.xcassets
-    templateName: swift3
-    output: Logos-Constants.swift
     params:
```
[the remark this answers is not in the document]
AliSoftware: e.g.

```yaml
xcassets:
  inputs: Logos.xcassets
  outputs:
   - templateName: swift3
     output: Logos-Constants.swift
     params: &assetParams
       enumName: Logos
   - templateName: swift3-listall
     output: Logos-Constants+All.swift
     params: *assetParams
   - templateName: html
     output: Logos-List.html
     params:
       columns: 4
       size: 200
```

CHANGELOG.md
```diff
@@ -31,7 +31,7 @@
 * Add ability to list all custom fonts and register them using `FontFamily.registerAllCustomFonts`.  
   [Olivier Halligon](https://github.com/AliSoftware)
   [#394](https://github.com/SwiftGen/SwiftGen/issues/394)
-* The `swiftgen.yml` config file now accepts multiple outputs for each command, allowing you to generate multiple outputs from the same files and content. This also means that the `output` parameter is now deprecated, in favour of the `outputs` parameter, and it may be removed in a future version of SwiftGen.  
+* The `swiftgen.yml` config file now accepts multiple outputs for each command, allowing you to generate multiple outputs from the same files and content. This also means that the `output` parameter is now deprecated, in favour of the `outputs` parameter, and it may be removed in a future version of SwiftGen. Similarly, the `paths` parameter has been renamed to `inputs` for consitency.
```
AliSoftware: s/consitency/consistency/

Sources/SwiftGen/Config/ConfigEntry.swift
```diff
@@ -38,40 +40,32 @@ extension Config {
 
 extension Config.Entry {
   init(yaml: [String: Any]) throws {
```
AliSoftware: Not sure if a change like this is worth doing in this current PR or not, but wondering if those `yaml:` labels were a good choice?

I mean, that configuration content could come in any form in the future, hell one might already use JSON if they wish, and maybe one day we would support any other format capable of serialising Hash+Arrays+Strings structures (Plist, whatever), all that to say that there's no real reason in those inits to name the label `yaml:` rather than anything else (I know, I personally use `init(json: )` constructors in my own projects when I know the data come from a REST/JSON web service, so guilty as charged here too, but worth opening the discussion anyway)

Nothing urgent, no impact on the behavior, and as I said could be an internal refactor later, but at least we could start discussing and wondering if it's worth it and find a new name, or too much nitpicking and bikeshedding ^^

Sources/SwiftGen/Config/ConfigEntry.swift
```diff
@@ -13,19 +13,21 @@ import PathKit
 extension Config {
   struct Entry {
     enum Keys {
+      static let inputs = "inputs"
       static let outputs = "outputs"
-      static let paths = "paths"
 
       // Legacy: remove this once we stop supporting the output key at subcommand level
       static let output = "output"
+      // Legacy: remove this once we sto supporting the old paths key (replaced by inputs)
```
AliSoftware: s/sto/stop/

Sources/SwiftGen/Config/ConfigEntry.swift
```diff
@@ -84,14 +78,24 @@ extension Config.Entry {
...
     return typedValue
   }
+
+  static func parseValueOrArray<T, U>(yaml: Any, key: String? = nil, parseValue: (T) throws -> U) throws -> [U] {
+    if let value = yaml as? T {
+      return [try parseValue(value)]
+    } else if let value = yaml as? [T] {
+      return try value.map { try parseValue($0) }
+    } else {
+      throw Config.Error.wrongType(key: key, expected: "\(T.self) or Array of \(T.self)", got: type(of: yaml))
```
AliSoftware: I know I'm the one who wrote that code in a previous command, but I'm not sure how the error will be displayed in the console when we use `(T.self)` in string interpolation, does it read nicely? I didn't actually test it

Tests/SwiftGenTests/ConfigLintTests.swift
```diff
@@ -114,31 +114,33 @@ class ConfigLintTests: XCTestCase {
...
       assertionMessage: "Linter should warn when 'output' key is missing"
     )
   }
 
   func testLintInvalidStructure() {
     _testLint(
       fixture: "config-invalid-structure",
-      expectedLogs: [(.error, "Wrong type for key strings.paths: expected Path or array of Paths, got Array<Any>.")],
+      expectedLogs: [(.error, """
+        Wrong type for key strings.inputs: expected String or Array of String, got Array<Any>.
```
AliSoftware: Cool I guess that answers my question above, displays nicer than I feared 👍

CHANGELOG.md
AliSoftware: Also, could be nice to remind people here that they could use `swiftgen config lint` to validate their config files and see the list of deprecations and migrations to do on their own config file :wink:

Sources/SwiftGen/Config/ConfigEntry.swift
djbe: I usually use `data`, but that's even less descriptive 🤷‍♂️

AliSoftware: All our inits similar to this take dictionaries as parameters, right (never arrays, always dicts)? So maybe `init(dictionary: )` would fit better?

_In Foundation/CoreGraphics Apple uses `CGPoint(dictionaryRepresentation: )` and the likes, so at least `Config.Entry(dictionary: ))` would be inspired by/similar to those…_

djbe:
The deprecation warnings are going to be a bit difficult?
- During linting, the config has been read and mapped to the new structure, so it won't be able to check what old keys were used.
- During read, we don't have any logger available.

AliSoftware:
Ah. Good point. Damn, I still feel like this would be very useful to have, especially for ease of migration for users, and because a lot of people upgrade without reading CHANGELOGs and would not be aware of the changes otherwise.

Maybe dedicate another PR to add a logger to the Config parser to be able to print those?

djbe:
A separate issue & PR might be a good idea. We need to discuss other migrations that'll need to happen, such as `storyboards` --> `ib`, `storyboards` --> `scenes` and `segues` templates, etc...

djbe:
> I've seen code to fallback to output when outputs.output isn't available, but did I miss the similar fallback for template[Name|Path] when outputs.template[Name|Path] isn't present too? (maybe I did, it's late and I'm also only reviewing a partial diff, so…)

Is that needed though? `output` is the only mandatory one. It'll already throw an error if you use the new `outputs`, but don't provide `templatePath|Name` at that level.

I think it should provide warnings if it encounters those old keys, especially during linting.

AliSoftware:
> I think it should provide warnings if it encounters those old keys, especially during linting.

Ok, seems fair, would probably go in the dedicated PR about depreciations warnings when reading the Config file anyway, where we could discuss all the possible use cases of unaware users mixing them up in different manners :wink:

djbe:
Allright, the latest commit contains a basic deprecation logger. We can extend it if needed in another PR, and see there if we've covered our bases.

AliSoftware:
🎉

PR #440, opened 2018-06-07
[WIP] Use swiftlint:disable all

### LINKED ISSUE #409: Use swiftlint:disable all
SwiftGen Version: 5.3.0
Xcode Version: 9.3

I suggest using `// swiftlint:disable all` for all files generated by SwiftGen instead of single `disable` commands:

```
// swiftlint:disable superfluous_disable_command
// swiftlint:disable file_length
```

The option to do this is first introduced in SwiftLint 0.25.1 and allows for also ignoring custom, user-defined rules in SwiftGen-generated files. Previously, the user had to manually exclude SwiftGen files, if they had created a custom rule not being respected by SwiftGen.

fredpi:
Here's my implementation to close #409.

As intended, the outcome for SwiftGen users is that they don't have to exclude SwiftGen-generated files from listing manually.

Meanwhile, SwiftGen internally still tests against generated files without a general `swiftlint:disable all`statement in addition to files with this general statement. Those two tests ensure that the feature works for the user, while the project is still able to verify the code quality. To keep those two file versions in sync, a test also ensures that the plain content (without swiftlint statements and blank lines) is the same for each file version.


Tests/Fixtures/Generated/AvoidDisableAll/IB-iOS/swift3-context-all-customname.swift
```diff
@@ -11,7 +11,6 @@ import UIKit
 
 // swiftlint:disable superfluous_disable_command
 // swiftlint:disable file_length
-
```
djbe: This change is not necessary.

Tests/Fixtures/Generated/DisableAll/Colors/literals-swift3-context-defaults-customname.swift
```diff
@@ -0,0 +1,21 @@
+// Generated using SwiftGen, by O.Halligon — https://github.com/SwiftGen/SwiftGen
```
djbe: The moved files and these new dupes are not needed. Use the existing variations mechanism.

Tests/TemplatesTests/TestsHelper.swift
```diff
@@ -116,8 +116,9 @@ class Fixtures {
     return string(for: name, subDirectory: "templates/\(sub.rawValue.lowercased())")
```
djbe: All of these changes can be removed, they are not needed. Use the existing variations mechanism.

templates/colors/literals-swift3.stencil
```diff
@@ -11,10 +15,12 @@
   {% if enumName != 'UIColor' %}{{accessModifier}} enum {{enumName}} { }{% endif %}
 #endif
 
+{% if avoidSwiftLintDisableAll %}
```
djbe: It's annoying that the templates will get noisier with these checks. But I'm not sure if there's a better way to do this.

We could keep the unneeded `// swiftlint:disable specific rule`, and only have the initial check to add the `disable all` line.

djbe:
Forgot to mention in my review: new parameters such as `avoidSwiftLintDisableAll` need to be documented in the template docs.

fredpi:
@djbe Thanks for the fast review!

I'm also shocked a bit about the size of the PR and recognise there are quite a few things that aren't needed and I will definitely change them.

However, I also want some clarification about the detailed design of the actual feature. I currently can't see any better variation mechanism than the custom one I used (that's why I implemented it this way). Correct me if I'm wrong about any of these:
- We need to test both versions of the files. One is for checking whether SwiftLint passes, the other is for checking that the version delivered to the user works.
- Currently, some variation is there for Swift3 and Swift4 versions of the generated files. This is realised by using different templates at all.
- Other variations are about representing different contexts. Those are realised by applying different contexts to the same template.
- Using two versions depending on whether `swiftlint:disable all` is used, multiplies the number of files by two. A decision has to be made whether the copy of a file belongs into the same folder as the other version of the file. As the two file versions target different goals (one for checking whether the file is in-line with SwiftLint, the other for checking whether the file works for the end user), I prefer keeping them in two separate folders at all.

What variation mechanism should I use in your opinion and why?

AliSoftware:
Random idea: what about stripping the disable command during our own lint?

That is: avoid duplicating the templates and generated files altogether, having the `swiftlint:disable all` in the templates and thus also in the generated files for the end user… but in our script running SwiftLint on the generated files, instead of running it on those files directly, run it on a version of those generated files where we used `sed s/swiftlint:disable all//` to remove that?

Something like `sed s/swiftlint:disable all// Generated/xxx.swift | swiftlint --stdin …` on each generated file maybe? (Didn't test, I'm writing this from my phone). Would avoid maintaining two variations and copies of all outputs just for this…

fredpi:
@AliSoftware Wow, that's a nice idea. Will definitely try it out!

fredpi:
@AliSoftware I couldn't implement it the exact way you proposed, as the generated result must still match the fixture, but I found another similar solution that looks quite good to me. See #442.

PR #454, opened 2018-07-20
Add parameter for xcasset namespacing configuration 

### LINKED ISSUE #453: xcassets ignores "Provides Namespace"
According to documentation, xcassets templates "take into account any namespacing folder in your Assets Catalogs". Yet no matter what value this option is, generated code is always the same.

SwiftGen
```swift
/// with "Provides Namespaces" unchecked on "Member" folder
let image = Asset.Member.profilSupp.image

/// with "Provides Namespaces" checked  on "Member" folder
let image = Asset.Member.profilSupp.image
```

R.swift

```swift
/// with "Provides Namespaces" unchecked on "Member" folder
let image = R.image.profil_supp

/// with "Provides Namespaces" checked on "Member" folder
let image = R.image.member.profil_supp
```

Tested with SwiftGen 5.3.0.

pjechris:
Resolves #453 by adding a new parameter called "forceProvidesNamespaces".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,9 @@
 * Renamed the `storyboards` command to `ib`, to better reflect it's purpose. An alias for `storyboards` still exists, but it will be removed at some point.  
   [David Jennes](https://github.com/djbe) 
   [#423](https://github.com/SwiftGen/SwiftGen/pull/423)
+* XCAssets: the generated templates won't namespace groups by default anymore, use the `forceProvidesNamespaces` flag to enable this behaviour again.  
+  [jechris](https://github.com/pjechris)
+  [#453](https://github.com/SwiftGen/SwiftGen/issues/453)
 
 ### New Features
 
--- a/Documentation/templates/xcassets/swift3.md
+++ b/Documentation/templates/xcassets/swift3.md
@@ -26,6 +26,7 @@ You can customize some elements of this template by overriding the following par
 | `imageTypeName` | `ImageAsset` | Allows you to change the name of the struct type representing an image. |
 | `noAllValues` | N/A | Setting this parameter will disable generation of the `allColors` and `allImages` constants. |
 | `publicAccess` | N/A | If set, the generated constants will be marked as `public`. Otherwise, they'll be declared `internal`. |
+| `forceProvidesNamespaces` | `false` | If set to true, generates namespaces even for non namespacing asset folders (i.e. "Provides Namespace" is unchecked) |
 
 ## Generated Code
 
--- a/Documentation/templates/xcassets/swift4.md
+++ b/Documentation/templates/xcassets/swift4.md
@@ -26,6 +26,7 @@ You can customize some elements of this template by overriding the following par
 | `imageTypeName` | `ImageAsset` | Allows you to change the name of the struct type representing an image. |
 | `noAllValues` | N/A | Setting this parameter will disable generation of the `allColors` and `allImages` constants. |
 | `publicAccess` | N/A | If set, the generated constants will be marked as `public`. Otherwise, they'll be declared `internal`. |
+| `forceProvidesNamespaces` | `false` | If set to true, generates namespaces even for non namespacing asset folders (i.e. "Provides Namespace" is unchecked) |
 
 ## Generated Code
 
--- a/Sources/SwiftGenKit/Stencil/AssetsCatalogContext.swift
+++ b/Sources/SwiftGenKit/Stencil/AssetsCatalogContext.swift
@@ -42,9 +42,10 @@ extension AssetsCatalog.Parser {
   private func structure(entries: [AssetsCatalog.Entry]) -> [[String: Any]] {
     return entries.map { entry in
       switch entry {
-      case .group(let name, let items):
+      case .group(let name, let isNamespaced, let items):
         return [
           "type": "group",
+          "isNamespaced": "\(isNamespaced)",
           "name": name,
           "items": structure(entries: items)
         ]
--- a/Sources/SwiftGenKit/Parsers/AssetsCatalog/CatalogEntry.swift
+++ b/Sources/SwiftGenKit/Parsers/AssetsCatalog/CatalogEntry.swift
@@ -9,7 +9,7 @@ import PathKit
 
 extension AssetsCatalog {
   enum Entry {
-    case group(name: String, items: [Entry])
+    case group(name: String, isNamespaced: Bool, items: [Entry])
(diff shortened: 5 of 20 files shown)


CHANGELOG.md
```diff
@@ -32,6 +32,10 @@
 
 ### New Features
 
+* Add `forceProvidesNamespaces` parameter to `xcassets` command.  
+  [jechris](https://github.com/pjechris)
+  [#453](https://github.com/SwiftGen/SwiftGen/issues/453)
+
```
djbe: No need for this empty line.

Documentation/templates/xcassets/swift3.md
```diff
@@ -26,6 +26,8 @@ You can customize some elements of this template by overriding the following par
 | `imageTypeName` | `ImageAsset` | Allows you to change the name of the struct type representing an image. |
 | `noAllValues` | N/A | Setting this parameter will disable generation of the `allColors` and `allImages` constants. |
 | `publicAccess` | N/A | If set, the generated constants will be marked as `public`. Otherwise, they'll be declared `internal`. |
+| `forceProvidesNamespaces` | true | If set to true, generates namespaces even for non namespacing asset folders (i.e. "Provides Namespace" is unchecked) |
+|
```
djbe: Why is this extra pipe char. here?

Documentation/templates/xcassets/swift4.md
```diff
@@ -26,6 +26,8 @@ You can customize some elements of this template by overriding the following par
 | `imageTypeName` | `ImageAsset` | Allows you to change the name of the struct type representing an image. |
 | `noAllValues` | N/A | Setting this parameter will disable generation of the `allColors` and `allImages` constants. |
 | `publicAccess` | N/A | If set, the generated constants will be marked as `public`. Otherwise, they'll be declared `internal`. |
+| `forceProvidesNamespaces` | true | If set to true, generates namespaces even for non namespacing asset folders (i.e. "Provides Namespace" is unchecked) |
+|
```
djbe: Same as above.

Documentation/templates/xcassets/swift3.md
```diff
@@ -26,6 +26,8 @@ You can customize some elements of this template by overriding the following par
 | `imageTypeName` | `ImageAsset` | Allows you to change the name of the struct type representing an image. |
 | `noAllValues` | N/A | Setting this parameter will disable generation of the `allColors` and `allImages` constants. |
 | `publicAccess` | N/A | If set, the generated constants will be marked as `public`. Otherwise, they'll be declared `internal`. |
+| `forceProvidesNamespaces` | true | If set to true, generates namespaces even for non namespacing asset folders (i.e. "Provides Namespace" is unchecked) |
```
djbe: Do we want to change the default to `false` while we're adding this functionality?

Sources/SwiftGenKit/Stencil/AssetsCatalogContext.swift
```diff
@@ -42,9 +42,10 @@ extension AssetsCatalog.Parser {
   private func structure(entries: [AssetsCatalog.Entry]) -> [[String: Any]] {
     return entries.map { entry in
       switch entry {
-      case .group(let name, let items):
+      case .group(let name, let isNamespaced, let items):
         return [
           "type": "group",
+          "isNamespaced": "\(isNamespaced)",
```
djbe: Stencil can handle booleans, no need to stringify the value here.

Tests/Fixtures/Generated/XCAssets/swift4-context-all-force-namespaces-false.swift
```diff
@@ -0,0 +1,134 @@
...
+    }
+    internal static let `private` = ImageAsset(name: "private")
+    // swiftlint:disable trailing_comma
+    internal static let allColors: [ColorAsset] = [
+    ]
+    internal static let allImages: [ImageAsset] = [
+      Exotic.banana,
+      Exotic.mango,
+      Round.apricot,
+      Round.Red.apple,
```
djbe: This fails to compile because there's no `Red` sub-enum. You still need to adapt the `allCases` bit of code in the template.

Sources/SwiftGenKit/Stencil/AssetsCatalogContext.swift
pjechris: It's harder than it seems.
It does work with Stencil but it breaks Yaml deserialization tests. Because tests are using Yaml `.basic` (so everything is string). So true would never equal "true".
So basically I had to use strings to ensure tests are passing :(

djbe: Hmmm, that's not the first time we've encountered issues with YAML deserialization. Lemme check

Tests/Fixtures/Generated/XCAssets/swift4-context-all-force-namespaces-false.swift
pjechris: Didn't see that one, i'll fix it

Documentation/templates/xcassets/swift3.md
pjechris: Good question. Changing it to `false` would be a breaking change and might alter users generated code. However it would be more logical :)

djbe: I know, and we are heading towards a breaking release (6.0), so this might be the time to correct the behaviour. @AliSoftware thoughts?

Sources/SwiftGenKit/Stencil/AssetsCatalogContext.swift
djbe: We could switch to non-basic parsing, but that would break one of our `strings` unit tests where a string key has as value "no":
https://github.com/SwiftGen/SwiftGen/blob/master/Tests/Fixtures/StencilContexts/Strings/multiple.yaml#L37

Ideally Yams'd have a mode to write quotes around strings, but not annotate all the other types (which is what `canonical` does I think? Can't remember, it's been a while).

Documentation/templates/xcassets/swift3.md
AliSoftware: Since the next release will be a breaking one anyway, I'm all for fixing this now and set it to false by default 👍
Especially since thanks to this PR people migrating to 6.0 would still be able to restore the old behavior with this param 😉

AliSoftware: Side note: not sure if we started a migration guide for 6.0? (if not could be a nice occasion)

PR #445, opened 2018-06-29
Add support for NSDataAsset

orj:
Resolves #444 

* Added appropriate CatalogEntry
* Updated unit tests
* Updated templates

PR #456, opened 2018-07-21
Reorder generated code for xcassets

AliSoftware:
So that all the static code is grouped on top
And all the generated constants are at the bottom

Makes it way easier to find the list of asset constants at first sight
when opening the generated file to lookup images & colors

---

⚠️ Can conflict with https://github.com/SwiftGen/SwiftGen/pull/445. If we merge #445 before, we would need to reorder the constants for the data assets at the bottom too. Otherwise, might be better to rebase #445 after merging this one, and merge #445 next.


djbe:
Maybe adapt the order of the static stuff to be alphabetical? First color-related types, then image related types? And add data-related types in between (in #445).

AliSoftware:
@djbe fixed :)

PR #461, opened 2018-07-29
Fix Yams dump for ambiguous strings

djbe:
Addresses an issue with our YAML contexts encountered in #454 and #455, due to how the default Yams `dump` function will omit quotes for ambiguous strings.

This is only a temporary fix until https://github.com/jpsim/Yams/issues/133 provides a better way to address this.


djbe:
Note: we can't just match on some predefined strings such as "true", "false", "on", "off", etc... There's actually quite a bit of logic behind the string parsers in Yams:
https://github.com/jpsim/Yams/blob/master/Sources/Yams/Resolver.swift#L108-L149

Tests/SwiftGenKitTests/TestsHelper.swift
```diff
@@ -187,11 +187,34 @@ class Fixtures {
...
+// MARK: - Temporary fix until Yams fixes string dumping
+
+enum YamsFix {
+  static func represent(_ any: Any) throws -> Node {
+    switch any {
+    case let string as String:
+      // if the tag after resolving is not "str", then this string is ambiguous
+      let tag = Resolver.`default`.resolveTag(of: Node(string))
+      let style: Node.Scalar.Style = (tag != .str) ? .doubleQuoted : .any
+      return Node(string, .implicit, style)
```
AliSoftware: I wonder if this test is worth it (as oppose to dump all strings as double quoted, regardless if they're ambiguous or not)?

Tests/SwiftGenKitTests/TestsHelper.swift
```diff
@@ -136,7 +136,7 @@ func XCTDiffContexts(_ result: [String: Any],
   if ProcessInfo().environment["GENERATE_CONTEXTS"] == "YES" {
     let target = Path(#file).parent().parent() + "Fixtures/StencilContexts" + directory.rawValue + fileName
     do {
-      let data = try Yams.dump(object: result)
+      let data = try Yams.serialize(node: YamsFix.represent(result))
```
AliSoftware: Instead of just extracting the represent fix in a dedicated enum maybe we could have a tiny wrapper around Yams — would especially be useful when we'll need it in the main target later to allow end users to output YAML from SwiftGen (eg the YAML/JSON PR)?

Like:
```swift
enum YAML {
  func decode(data: Data) throws -> Any {
    // Simple forwarding for now, but maybe
    // later well use a custom Construc
    // to parse only `true`/`false` as Booleans
    // And/or to parse env variables like SwiftLint does etc
    return Yams.load(data: data, .basiit
  }
  func encode(object: Any) throws -> Data {
    return Yams.serialize(fixedRepresent(object))
  }
  private fixedRepresent(…) …
}
```

I know this is intended to only be a temporary fix but extracting the  whole YAML load/dump in a dedicated type in SwiftGen will make it easier to generalize  and uniformize our YAML parsing/dumping configurations/rules across SwiftGen, both for the test contexts, for the config file (not much affected by this dumping fix but still) and future PRs.

AliSoftware: And that would make changing the fix to the official solution (once implemented in YAMS) easier to do as all load/dump code invoking YAMS will be in one place 😉

Tests/SwiftGenKitTests/TestsHelper.swift
djbe: Well it addresses your comment from the original code:
> // maybe only if `["true", "false"].contains(string) ? Or for all strings?

Wether it should be done or not, dunno. YAML doesn't quote strings by default, so I tried to emulate that behaviour.

AliSoftware: Tbh when I first read the did on GitHub the first diff shown is the context for colors, and I wondered why / found it odd that `"66"` was quoted but `ff` wasn't 😜

```diff
  - colors:
    - alpha: ff
 -    blue: 66
 -    green: 96
+     blue: "66"
+     green: "96"
      name: ArticleBody
-     red: 33
+     red: "33"
```
(then saw your implementation later in the diff and got why)
Maybe quoting everything would be more consistent and eye-pleasing? 🧐😛

Tests/SwiftGenKitTests/TestsHelper.swift
djbe: Where should the file with this code go? (which target?)

AliSoftware: Probably SwiftGenKit?

djbe:
Huh, that's weird, those tests run fine locally.

Podfile.lock
```diff
@@ -7,15 +7,25 @@ PODS:
   - StencilSwiftKit (2.5.0):
     - Stencil (~> 0.11)
   - SwiftLint (0.25.1)
-  - Yams (0.7.0)
+  - Yams (1.0.0)
```
AliSoftware: 🤔 what's new in Yams 1.0.0? Maybe we could use some new stuff…

djbe: > Rename ScalarRepresentableCustomizedForCodable to YAMLEncodable.
And some new documentation available at https://www.jpsim.com/Yams/.

djbe: So nothing for us really, I think. It's just nice that we now have a 1.0 version we can depend on.

AliSoftware: 👍

PR #464, opened 2018-08-01
Restrict SceneType to UIViewController in UIKit

### LINKED ISSUE #463: SceneType swift4 template Any can be UIViewController when targeting iOS
When targeting iOS using SwiftGen 5.3.0 and swift4 storyboard template, SceneType generates as:

```
internal struct SceneType<T: Any> {
  internal let storyboard: StoryboardType.Type
  internal let identifier: String

  internal func instantiate() -> T {
    let identifier = self.identifier
    guard let controller = storyboard.storyboard.instantiateViewController(withIdentifier: identifier) as? T else {
      fatalError("ViewController '\(identifier)' is not of the expected class \(T.self).")
    }
    return controller
  }
}
```

Rather than `Any`, these can be typed as `UIViewController` since the call to `instantiateViewController` can only return a `UIViewController` in UIKit: https://developer.apple.com/documentation/uikit/uistoryboard/1616214-instantiateviewcontroller

This only applies to iOS; AppKit's corresponding instantiateController still returns `Any`:  https://developer.apple.com/documentation/appkit/nsstoryboard/1426549-instantiatecontroller

(I'm happy to submit a PR to modify the template for this if you'd like me to)

darronschall:
This change resolves https://github.com/SwiftGen/SwiftGen/issues/463.

Both `instantiateViewController` and `instantiateInitialViewController` in UIKit return `UIViewController` instances. In AppKit, the corresponding `instantiateController` and `instantiateInitialController` still return `Any`.

---

### DIFF
--- a/templates/ib/swift4.stencil
+++ b/templates/ib/swift4.stencil
@@ -26,7 +26,7 @@ import {{module}}
   }
 }
 
-{{accessModifier}} struct SceneType<T: Any> {
+{{accessModifier}} struct SceneType<T{% if not isAppKit %}: UIViewController{% endif %}> {
   {{accessModifier}} let storyboard: StoryboardType.Type
   {{accessModifier}} let identifier: String
 
@@ -39,7 +39,7 @@ import {{module}}
   }
 }
 
-{{accessModifier}} struct InitialSceneType<T: Any> {
+{{accessModifier}} struct InitialSceneType<T{% if not isAppKit %}: UIViewController{% endif %}> {
   {{accessModifier}} let storyboard: StoryboardType.Type
 
   {{accessModifier}} func instantiate() -> T {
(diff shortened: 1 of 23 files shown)


templates/ib/swift4.stencil
```diff
@@ -26,7 +26,7 @@ import {{module}}
   }
 }
 
-{{accessModifier}} struct SceneType<T: Any> {
+{{accessModifier}} struct SceneType<T: {% if isAppKit %}Any{% else %}UIViewController{% endif %}> {
```
AliSoftware: I wonder if it wouldn't be more idiomatic to just use `SceneType<T>` rather than `SceneType<T: Any>` for AppKit?

AliSoftware:
@darronschall FYI to update the test fixtures, so that the Unit Tests continue to pass, there's a dedicated scheme in the Xcode project to help you auto-update them.

Obviously this magic scheme only has to be used if the only changes you made are on templates, like you did here, and you should still check on the git diff it the changes applied to all outputs are what you expected to produce… (as this will overwrite the expected output files with the generated one instead of comparing them, the tests will always succeed with that scheme… so it's really intended to update the tests fixtures as a one-shot hoping you'll manually check the generated output before committing the changes), but that will save you a lot of time compared to updating the unit test fixtures by hand yourself :wink:

AliSoftware (approved):

Issue #436, opened 2018-06-04
SwiftLint Violation: Conditional Returns on Newline


iotize:
When using SwiftGen to generate fonts using the Swift 4 template in combination with SwiftLint with the rule `conditional_returns_on_newline` enabled, SwiftLint raises a warning:

```
warning: Conditional Returns on Newline Violation: Conditional statements should always return on the next line (conditional_returns_on_newline)
```

This is caused by [this line](https://github.com/SwiftGen/SwiftGen/blob/3d6788dfd8f7699676f1c4aa6f9a2894531a75dc/templates/fonts/swift4.stencil#L26) in the Fonts template.

This can be fixed either by adding a newline:

```swift
{{accessModifier}} func register() {
  guard let url = url else {
    return
  }
  CTFontManagerRegisterFontsForURL(url as CFURL, .process, nil)
}
```

Or by adding this rule to the list of ignored rules at the top of the file:

```swift
// swiftlint:disable superfluous_disable_command
// swiftlint:disable file_length
// swiftlint:disable conditional_returns_on_newline
```

Or, as discussed elsewhere, by excluding generated files from SwiftLint, but assuming that's not the case...

I'm happy to submit a pull request to fix this but I'm not sure which approach is preferable?

djbe:
Hmmm, I wonder why that didn't get flagged in #402.

I think a disable would be preferable here, as I'm pretty sure we use the same style in the codebase for SwiftGen itself as well. @AliSoftware what do you think?

If we go the "disable that rule in the template" route, please note how we handle this in the other templates. All such rules are disabled only for that bit of the template, and re-enabled again as soon as possible in the same file. And we should cover the warning from #427 while we're at it.

AliSoftware:
Tbh for this one I wouldn't be shocked to add the extra missing newline and fix the violation instead of ignoring it

djbe:
I think there're more violations in the codebase itself (not sure though)

iotize:
If it would be consistent with the rest of the codebase, perhaps disabling the rule is the best way to avoid a regression later on?

PR #467, opened 2018-08-07
Fix SwiftLint violation in fonts template

ryanmasondavies:

---

As per #436.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@
   [Darron Schall](https://github.com/darronschall)
   [#463](https://github.com/SwiftGen/SwiftGen/issues/463)
   [#464](https://github.com/SwiftGen/SwiftGen/pull/464)
-* Fonts: disable a warning in generated font files for projects with `conditional_returns_on_newlines` SwiftLint rule enabled.
+* Fonts: disable a warning in generated font files for projects with `conditional_returns_on_newlines` SwiftLint rule enabled.  
   [Ryan Davies](https://github.com/iotize)
   [#436](https://github.com/SwiftGen/SwiftGen/issues/436)
   [#465](https://github.com/SwiftGen/SwiftGen/pull/465)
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -21,6 +21,10 @@
   [Darron Schall](https://github.com/darronschall)
   [#463](https://github.com/SwiftGen/SwiftGen/issues/463)
   [#464](https://github.com/SwiftGen/SwiftGen/pull/464)
+* Fonts: disable a warning in generated font files for projects with `conditional_returns_on_newlines` SwiftLint rule enabled.
```
AliSoftware: Please add 2 spaces at the end of that line (see other entries for inspiration). _That's a formatting trick in Markdown to make the next line be on its own line once rendered_

templates/fonts/swift3.stencil
```diff
@@ -23,7 +23,9 @@
   }
 
   {{accessModifier}} func register() {
+    // swiftlint:disable conditional_returns_on_newline
     guard let url = url else { return }
+    // swiftlint:enable conditional_returns_on_newline
```
AliSoftware: You could also instead directly use `// swiftlint:disable:this conditional_returns_on_newline` just on the affected line,
or `// swiftlint:disable:next conditional_returns_on_newline` on the previous line. This avoids the need to re-enable it on the next line like here.

templates/fonts/swift4.stencil
```diff
@@ -23,7 +23,9 @@
   }
 
   {{accessModifier}} func register() {
+    // swiftlint:disable conditional_returns_on_newline
     guard let url = url else { return }
+    // swiftlint:enable conditional_returns_on_newline
```
AliSoftware: same

AliSoftware (commented):
Thanks for the PR!

Just some nitpicks to address.

AliSoftware:
Also, since you changed the templates, you'll need to fix the unit tests — as they're now generating a different output than what's expected by the tests.

To do that, you could either modify the files in `fixtures/generated/` manually to match your changes in the templates… or use the "Templates - Generate Output" scheme that will automate the re-generation of those (but in that latter solution, be sure to then inspect the git diff to manually check that the changes applied to the generated files match what you expect)

templates/fonts/swift3.stencil
iotize: TIL! This is much neater.

iotize:
Thanks for the feedback! The issues should be resolved now.

AliSoftware (approved):

PR #476, opened 2018-08-30
Swiftlint 0.27 and enabled some extra rules

djbe:
The only potentially contentious one might be `object_literal`.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,13 +1,16 @@
 opt_in_rules:
+ - anyobject_protocol
  - array_init
  - attributes
  - closure_end_indentation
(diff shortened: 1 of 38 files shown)


djbe:
It enforces it. Might sound weird for a project like swiftgen, and it's only really for our tests. But I personally don't think they're too bad? It's easier for visual checks.

AliSoftware:
Do we actually have object literals in the SwiftGen codebase anyway? We don't have UIColors or UIImage or any other UI type in SwiftGen survey code proper, it being a CLI tool, right?

djbe:
It's only in the color parsing tests.

PR #481, opened 2018-09-04
Templates: Organise generated code

### LINKED ISSUE #479: Organise the order of the generated output
Make sure that the static content (declaration of our internal `struct` etc) and the content which really depends on the files being parsed (list of storyboard scenes, assets, etc…) are clearly isolated from one another, via a `// MARK -` or something, and always at the very end of the generated file, to make it easy to access without being polluted by the helper types and implementation details declared in the rest of the files.

So:
- Add `// MARK: - ` comments for the static and dynamic parts
- Move the dynamic part to the top of the file.

djbe:
Fixes #479.

Moves all generated constants to the top, and all other content (implementation details) to the bottom. Also separate the to by a `// MARK: - ...` for easier navigation.

### DIFF
--- a/Tests/Fixtures/Generated/XCAssets/swift3-context-all-customname.swift
+++ b/Tests/Fixtures/Generated/XCAssets/swift3-context-all-customname.swift
@@ -17,6 +17,93 @@ internal typealias XCTData = NSDataAsset
 // swiftlint:disable superfluous_disable_command
 // swiftlint:disable file_length
 
+// MARK: - Asset Catalogs
+
+// swiftlint:disable identifier_name line_length nesting type_body_length type_name
+internal enum XCTAssets {
+  internal enum Colors {
+    internal enum _24Vision {
+      internal static let background = XCTColorAsset(name: "24Vision/Background")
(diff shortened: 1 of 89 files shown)


Tests/Fixtures/Generated/XCAssets/swift3-context-all-customname.swift
```diff
@@ -17,6 +17,93 @@ internal typealias XCTData = NSDataAsset
...
+      Vengo.tint,
+    ]
+    internal static let allDataAssets: [XCTDataAsset] = [
+    ]
+    internal static let allImages: [XCTImageAsset] = [
+      orange,
+    ]
+    // swiftlint:enable trailing_comma
+    @available(*, deprecated, renamed: "allImages")
+    internal static let allValues: [XCTAssetsType] = allImages
```
AliSoftware: ⚠️ This is gonna conflict with the other PR removing that deprecated property, we should be careful when we're gonna rebase

djbe: No problem 😄

PR #482, opened 2018-09-04
XCAssets: remove deprecated constants and types.

djbe:
Implements this comment: https://github.com/SwiftGen/SwiftGen/pull/445#discussion_r200658498.

- We had an old left-over constant `allValues`, from before our asset catalogs parser supported colors and data items.
- I've also removed the old deprecated `{enumName}Type` alias to our image-type alias.
- The default value for our image aliases has also changed from `Image` to `AssetImageTypeAlias`, to be consistent with the other type aliases.

Quick question: should we keep the type alias to `NSDataAsset`? It's a cross platform type, so we can just use `NSDataAsset` directly.
https://github.com/SwiftGen/SwiftGen/blob/master/Tests/Fixtures/Generated/XCAssets/swift4-context-all.swift#L14


AliSoftware:
I say we drop it

Tests/Fixtures/Generated/XCAssets/swift4-context-all-forceNamespaces.swift
```diff
@@ -45,14 +42,14 @@ internal struct DataAsset {
...
-    return AssetDataTypeAlias(asset: self)
+  internal var data: NSDataAsset {
+    return NSDataAsset(asset: self)
   }
   #endif
 }
 
 #if os(iOS) || os(tvOS) || os(OSX)
-internal extension AssetDataTypeAlias {
+internal extension NSDataAsset {
```
djbe: Now that I think about it, shouldn't this also be surrounded by platform checks? So, move the `@available` up to mark this extension, instead of the `init`.

djbe: The documentation about this, and SO answers, are a bit confusing...

djbe: Swift compiler OTOH is pretty clear (tested in a small project): the `@available` must be on the extension.

AliSoftware: That's unfortunate that none of our tests compiling the generated output caught that mistake. Any idea why? Maybe because it's available in macOS and iOS just not on watchOS which might be the only one we don't test?

djbe: We test all platforms (even watchOS and tvOS), but I think we have a "default" deployment target of "latest iOS/macOS/..."

PR #485, opened 2018-09-06
Interface Builder: support `customModuleProvider` property

### LINKED ISSUE #435: Interface Builder: support "inherit module from target" setting
Refs #405, #443.

While dealing with multi-target storyboards today, I saw the following entry in the IB XML:
```xml
<viewController id="0fr-m9-z70" customClass="FiltersViewController" customModule="Graco" customModuleProvider="target" sceneMemberID="viewController">
```

When the "Inherit module from target" flag is set in IB, the XML for that scene is modified, the attribute `customModuleProvider="target"` is added. When it is toggled of, the attribute is completely removed. Supporting this attribute would solve all our issues with `modules` I think (hope)?

Suggested workflow:
Modify the IB parser to check for this attribute. If found, replace the value for `customModule` with whatever's in the `CURRENT_PRODUCT_MODULE` env. variable. If that env. variable is not set, don't replace anything.

djbe:
Fixes #435.

This adds a `moduleIsPlaceholder` property to the context, so that the template can try to use the environment/param module (during generation).

I had to update some resources a bit, and change-up the compilation environment a bit to better handle these changes.


Sources/SwiftGenKit/Parsers/InterfaceBuilder/Scene.swift
```diff
@@ -70,11 +74,16 @@ func == (lhs: InterfaceBuilder.Scene, rhs: InterfaceBuilder.Scene) -> Bool {
...
+    lhs.platform == rhs.platform
 }
 
 extension InterfaceBuilder.Scene: Hashable {
   var hashValue: Int {
-    return identifier.hashValue ^ tag.hashValue ^ (customModule?.hashValue ?? 0) ^ (customClass?.hashValue ?? 0)
+    let part1: Int = identifier.hashValue ^ tag.hashValue
+    let part2: Int = (customModule?.hashValue ?? 0) ^ (customClass?.hashValue ?? 0)
+    let part3: Int = moduleIsPlaceholder.hashValue ^ platform.hashValue
+    return part1 ^ part2 ^ part3
```
AliSoftware: Use the new `hash(into:)` instead?

Sources/SwiftGenKit/Parsers/InterfaceBuilder/Segue.swift
```diff
@@ -57,11 +61,15 @@ extension InterfaceBuilder.Segue: Equatable { }
...
+    lhs.moduleIsPlaceholder == rhs.moduleIsPlaceholder &&
+    lhs.platform == rhs.platform
 }
 
 extension InterfaceBuilder.Segue: Hashable {
   var hashValue: Int {
-    return identifier.hashValue ^ (customModule?.hashValue ?? 0) ^ (customClass?.hashValue ?? 0)
+    let part1: Int = identifier.hashValue ^ (customModule?.hashValue ?? 0) ^ (customClass?.hashValue ?? 0)
+    let part2: Int = moduleIsPlaceholder.hashValue ^ platform.hashValue
+    return part1 ^ part2
```
AliSoftware: `hash(into:)`?

Sources/SwiftGenKit/Parsers/InterfaceBuilder/Scene.swift
djbe: I wanted to but it’s swift 4.2 only.

AliSoftware: Oh. Thought it was introduced earlier than that.

Maybe a `// FIXME:` for later then

djbe: That'll trigger a swiftlint warning though? Which will cause CI to fail 😆
(unless we add a `swiftlint:disable:next`)

AliSoftware: Ah! Damn. Ok so maybe create a "Migrate the Codebase to Swift 4.2" dedicated issue and note that in there to not forget.

djbe:
Hmmm, I'm not exactly sure how such a test would work? Our current tests treat our implementations a bit as black boxes, given an input, and a invocation similar to in the app, check if it matches the output.

Would the test take a generated context, walk recursively and remove all `moduleIsPlaceholder` properties? Would it modify the resource? Some flag in the parser to ignore this implementation? (wouldn't this break the separation between tests and implementation?)

Note that I can't merge this yet as it depends on #419 (split scenes/segues).

AliSoftware:
I think you misunderstood my comment 😅 

What I meant is: are we sure that the tests you wrote in that PR will fail properly if, in the future, we add new templates for which we forget to do the `{% if isModulePlacehodler %}` test. Sure this PR is green today, because in addition to writing the tests, you also fixed the templates. But idk if you did the red-green-refactoring dance (started to write the tests first, made sure they went red, then fix the templates to make them turn green) while doing that?

My point is, if some contributor proposes a new template but fails to do that part in the template:
```
{% macro moduleName item %}{% filter removeNewlines %}
   {% if item.moduleIsPlaceholder %}
     {{ env.PRODUCT_MODULE_NAME|default:param.module }}
   {% else %}
     {{ item.module }}
   {% endif %}
 {% endfilter %}{% endmacro %}
```
But instead just use the `{{item.module}}` in their template, because when they made that template as a custom template for their own specific project they didn't have a storyboard with "inherit from target" in their project, and now they make a PR to propose sharing that template be we want to make sure it will work for those configurations that maybe they never encountered in their own project… Then in those cases, are we sure that the tests we have in this PR will make sure that their template generate code that doesn't compile, and thus make the tests fail and let us know before we validate accepting the new template that this template might not cover those cases.

I guess with the tests you currently have, especially the onces using fixtures with exactly those use cases (= module inherited from target), if we test those fixtures against a new template and that template doesn't handle that case, it will likely produce code that won't compile, and we'll be able to catch it before accepting the new template? That's what I'm asking.

djbe:
Aha, ok, I get what you mean now. So the question is, if I undo the template changes this once (not in the PR), and generate the output, will it fail to compile?

I'll give that a try later tonight or tomorrow.

AliSoftware:
Yeah, the "ending the changes in templates" was meant to be done locally on your Mac just as a temporary change to try it, not a change to be committed in the PR 😅

djbe:
Right so ad893b7 addresses the issue of "bad templates". It simulates the behaviour of:
- Having scenes with the "inherit module from target" setting
- Editing the storyboard files in one target (called `NotCurrentModule` for clarity)
- Then generating swiftgen code in (for example) a build phase of another target (called `SwiftGen` for our purposes)

Templates that don't take the `moduleIsPlaceholder` property into account will instead use the `module` property with as value `NotCurrentModule`, which will generate compilation errors in almost every file.

djbe:
I think only the last option is really viable? It's the only one that can be checked in an automated way, and the impact is pretty minimal (just a small test to see if it finds a scene & segue with that flag and a module).

AliSoftware:
Yeah when you see me writing long comments like that it's generally me thinking aloud and the last bullet point is generally the epiphany I have at the end after writing it all #rubberduckmethod 😂

So last option it should be 👍

PR #495, opened 2018-09-12
Update playgrounds for SwiftGen 6.0

djbe:
Updates the playgrounds with:
- New pages for the new commands
- Update the generated code in the other pages:
  - Swift 4.
  - Split scenes/segues in IB.
  - Color literals template, as it's visually more appealing in a playground, and isn't a production codebase.
  - Structured strings template instead of flat.
- I've also tried moving out all the "implementation details" to a separate swift file for each page. That way the playground page contains only the generated `enum` and some code on how to use it. Do note that, to achieve this I had to add some modifications:
  - Use a `public var bundle: Bundle`, because playgrounds & bundles is kinda broken (we knew this with IB & custom classes already).
  - Add some `public init`s here and there, I've prefixed them with a `// Extra for playgrounds`. This is needed because the extra swift files for pages are compiled as a separate module, and `public struct`s don't have an automatic `public init`. I also tried having those `public init`s in an extension, but then Swift complains about a duplicate definition of the `init` 🤷‍♂️


djbe:
@AliSoftware I accidentally closed #494 when I deleted the `feature/json-yaml-optional` branch, so I had to create this new PR, now with changelog entry 😄

SwiftGen.playground/Pages/Plist-Demo.xcplaygroundpage/Sources/Implementation Details.swift
```diff
@@ -0,0 +1,32 @@
...
+
+  public subscript<T>(key: String) -> T {
+    guard let result = data[key] as? T else {
+      fatalError("Property '\(key)' is not of type \(T.self)")
+    }
+    return result
+  }
+}
+
+public var bundle: Bundle!
```
AliSoftware: Funny that you put this one at the end while the same bundle var is at the top in the other (the JSON-Demo one)

djbe: Ah, must have missed that one, I was moving those around at the end to have them in the same place.

PR #502, opened 2018-09-25
Strings: fix format parser to handle `% d` and `%#x`

### LINKED ISSUE #501: Strings: valid format `% d` is ignored by SwiftGen, but still formatted.
`100% daily` becomes `1000aily` because the Swift template function `tr` always returns `String(format: format, locale: Locale.current, arguments: args)` even if `args` is empty.

`% ` (followed by space) is afaik the *invisible plus sign* so somehow `String(format:` interprets it like a valid format argument (not sure if it should).

But the problem is that the `tr` function should just return the result from `NSLocalizedString` if no args. So just adding `guard !args.isEmpty else { return format }` before the last line would fix it.

djbe:
Fixes #501.

We were missing some cases such as (see docs https://en.wikipedia.org/wiki/Printf_format_string#Flags_field):
- `% d` --> pad with spaces
- `%#x` --> "alternate form" (no idea)
- `%0d` --> pad with zeroes

What's still missing:

Apparently this is possible:
```c
printf("%*d", 5, 10)
printf("%.*s", 3, "abcdef)
```

So both the width and precision fields can accept dynamic values (`int`). After some testing we've noticed that the `String(format:)` initialiser has quite a few bugs with these dynamic fields, so we're not adding them for now.

### DIFF
--- a/Tests/SwiftGenKitTests/StringParserTests.swift
+++ b/Tests/SwiftGenKitTests/StringParserTests.swift
@@ -50,6 +50,18 @@ class StringParserTests: XCTestCase {
     XCTAssertEqual(placeholders, [.unknown, .int, .object, .float, .char])
   }
 
+  func testParseFlags() throws {
+    let formats = ["%-9d", "%+9d", "% 9d", "%#9d", "%09d"]
+    for format in formats {
+      let placeholders = try Strings.PlaceholderType.placeholders(fromFormat: format)
+      XCTAssertEqual(placeholders, [.int], "Failed to parse format \"\(format)\"")
+    }
+
+    let invalidFormat = "%_9d %_9f %_9@ %!9@"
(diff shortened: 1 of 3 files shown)


Tests/SwiftGenKitTests/StringParserTests.swift
```diff
@@ -50,6 +50,15 @@ class StringParserTests: XCTestCase {
...
 
+  func testParseFlags() throws {
+    let formats = ["%-d", "%+d", "% d", "%#d", "%0d"]
+
+    for format in formats {
+      let placeholders = try Strings.PlaceholderType.placeholders(fromFormat: format)
+      XCTAssertEqual(placeholders, [.int], "Failed to parse format \"\(format)\"")
+    }
+  }
+
```
AliSoftware: Might be also worth a reverse-test, like ensure that `Strings.PlaceholderType.placeholders(fromFormat: "%_d")` does NOT lead to `[.int]`.

PR #503, opened 2018-09-25
Strings: Use `tr` without params for strings without params

djbe:
Refs #501.

Adds a `tr` function that doesn't accept any parameters, and directly returns the result of `NSLocalizedString`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,9 @@
 * Strings: the parser now correctly handles formats such as `% d` and `%#x`.  
   [David Jennes](https://github.com/djbe)
   [#502](https://github.com/SwiftGen/SwiftGen/pull/502)
+* Strings: ensure strings without arguments are not processed using `String(format:)`.  
+  [David Jennes](https://github.com/djbe)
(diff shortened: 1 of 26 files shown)


AliSoftware:
I wonder what's best / more readable:

* Having two `tr` functions, one without extra param and one with `CVaArgs…`
* Or having only one, and inserting `if args.isEmpty { return format }` after the call to `NSLocalizedString`

djbe:
My vote goes to the two `tr` functions.

Having 2 makes it very clear what the difference is, instead of having to mentally think about why it would test if the `args` are empty or not.

djbe:
@AliSoftware any final decision on this?

AliSoftware:
Go for two separate `tr` functions!

CHANGELOG.md
```diff
@@ -32,6 +32,9 @@
 * Strings: the parser now correctly handles formats such as `% d` and `%#x`.  
   [David Jennes](https://github.com/djbe)
   [#502](https://github.com/SwiftGen/SwiftGen/pull/502)
+* Strings: ensure strings without arguments are not passed through the `String` formatter.
```
AliSoftware: are not processed using `String(format:)`.

PR #498, opened 2018-09-19
Xcode 10, Swift 4.2, Stencil 0.13.0

djbe:
This updates the project to use Xcode 10 and Swift 4.2.

First order of business is a Stencil issue, where iteration through a Dictionary is now random due to the Hashable changes.

PR #504, opened 2018-09-26
JSON/Plist/Yaml: deeper metadata info

djbe:
⚠️ This is based on the #498 (xcode 10) PR, and should be merged into it to fix the tests.

This is needed to correctly output in a consistent way in the `inline` templates.

It adds an `items` property to `element` metadata objects (for arrays), so that the user can check the metadata information for each array element. This is only done for non-basic (non-uniform) arrays, where we default to the `[Any]` type.

The templates have been updated to handle this "deeper" metadata information, and use it to generate consistent output (that doesn't change between runs). Bonus advantage: the template no long outputs `AnyHashable("myKey")` for dictionary keys.

See the `Info.plist` context for [the Fabric item](https://github.com/SwiftGen/SwiftGen/blob/feature/deeper-metadata/Tests/Fixtures/StencilContexts/Plist/info.yaml#L77), for an example of how this looks.


Tests/Fixtures/StencilContexts/Plist/all.yaml
```diff
@@ -74,7 +74,15 @@ files:
...
-                type: "Dictionary"
+                items:
+                - properties:
+                    KitInfo:
+                      properties: {}
+                      type: "Dictionary"
+                    KitName:
+                      type: "String"
+                  type: "Dictionary"
+                type: "Any"
```
AliSoftware: Isn't that a regression that we go from `type: Dictionary` to `type: Any`? I mean, Sure we also now have `items` and see that we have 2 items, one being a dict and the other being a string… so that's better… but still not sure why the parent element `"Kits` is reported as `Any` now and not `Dictionary`

I guess it depends what definition we give for the metadata we name `Dictionary`: is it supposed to mean "Swift's `Dictionary<Key, Value>`" or "ObjC's `NSDictionary`, / generic _dictionary_ concept"? If it's supposed to mean the Swift & generic type, then I get why it's reported as `Any` here because we have an heterogeneous dict… but then maybe we should add some notes in the SwiftGen Contexts docs then that `Dictionary` means "homogeneous / same key/value types" to just be extra clear about the meaning of the `type` for those metadata structures :wink:

templates/json/inline-swift3.stencil
```diff
@@ -47,13 +47,25 @@ import Foundation
...
 {% endfilter %}{% endmacro %}
-{% macro valueBlock value type %}{% filter removeNewlines:"leading" %}
-  {% if type == "String" %}
+{% macro valueBlock value metadata %}{% filter removeNewlines:"leading" %}
+  {% if metadata.type == "String" %}
     "{{ value }}"
-  {% elif type|hasSuffix:"?" %}
+  {% elif metadata.type == "Optional" %}
     nil
+  {% elif metadata.type == "Array" and metadata.element.items %}
```
AliSoftware: I'm a bit confused. In the Fabric example above, we had the `items` key present (because heterogeneous dict), and `type` was `"Any"`… while here you use `metadata.type == "Array"`… I think i got lost somewhere in there, are heterogeneous structures supposed to be Any or Array, and if they're Any like in the Fabric example able, shouldn't that condition here be `metadata.type == "Any"` then?

templates/json/inline-swift4.stencil
```diff
@@ -47,13 +47,25 @@ import Foundation
...
 {% endfilter %}{% endmacro %}
-{% macro valueBlock value type %}{% filter removeNewlines:"leading" %}
-  {% if type == "String" %}
+{% macro valueBlock value metadata %}{% filter removeNewlines:"leading" %}
+  {% if metadata.type == "String" %}
     "{{ value }}"
-  {% elif type|hasSuffix:"?" %}
+  {% elif metadata.type == "Optional" %}
     nil
+  {% elif metadata.type == "Array" and metadata.element.items %}
```
AliSoftware: Same q obviously here and for the `inline` templates too

templates/json/inline-swift3.stencil
djbe: Forget for a sec about that `Any` case, I've updated the `Metadata.swift` code a bit.

Remember that the Fabric property looks like this:
```yaml
Fabric:
  APIKey: "512345678900aaafffff"
  Kits:
  - KitInfo: {}
    KitName: "Crashlytics"
```

- So the template starts with the value of "Fabric", which is a dict, so it goes into the `{% elif metadata.type == "Dictionary" %}` branch.
- For the "Kits" property, which is an array, it goes into the `{% elif metadata.type == "Array" and ... %}` branch
- For the contents of that array, which is a dictionary, it goes again into the `{% elif metadata.type == "Dictionary" %}` branch
- And so on...

djbe: Remember that for arrays:

- type = "Array"
- element = metadata description of the Array.Element:
  - type = the type of the array element (Any/Dictionary/Array/String/....)
  - items = list of metadata description of each Array item

djbe: See also https://github.com/SwiftGen/SwiftGen/blob/30ec6ff7fb3785e8ed8d38bab8cc5bcb3abf2583/Documentation/SwiftGenKit%20Contexts/Plist.md#example which might clarify things.

AliSoftware: Ohhh right Kits is an array not a dict, missed that.

In the md you linked is it normal that for Kits though we have both `element: type: Dictionary` (as in: "each element in the Kits array is a Dictionary") and `element: items: type: Dictionary` (… same?)

```yaml
              element:
                type: "Dictionary"
                items:
                - type: "Dictionary"
```

djbe: That's simple repetition of the metadata object, yeah. It could just have well been something like:
```swift
["hello", 2, "world", ["weird": "example"]
```
Where the `element.type` = `Any`, and:
- `items.0.type` = `String`
- `items.1.type` = `Int`
- `items.2.type` = `String`
- `items.3.type` = `Dictionary`

PR #506, opened 2018-09-27
Fix (completely) broken swiftlint for output

djbe:
Seems no one noticed that since #409, the linting of output on CI has been broken (not on device though, if you have `swiftlint` installed globally).

Also, the linting performance is abysmal, this PR improves things. It goes from `22.44s` down to `9.80s`.

### DIFF
--- a/Scripts/SwiftLint.sh
+++ b/Scripts/SwiftLint.sh
@@ -26,8 +26,23 @@ if [ -z "$selected_path" ]; then
 	exit 1
 fi
 
+# temporary work directory
+scratch=`mktemp -d -t SwiftGen`
+function finish {
+  rm -rf "$scratch"
+}
+trap finish EXIT
+
+# actually run swiftlint
 if [ "$key" = "templates_generated" ]; then
-	find "${PROJECT_DIR}/${selected_path}" -name "*.swift" -exec ${PROJECT_DIR}/Scripts/swiftlint_no_disable_all.sh {} \;
+	# copy the generated output to a temp dir and strip the "swiftlint:disable:all"
+	for f in `find "${PROJECT_DIR}/${selected_path}" -name '*.swift'`; do
+		temp_file="${scratch}${f#"$PROJECT_DIR"}"
+		mkdir -p $(dirname "$temp_file")
+		sed s@"swiftlint:disable all"@" --"@ "$f" > "$temp_file"
+	done
+
+	"$SWIFTLINT" lint --strict --config "$CONFIG" --path "$scratch" | sed s@"$scratch"@"${PROJECT_DIR}"@
 else
 	"$SWIFTLINT" lint --strict --config "$CONFIG" --path "${PROJECT_DIR}/${selected_path}"
 fi
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,7 +70,9 @@
   [#424](https://github.com/SwiftGen/SwiftGen/pull/424)
 * Use `swiftlint:disable all` in generated files to avoid interference with SwiftLint rules custom to the host project.  
   [Frederick Pietschmann](https://github.com/fredpi)
+  [David Jennes](https://github.com/djbe)
   [#409](https://github.com/SwiftGen/SwiftGen/issues/409)
+  [#506](https://github.com/SwiftGen/SwiftGen/issues/506)
 * XCAssets: Added support for `NSDataAssets`.  
   [Oliver Jones](https://github.com/orj)
   [#444](https://github.com/SwiftGen/SwiftGen/issues/444)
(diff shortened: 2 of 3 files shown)


Scripts/SwiftLint.sh
```diff
@@ -26,8 +26,19 @@ if [ -z "$selected_path" ]; then
...
 
+# copy the generated output to a temp dir and strip the "swiftlint:disable:all"
 if [ "$key" = "templates_generated" ]; then
-	find "${PROJECT_DIR}/${selected_path}" -name "*.swift" -exec ${PROJECT_DIR}/Scripts/swiftlint_no_disable_all.sh {} \;
+	temp_dir=`mktemp -d -t SwiftGen`
+
+	for f in `find "${PROJECT_DIR}/${selected_path}" -name '*.swift'`; do
+		temp_file="${temp_dir}${f#*"$PROJECT_DIR"}"
+		mkdir -p $(dirname "$temp_file")
+		cat "$f" | sed s@"swiftlint:disable all"@" --"@ > "$temp_file"
```
AliSoftware: `cat | sed` is suboptimal as it spawns two processes. `sed` can take the input file as argument instead.

Scripts/SwiftLint.sh
```diff
@@ -26,8 +26,19 @@ if [ -z "$selected_path" ]; then
 	exit 1
 fi
 
+# copy the generated output to a temp dir and strip the "swiftlint:disable:all"
 if [ "$key" = "templates_generated" ]; then
-	find "${PROJECT_DIR}/${selected_path}" -name "*.swift" -exec ${PROJECT_DIR}/Scripts/swiftlint_no_disable_all.sh {} \;
+	temp_dir=`mktemp -d -t SwiftGen`
+
+	for f in `find "${PROJECT_DIR}/${selected_path}" -name '*.swift'`; do
+		temp_file="${temp_dir}${f#*"$PROJECT_DIR"}"
```
AliSoftware: Not sure why you'd need the `*` here, `$f` should immediately start with `$PROJECT_DIR` since it's where we ran find, right? So no need to glob more ahead?

Scripts/SwiftLint.sh
```diff
@@ -26,8 +26,19 @@ if [ -z "$selected_path" ]; then
...
+		cat "$f" | sed s@"swiftlint:disable all"@" --"@ > "$temp_file"
+	done
+
+	files_path="$temp_dir"
 else
-	"$SWIFTLINT" lint --strict --config "$CONFIG" --path "${PROJECT_DIR}/${selected_path}"
+	files_path="${PROJECT_DIR}/${selected_path}"
 fi
+
+"$SWIFTLINT" lint --strict --config "$CONFIG" --path "${files_path}"
```
AliSoftware: Might be worth explicitly/manually remove the temp dir before exiting? (Or does mkdir do that automatically on process exit?)

Either by wrapping that in another `if` here at the end of the script, or by using the `trap` shell command after the mkdir to program a deferred removal.

Scripts/SwiftLint.sh
AliSoftware: Btw I think that was one reason for the performance issue (spawning too many processes) so that change might improve that even further

djbe: No the performance issue is mainly invoking `swiftlint` on a per-file basis, it has to spawn, parse it's own config, etc... `swiftlint` is much faster if you give it a directory to parse.

Scripts/SwiftLint.sh
```diff
@@ -26,8 +26,23 @@ if [ -z "$selected_path" ]; then
...
+trap finish EXIT
+
+# actually run swiftlint
 if [ "$key" = "templates_generated" ]; then
-	find "${PROJECT_DIR}/${selected_path}" -name "*.swift" -exec ${PROJECT_DIR}/Scripts/swiftlint_no_disable_all.sh {} \;
+	# copy the generated output to a temp dir and strip the "swiftlint:disable:all"
+	for f in `find "${PROJECT_DIR}/${selected_path}" -name '*.swift'`; do
+		temp_file="${scratch}${f#"$PROJECT_DIR"}"
+		mkdir -p $(dirname "$temp_file")
+		sed s@"swiftlint:disable all"@" --"@ "$f" > "$temp_file"
```
AliSoftware: I wonder why I chose `@` instead of the usual `/` we commonly use there. Maybe the original script had file paths somewhere in there so I needed something other than `/`, but now that it's not really the case anymore, it's more common and thus easier to read to have `"s/swiftlint:disable all/ --/"` instead of `s@"swiftlint:disable all"@" --"@`

djbe: I get this warning then:
> sed: 1: "s/swiftlint:disable": unterminated substitute pattern

djbe: Woops, nvm, added quotes and it worked

AliSoftware: Even when you do wrap the whole thing in quotes?

AliSoftware: `sed "s/swiftlint:disable all/ --/" Tests/Fixtures/Generated/Strings/structured-swift4-context-multiple.swift` works from the Terminal for me at least

PR #508, opened 2018-09-27
Update building instructions in README

skagedal:
This PR has two commits. First an update to what the README says on where to find the built products, which did not match what I found.

Second, some instructions on how to properly set up Ruby environment before building. I'm not sure if you want these instructions in this form, but it would have helped me though! Also, maybe there are better ways to do this that I'm missing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,10 +129,21 @@ brew install swiftgen --HEAD
 
 * Alternatively, you can clone the repository and use `rake cli:install` to build the tool and install it from any branch, which could be useful to test SwiftGen in a fork or a Pull Request branch.
 
-You can install to the default locations (no parameter) or to custom locations:
+Some Ruby tools are used in the build process, and the system Ruby works well if you are running a recent macOS.  However, if you are using `rbenv` you can run `rbenv install` to make sure you have a matching version of Ruby installed.  
+
+Then install the Ruby Gems:
 
 ```sh
-# Binary is installed in `./swiftgen/bin`, frameworks in `./swiftgen/lib` and templates in `./swiftgen/templates`
+# Install bundle if it isn't installed
+gem install bundle
+# Install the Ruby gems from Gemfile
+bundle install
+```
+
+You can now install to the default locations (no parameter) or to custom locations:
+
+```sh
+# Binary is installed in `./build/swiftgen/bin`, frameworks in `./build/swiftgen/lib` and templates in `./build/swiftgen/templates`
 $ rake cli:install
 # - OR -
 # Binary will be installed in `~/swiftgen/bin`, frameworks in `~/swiftgen/fmk` and templates in `~/swiftgen/tpl`


README.md
```diff
@@ -115,10 +115,21 @@ brew install swiftgen --HEAD
 
 * Alternatively, you can clone the repository and use `rake cli:install` to build the tool and install it from any branch, which could be useful to test SwiftGen in a fork or a Pull Request branch.
 
-You can install to the default locations (no parameter) or to custom locations:
+It is recommended to set up your Ruby environment for building using `rbenv` and `bundle`, like this.
 
 ```sh
-# Binary is installed in `./swiftgen/bin`, frameworks in `./swiftgen/lib` and templates in `./swiftgen/templates`
+# Installs the version of Ruby specified in .ruby-version
+rbenv install
```
AliSoftware: I'm not sure I see why `rbenv` is necessary. And tbh I don't really like suggesting _by default_ to install `rbenv` to build SwiftGen, as if think it's only necessary if you don't have the right ruby version because maybe you don't have the latest macOS (with latest ruby) installed.

I agree that we should mention `rbenv` though, because we have that `.ruby-version` file in order to ensure that if people don't have the same ruby version as we do at least the have that solution. But having people install `rbenv` by default, to be able to compile `swiftgen`, might both scare people not at ease with Ruby and not confortable with having to install a different ruby etc, and I think for 90% of people they already have the right ruby version installed in their macOS, so if they don't have `rbenv` and we tell them that they have to install it, it might be a blocker even though it would end up not being necessary for them…

So maybe we should reword that, like "if you don't have the latest macOS version and especially your version of ruby is less than the minimum ruby version we used that is listed in the `.ruby-version` file, you might want to use `rbenv` to use the proper ruby for this project …" but otherwise, by default, don't really force people to install brent since they're likely to have the right version of ruby already? (to be honest I don't use brent at all even if I contribute regularly to SwiftGen, so…)

skagedal: Ah, yeah that makes sense! The situation for me was that I _did_ have `rbenv` installed, but did not have this particular version installed. Running `rake` then complained about that, and I had to go figure out how to install the right version. Not too hard of course, but it would have been nice. :)

So maybe just like "if you use `rbenv`, do `rbenv install` to make sure you have the same Ruby version"...?

AliSoftware: Yeah, sounds good to me :+1:

AliSoftware: > if you use `rbenv`, do `rbenv install` to make sure you have the same Ruby version (if you don't use `rbenv`, you probably have the right Ruby version already if your macOS is up-to-date, as we just use system ruby)

djbe:
This will need a rebase (or merge from master). Be careful as I've just merged the migration guide PR, which also modified quite a bit of the Readme.

skagedal:
@AliSoftware How about this wording?

@djbe Sure, I'll rebase and squash to one commit after we're happy with the wording, is that good..?

djbe:
Jep, good for me.

README.md
```diff
@@ -115,11 +115,9 @@ brew install swiftgen --HEAD
 
 * Alternatively, you can clone the repository and use `rake cli:install` to build the tool and install it from any branch, which could be useful to test SwiftGen in a fork or a Pull Request branch.
 
-It is recommended to set up your Ruby environment for building using `rbenv` and `bundle`, like this.
+Some Ruby tools are used in the build process, and the system Ruby works well if you are running a recent macOS.  However, if you are a using `rbenv` you can run `rbenv install` to make sure you have a matching version of Ruby installed.  Then install the Ruby Gems:
```
AliSoftware: s/if you are a using/if you are using/

I'd start the "Then install the Ruby Gems" in a dedicated separate paragraph too

PR #509, opened 2018-09-27
Release 6.0.0

djbe:
I've moved the "bug fixes" changelog section to be after "new features". I've adapted the rake script as well. The order is now:

- Breaking changes
- New features
- Bug fixes
- Internal changes

### DIFF
--- a/Tests/SwiftGenTests/ConfigLintTests.swift
+++ b/Tests/SwiftGenTests/ConfigLintTests.swift
@@ -187,9 +187,9 @@ class ConfigLintTests: XCTestCase {
         (.error, "xcassets.inputs: Fixtures/XCAssets/Images.xcassets does not exist"),
         (.error, "xcassets.inputs: Fixtures/XCAssets/Images.xcassets does not exist"),
         (.error, """
-          xcassets.outputs: Template named custom-swift3 not found. Use `swiftgen templates` to \
-          list available named templates or use `templatePath` to specify a template by its full \
-          path.
+          xcassets.outputs: Template named custom-swift3 not found. Use `swiftgen templates list` \
+          to list available named templates or use `templatePath` to specify a template by its \
+          full path.
           """),
         (.error, """
           xcassets.outputs.output: Generated does not exist. Intermediate folders up to the output \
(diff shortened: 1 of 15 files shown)


AliSoftware:
I'll do a PR to link to the new wiki page about swiftlint custom rules before we merge this PR

AliSoftware:
(We should also include https://github.com/SwiftGen/SwiftGen/pull/508 in this release)

Tests/SwiftGenTests/ConfigLintTests.swift
```diff
@@ -187,9 +187,9 @@ class ConfigLintTests: XCTestCase {
         (.error, "xcassets.inputs: Fixtures/XCAssets/Images.xcassets does not exist"),
         (.error, "xcassets.inputs: Fixtures/XCAssets/Images.xcassets does not exist"),
         (.error, """
-          xcassets.outputs: Template named custom-swift3 not found. Use `swiftgen templates` to \
-          list available named templates or use `templatePath` to specify a template by its full \
-          path.
+          xcassets.outputs: Template named custom-swift3 not found. Use `swiftgen templates list` \
```
AliSoftware: Wait, to we still have to use `\` at the end of line even if we're using Swift 4's multi-line string literals?

djbe: It’s to keep it as one line

AliSoftware: Ah yeah of course 👍

djbe:
I've updated the `Update the version of SwiftGen` section in that wiki page. I'm not familiar enough with the release process to update the `Publish a new release` accordingly.

djbe:
@AliSoftware right, so I wrongly assumed that the CLI template name option was called `--templateName` when I updated the error messages and docs (it's called `--template`).

We can either leave them as is, or add new options that match those docs and deprecate the old one. Not sure if that warrants a mention in the MG, as the old option is still around.

AliSoftware:
CHANGELOG entry I think yes, we need to trace that somewhere at least. MG entry probably not worth it, the overall suggestion in the MG already being basically "Y U Not use a config file already?" anyway 😜

AliSoftware:
TODO: Update the playground + `rakelib/playground.rake` since we recently renamed the YAML resources

PR #514, opened 2018-09-28
[Rakefile] Move release task from Eve to SwiftGen

AliSoftware:
And split some namespaces into their own rakelib files too

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,4 +17,5 @@ end
 # For releasing to GitHub
 group :release do
 	gem 'octokit', '~> 4.9'
+	gem 'plist', '~> 3.2'
 end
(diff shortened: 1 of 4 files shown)


AliSoftware:
Could be interesting indeed. For now I'm just checking that all the tasks are still working and that I didn't break anything, but yeah, might be worth adding some to the CI indeed.

We'll also need to revisit Eve. We had a task that synchronised the rakelibs between our repos, but now maybe only some of those `*.rake` files will make sense to synchronize, not others. And probably not as many common tasks as before the Big Merge anyway.

djbe:
We can add it later.

Regarding Eve, I wouldn't delete it now. Even if it's for a few tasks, it can always be helpful to still have it.

AliSoftware:
I wasn't suggesting to delete Eve, nor to delete Eve's Rakefile. Just pointing out that our `sync:all_files` task back there might override the changes I made here by overriding some files in `rakelib/` so we should be careful about that and either get rid of the `sync:all_files` task in Eve, or adapt it to only synchronize some of those `*.rake` files but not all, to avoid override.

PR #517, opened 2018-09-29
Fix various typos in README.md

skagedal:


djbe:
@skagedal Thx for the quick fix.

Could you rebase on `develop` please? We're switching to gitflow.

skagedal:
Hm, did that rebase turn out correct? Seems there's a commit from @AliSoftware that shouldn't be there 🤔

djbe:
It looks ok 👍

PR #518, opened 2018-09-29
Make the output:compile task test both Swift 4 and 4.2

AliSoftware:
* [ ] I've added an entry in the `CHANGELOG.md` file to explain my changes and credit myself
  * [ ] Add the entry in the appropriate section (Bug Fix / New Feature / Breaking Change / Internal Change)
  * [ ] Add a period and 2 spaces at the end of your short entry description

---


djbe:
Oh boy, there go our build times 😱

AliSoftware:
Yeah, it's gonna be quite urgent to find a way to parallelize those soon!

djbe:
Thing is, we can only build on 1 mac device at a time 😢

CHANGELOG.md
```diff
@@ -14,7 +14,10 @@ _None_
 
 ### Bug Fixes
 
-_None_
+* Fix issue with Storyboard templates when compiled with Swift 4.2 on macOS.
```
djbe: I try to, when applicable, prefix my changelog entries with the command, so in this case `IB: `. The issue also only applies to `segues-swift4`, so we may want to note that.

PR #533, opened 2018-10-02
Config: error reporting

### LINKED ISSUE #516: When swiftgen.yml parsing fails, output error messages that Xcode parses
1. Integrate SwiftGen with Xcode as a Script Build Phase.
2. Have an invalid swiftgen.yml, for example by making it just contain the word `failure`. 
3. Build your project. 

Expected behavior: the build fails.
Observed behavior: none – you have to open the build output to see the failure. 

In this case it prints:
```
Wrong type for key root: expected Dictionary, got Optional<Any>.
```

It would be nice if it instead printed:
```
/path/to/swiftgen.yml:1:error:Wrong type for key root: expected Dictionary, got Optional<Any>.
```

(In this case an even better error message like "Invalid YAML" would have been better of course, but that is not the point of this issue.)

It could be done so it only uses this format if run within Xcode, or if a certain flag is passed. (I could not find such an existing flag.)

AliSoftware:
* [X] I've added an entry in the `CHANGELOG.md` file to explain my changes and credit myself
  * [X] Add the entry in the appropriate section (Bug Fix / New Feature / Breaking Change / Internal Change)
  * [X] Add a period and 2 spaces at the end of your short entry description
  * [X] Add links to your GitHub profile to credit yourself and to the related PR and issue after your description

---

Fixes #516

### DIFF
--- a/Sources/SwiftGen/Logs.swift
+++ b/Sources/SwiftGen/Logs.swift
@@ -46,9 +46,9 @@ func logMessage(_ level: LogLevel, _ string: CustomStringConvertible) {
   case .info:
     fputs(ANSIColor.green.format("\(string)\n"), stdout)
   case .warning:
-    fputs(ANSIColor.yellow.format("Warning: \(string)\n"), stderr)
+    fputs(ANSIColor.yellow.format("swiftgen: warning: \(string)\n"), stderr)
   case .error:
-    fputs(ANSIColor.red.format("Error: \(string)\n"), stderr)
+    fputs(ANSIColor.red.format("swiftgen: error: \(string)\n"), stderr)
   }
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,10 @@ _None_
 * Ensure the `templates` subcommand properly works with the deprecated `storyboards` subcommand.
   [David Jennes](https://github.com/djbe)
   [#529](https://github.com/SwiftGen/SwiftGen/pull/529)
+* Ensure configuration file errors show as red in Xcode.  
+  [Olivier Halligon](https://github.com/AliSoftware)
+  [#533](https://github.com/SwiftGen/SwiftGen/pull/533)
+  [#516](https://github.com/SwiftGen/SwiftGen/issues/516)
 
 ### Internal Changes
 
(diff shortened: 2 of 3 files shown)


SwiftGen-Eve:

</p>

Sources/SwiftGen/Logs.swift
```diff
@@ -46,9 +46,9 @@ func logMessage(_ level: LogLevel, _ string: CustomStringConvertible) {
   case .info:
     fputs(ANSIColor.green.format("\(string)\n"), stdout)
   case .warning:
-    fputs(ANSIColor.yellow.format("Warning: \(string)\n"), stderr)
+    fputs(ANSIColor.yellow.format("swiftgen: warning: \(string)\n"), stderr)
```
djbe: This works? (not having error/warning at the start of the line)

AliSoftware: Not only does it work but it's part of the expected format:

`<filename>:<line>:<col>: warning: <message>`

Where the `<col>:` can be (and often is) omitted, and if it is,`<line>:` can be omitted too.
`<filename>:` can _also_ be omitted completely, but it's less common; and `warning:`/`error:` are lowercase, even if maybe Xcode might be case insensitive about that to trigger a yellow/red banner on the file/line

djbe: 👍

AliSoftware: Note that for a minute there I thought about changing the enum (.info/.warning/.error) to have associated values for file and line. Then realized for config file we don't have line references back in the original YAML. And I wanted to do a quick PR that we could merge in 6.0.1 fast, while doing such a change would affect quite a lot in the codebase. So in the end I chose to keep it simple

In the ideal solution we wouldn't just print `swiftgen:` at the beginning of that log line but `swiftgen.yml` if the error is coming from the (default) config file and potentially something else if it's become from some other source (like error parsing an input storyboard or whatnot), hence the idea of an enum with associated value

But didn't want to use `swiftgen.yml:` in the generic log message here because until we move to an EAV we are not sure if the error comes from another place other than the config file or if the config file used isn't the default one and have a different name… so I kept all that for later and kept it simple

PR #543, opened 2018-10-04
Strings: correct the types for character and pointer parameters

### LINKED ISSUE #541: Strings: generated type for `%c` should be `CChar` instead of `Character`
From https://github.com/SwiftGen/SwiftGen/issues/501#issuecomment-427085029.

Correspondingly, we should update the type for:
- `%s`: `UnsafePointer<CChar>` instead of `UnsafePointer<unichar>`
- `%p`: `UnsafeRawPointer`, because `UnsafePointer<Void>` is deprecated (since Swift 3)
- unknown: same as above

djbe:
Fixes #541.

Ideally the `type` field would contain an abstract type such as "Integer", "Float", "Object", etc.. And the templates decide how to map that type to a more concrete type depending on the Swift version. Who knows what'll happen with Swift 5 for example.


SwiftGen-Eve:



---

</p>

AliSoftware:
> Who knows what'll happen with Swift 5 for example.

Also think about those other WIP PRs that aim to add templates for ObjC… 😉

AliSoftware:
Is there an easy way in Stencil to create a map from within the template, instead of having to do a long `{% if %}{% else %}` dance for all possibilities?

(Like maybe something like `{{ "@:String,d:Integer,f:Double"|split:","|whatnot }}`)

djbe:
Eventually we should switch from concrete types to more general ones, but that'll be a breaking change that is not necessary now. I'd only consider it for the next breaking release, if we ever need it.

I say we tackle whatever template question then, we'll (hopefully) have some advanced filter/selection stuff by then.

djbe:
We can keep track of these long-term, general TODOs in https://github.com/SwiftGen/SwiftGen/projects/3

PR #546, opened 2018-10-07
Release/6.0.2

AliSoftware:
Targetting master, so Big PR, but that's what we said the new process was, right?


AliSoftware:
For refs: 

> ⚠️  You should make sure that Podfile.lock and Package.resolved are changed in sync

This is normal: it's because one of the changes was #544 which added a `post_install` hook in the `Podfile`, making the `Podfile.lock` be modified just to change the `PODFILE CHECKSUM`, nothing more. So nothing to worry about :+1:

Issue #552, opened 2018-10-13
Xcode 10.1 beta 2 Error (Missing entry for key xcassets.paths.)


YiYiZheng:
1 installed SwiftGen via cocoapods. 
2 add run script "${PODS_ROOT}/SwiftGen/bin/swiftgen"
3 add swiftgen.yml to project root dir. 
4 build project

compile error: 
Missing entry for key xcassets.paths.
Command PhaseScriptExecution failed with a nonzero exit code

What could I do? Thanks

AliSoftware:
Hi

The message is telling you that you have an error in your `swiftgen.yml` configuration file.

_You can use `swiftgen config lint` in your terminal to try to have more information about the syntax error that you have in that config file of yours._

But according to the error SwiftGen already reports to you, you're missing the key `paths` under the `xcassets` key of your config file.
What's odd though is that this `paths` key is deprecated and was renamed in the latest versions of SwiftGen; so what probably happened is that you installed an old version of SwiftGen (5.x) but you followed the documentation of the new SwiftGen (6.0) — which correctly documents the config file structure to expect `inputs` for the key listing the path to scan (while this key was named `paths` in 5.x)

Could you just check that you installed the latest version of SwiftGen and not an old 5.x version? My best guess is that you probably forgot to run `pod repo update` to update CocoaPods' repo of podspecs before installing your pods (including SwiftGen) with `pod install`

AliSoftware:
Sorry I meant `Pods/SwiftGen/bin/swiftgen config lint` not just `swiftgen config lint` (since you installed SwiftGen using CocoaPods so that's where it got installed in your project). Run that command in the terminal after having moved (using `cd`) to the directory where your project is, obviously 😉

Did you check the version of SwiftGen you did installed btw? You can check it either by running `Pods/SwiftGen/bin/swiftgen --version`, out by looking into your `Podfile.lock` generated by CocoaPods

YiYiZheng:
Yeah! After run `pod repo update`, I can get the new version of SwiftGen. Xcode compiles successfully. 
I will continue to dive in.
My suggestion is specify this problem in the README.md. Also `Pods/SwiftGen/bin/swiftgen config lint`. I took a long time for these problems.   
Thx.

AliSoftware:
Thanks for the suggestion and the follow up!

[...]

PR #553, opened 2018-10-13
README: Additional info about CocoaPods installations — #trivial PR

AliSoftware:
This addresses the feedback and suggestions in https://github.com/SwiftGen/SwiftGen/issues/552#issuecomment-429523054

\cc @YiYiZheng

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,4 +1,5 @@
 * [ ] I've started my branch from the `develop` branch (gitflow)
+  * [ ] And I've selected `develop` as the "base" branch for this Pull Request I'm about to create
 * [ ] I've added an entry in the `CHANGELOG.md` file to explain my changes and credit myself  
       _(or added `#trivial` to the PR title if I think it doesn't warrant a mention in the CHANGELOG)_
   * [ ] Add the entry in the appropriate section (Bug Fix / New Feature / Breaking Change / Internal Change)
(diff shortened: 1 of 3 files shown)


AliSoftware:
I'm not sure I like suggesting `pod update SwiftGen` as `pod update` is intended to _update_ the version of SwiftGen _already installed_ by a previous `pod install`.

Instead we could suggest `pod install --repo-update` (which essentially is a shorthand for `pod repo update && pod install`).

That being said, should be also mention to run `pod update SwiftGen` from time to time to _update_ to the latest SwiftGen when you want to upgrade to the newest version on a project already having it? (I'm feeling like we'd explain CocoaPods basics to users here though. It's nice to guide the users and give useful commands, but it's also not our job to replicate CocoaPods' own documentation…)

djbe:
Well, `pod update SwiftGen` is just easier to write and remember compared to `--repo-update`, and has the added benefit to update SwiftGen for old users.

But I get your point, I'm kind of misusing the command here. I honestly don't think we should start documenting CocoaPods, we already have enough installation instructions as it is.

What we could do though, and this will "force" the issue a bit, is adding a version parameter to the Podfile dependency:
> If you're using CocoaPods, you can simply add `pod 'SwiftGen'` to your `Podfile`.

Becomes:
> If you're using CocoaPods, you can simply add `pod 'SwiftGen', '~> 1.6.0'` to your `Podfile`.

And we can keep that line up-to-date with whatever version we're at (including bugfixes). Just add it as a step to our release process.

AliSoftware:
idk, even if we could try to automate that in our release process or something, I tend to avoid using 3-digit versions in a `~>` expression (because semver, so you usually want not to bump the major version unexpectedly but always want the minor and patch versions of that major version).

But indeed we could suggest `pod 'SwiftGen', '~> 6.0'` then

PR #559, opened 2018-11-18
Symbolic link workaround

Liquidsoul:
* [ ] I've added an entry in the `CHANGELOG.md` file to explain my changes and credit myself  
  * [ ] Add the entry in the appropriate section (Bug Fix / New Feature / Breaking Change / Internal Change)
  * [ ] Add a period and 2 spaces at the end of your short entry description

---

Seems like there is an issue if we call swiftgen using a symbolic link.
The issue does not come from swiftgen implementation but seems to be an Apple bug which does not return the same `arguments[0]` depending on where you call the symbolic link from. 
[This sample](https://github.com/Liquidsoul/apple-radars/tree/master/symlink-arg0) shows the issue.
In my opinion, the value should always be the symbolic link path, meaning that bundled templates resolution would not work. So, in order to solve this, swiftgen needs to go to the concrete swiftgen path.

### DIFF
--- a/Sources/SwiftGen/Path+AppSupport.swift
+++ b/Sources/SwiftGen/Path+AppSupport.swift
@@ -36,4 +36,15 @@ let templatesRelativePath: String = {
 private final class BundleToken {}
 
 let appSupportTemplatesPath = Path.applicationSupport + "SwiftGen/templates"
-let bundledTemplatesPath = Path(ProcessInfo.processInfo.arguments[0]).parent() + templatesRelativePath
+let binaryPath: Path = {
+  var binaryPath = Path(ProcessInfo.processInfo.arguments[0])
+  do {
+    while binaryPath.isSymlink {
+      binaryPath = try binaryPath.symlinkDestination()
+    }
+  } catch {
+    print("Warning: could not resolve symlink of \(binaryPath) with error \(error)")
+  }
+  return binaryPath
+}()
+let bundledTemplatesPath = binaryPath.parent() + templatesRelativePath


SwiftGen-Eve:

---

</p>

Sources/SwiftGen/Path+AppSupport.swift
```diff
@@ -36,4 +36,15 @@ let templatesRelativePath: String = {
 private final class BundleToken {}
 
 let appSupportTemplatesPath = Path.applicationSupport + "SwiftGen/templates"
-let bundledTemplatesPath = Path(ProcessInfo.processInfo.arguments[0]).parent() + templatesRelativePath
+let binaryPath: Path = {
+  var binaryPath = Path(ProcessInfo.processInfo.arguments[0])
+  do {
+    while binaryPath.isSymlink {
+      binaryPath = try binaryPath.symlinkDestination()
+    }
```
AliSoftware: Playing the devil's advocate here, isn't there a risk of a symlink reference cycle on the filesystem, that would make this kind of loop run forever? In such case, does this code automatically throw already (the `NSFileManager` that is used under the hood by `PathKit` ends up throwing its own `Error`)? Or is it our responsibility to add a max redirection threshold?

```
$ ln -s a b
$ ls -s b a
$ cat a
cat: a: Too many levels of symbolic links
```

Liquidsoul: How could your binary be a symlink to itself 🤔 
If this code is called, this means you're executing `swiftgen` binary, so `arguments[0]` can only reference the `swiftgen` binary in the end, even if you have multiple redirections 😉

AliSoftware: 😅 very good point 😂

Also I'm wondering, isn't there any other means to get the real path to the current process rather than rely on `args[0]`? Maybe there's some property on `ProcessInfo` or whatnot? 🤔

Liquidsoul: I did not find any on [`ProcessInfo`](https://developer.apple.com/documentation/foundation/processinfo). I tried [`NSRunningApplication.executableURL`](https://developer.apple.com/documentation/appkit/nsrunningapplication/1531062-executableurl) but this does not seem to work. It gives me a `nil` result.

AliSoftware: Ok, let's keep that symlink resolution loop then

Issue #355, opened 2017-11-01
[Feature] Various Config file improvements (env vars, better lint…)


AliSoftware:
Various ideas to improve `swiftgen.yml` a bit:

* [ ] Get inspiration from https://github.com/realm/SwiftLint/pull/1703/files to expand ENV vars from inside the `swiftgen.yml` file
* [ ] Use a simple command-line argument (instead of a `--config` option) to specify an alternate config file. E.g. make it be `swiftgen config run swiftgen-config.yml` instead of `swiftgen config run --config swiftgen-config.yml`
* [ ] Make `swiftgen config lint` warn about invalid keys

Lutzifer:
@AliSoftware 

Are there any news on  the "expand Env Vars" feature?

AliSoftware:
@Lutzifer haven't found the time to implement it yet, nope 😅 

Actually I think it could be a nice starter PR, given that a similar code exists for SwiftLint (see link) so it's more or less a matter of duplicating and adapting that SwiftLint code to SwiftGen :wink: so feel free to give it a try if you are willing 😜

Lutzifer:
I'll try to have a closer look on of the next weekends.

PR #564, opened 2018-12-23
Expand environment variables in config YML files

Lutzifer:


Sources/SwiftGen/Config/Config.swift
```diff
@@ -27,7 +28,20 @@ extension Config {
     if !file.exists {
       throw Config.Error.pathNotFound(path: file)
     }
-    let anyConfig = try YAML.read(path: file)
+    
+    // Read contents and replace env variables
+    var contents: String = try file.read()
+
+    ProcessInfo.processInfo.environment.keys.forEach { key in
```
djbe: I've always found `forEach` a bit weird, especially if you need the value later:
```suggestion
    for (key, value) in ProcessInfo.processInfo.environment {
```

djbe (changes requested):

SwiftGen.xcodeproj/xcshareddata/xcschemes/swiftgen.xcscheme
```diff
@@ -88,6 +88,13 @@
             ReferencedContainer = "container:SwiftGen.xcodeproj">
          </BuildableReference>
       </MacroExpansion>
+      <EnvironmentVariables>
```
djbe: Instead of adding an env. variable here (which is quite hidden), I'd modify the `Config.init` to accept an environment dictionary (defaulting to `ProcessInfo.processInfo.environment`). That way, in the test below you can provide the necessary values in the test itself.

Tests/SwiftGenTests/ConfigReadTests.swift
```diff
@@ -44,6 +44,32 @@ class ConfigReadTests: XCTestCase {
...
   }
 
+  func testConfigWithENVS() throws {
+    guard let paramsConfigFilePath = Bundle(for: type(of: self)).path(forResource: "config-with-params", ofType: "yml") else {
+      fatalError("Fixture not found")
+    }
+
+    // this test needs the env varibale for "SWIFTGEN_OUTPUT_DIR" set to "Common/Generated" in the test section of the swiftgen scheme
+    guard let envConfigFilePath = Bundle(for: type(of: self)).path(
+      forResource: "config-with-env_placeholder",
```
djbe: This filename breaks a bit with the other files, we probably want "config-with-env-placeholder" here.

Tests/SwiftGenTests/ConfigReadTests.swift
```diff
@@ -44,6 +44,32 @@ class ConfigReadTests: XCTestCase {
...
+    ) else {
+      fatalError("Fixture not found")
+    }
+
+    let paramsFile = Path(paramsConfigFilePath)
+    let envFile = Path(envConfigFilePath)
+
+    do {
+      let paramsConfig = try Config(file: paramsFile)
+      let envConfig = try Config(file: envFile)
```
djbe: You'd call the init with a custom dictionary here.

djbe (changes requested):

djbe:
Thanks for the PR!

Right, so besides the swiftlint issues and some code nitpicks, this seems like a great improvement! We may want to add some documentation about this feature, probably in the [config file docs](https://github.com/SwiftGen/SwiftGen/blob/master/Documentation/ConfigFile.md).

AliSoftware:
Thanks a lot for the PR, will make a great improvement, and we'll finally be able to check at least one box from #355!

Regarding the implementation, I think I'd be safer to go [the same route as SwiftLint thought, which decodes environment variables during the reading of the YML file] (https://github.com/realm/SwiftLint/pull/1703/files). This would especially avoid issues with the current implementation in case some env variables contain special characters once expanded that would make the subsequent YAML parsing fail.

Maybe we can even steal some of the unit tests from SwiftLint as well on this feature btw 😉

djbe:
Do ignore the "`yes`, `no`, `on` and `off` as strings" part of that PR.

AliSoftware:
PS @djbe I didn't investigate much (on iPhone RN) but any idea why Danger has been silent?

AliSoftware:
Ah right. Maybe we should run it first. But otoh would be nice to have Danger report lint warnings right in the PR as comment as well so…

Btw not sure I get the error messages about $TERM and -T in the logs of the lint task, and if that failure of lint is due to the violations maybe rake should fail more gracefully too to be less scary and make it differ from a potential Ruby error in our scripts  😅

Lutzifer:
thanks @marcelofabri for the original implementation

SwiftGen-Eve:

---

</p>

Sources/SwiftGenKit/Utils/YAML.swift
```diff
@@ -61,3 +61,50 @@ public enum YAML {
     }
   }
 }
+
+// Copied from /Source/SwiftLintFramework/Models/YamlParser.swift
+// of https://github.com/realm/SwiftLint/blob/d1dbc31aa9269364d6d7f43d2f99c82e12ceca6f
```
AliSoftware: I think you can directly link to the file at a given commit in one link: https://github.com/marcelofabri/SwiftLint/blob/33ca7d71f77025100dcb4bab5f464d342664cc5b/Source/SwiftLintFramework/Models/YamlParser.swift

Sources/SwiftGenKit/Utils/YAML.swift
```diff
@@ -61,3 +61,50 @@ public enum YAML {
...
+      result = result.replacingOccurrences(of: "${\(key)}", with: value)
+    }
+
+    return result
+  }
+}
+
+private extension Bool {
+  // swiftlint:disable:next discouraged_optional_boolean
+  static func constructUsingOnlyTrueAndFalse(from scalar: Node.Scalar) -> Bool? {
```
AliSoftware: I'd have to check from a computer to have the big picture and navigate in the codebase at ease, but didn't we have that logic for special handling of booleans in the YAML already? I remember we had to make sure `true` would be interpreted as Boolean but `"true"` as a string and thus already had something for that…

Tests/Fixtures/Configs/config-with-env-placeholder.yml
```diff
@@ -0,0 +1,18 @@
+output_dir: ${SWIFTGEN_OUTPUT_DIR}
+strings:
+  inputs: Sources1/Folder
```
AliSoftware: I'd add another env var here (typically `${PROJECT_DIR}`) or somewhere else in the YAML, so that the test will also ensure that all variables would be expanded, not just the first one

Tests/Fixtures/Configs/config-with-env-placeholder.yml
```diff
@@ -0,0 +1,18 @@
...
+  outputs:
+    templateName: structured-swift3
+    output: strings.swift
+    params:
+      foo: 5
+      bar:
+        bar1: 1
+        bar2: 2
+        bar3:
```
AliSoftware: We should also check in the test (maybe a separate test) that env vars are NOT replaced where they are not supposed to. Especially if they are used as keys in the YAML they shouldn't be expanded, and similarly if they are not wrapped inside `${…}` either (so that `inputs: "Card$Game/Sources"` doesn't replace the `$Game` if there happen to be a `Game` env var defined

Sources/SwiftGenKit/Utils/YAML.swift
djbe: Yeah, like I said in a previous comment, we don't need the custom boolean logic, we already solve that by enforcing all strings to have `"` (quotes).

djbe (approved):

AliSoftware:
Thanks a lot for contributing to SwiftGen @Lutzifer ! 👍 
To thank you for your work and for taking time to improve the tool, I've invited you to join SwiftGen's GitHub organization's "CoreContributors" team 😃 

No pressure to accept nor any obligation in that! It will just give you the ability to help triage issues or push directly branches to the repo (instead of your fork)… in case you want to contribute again in the future 😉 

For more information, read the [COMMUNITY.md](https://github.com/SwiftGen/SwiftGen/blob/master/COMMUNITY.md) and [CONTRIBUTING.md](https://github.com/SwiftGen/SwiftGen/blob/master/CONTRIBUTING.md) documents, and if you have any questions don't hesitate to reach out!

Lutzifer:
@djbe Thanks for finishing, did not find the time in-between-years

djbe:
No problem, I had a pretty quiet new years eve and some extra time on my hands 😆

Sources/SwiftGenKit/Utils/YAML.swift
```diff
@@ -14,17 +14,17 @@ public enum YAML {
   ///
   /// - parameter path: The path to the YAML file
   /// - returns: The decoded document
-  public static func read(path: Path) throws -> Any? {
+  public static func read(path: Path, env: [String: String] = [:]) throws -> Any? {
```
AliSoftware, line 17: @Lutzifer Just realized while re-reading that PR that we forgot to update the doc-comments to add `- parameter env: ...` description 😉

Sources/SwiftGenKit/Utils/YAML.swift
```diff
@@ -14,17 +14,17 @@ public enum YAML {
...
+    return try decode(string: contents, env: env)
   }
 
   /// Decode the contents of YAML string (only the first document).
   ///
   /// - parameter string: The YAML string
   /// - returns: The decoded document
-  public static func decode(string: String) throws -> Any? {
-    return try Yams.load(yaml: string)
+  public static func decode(string: String, env: [String: String] = [:]) throws -> Any? {
```
AliSoftware, line 26: and same here 😉

PR #567, opened 2018-12-27
Core Data: optional scalar attributes

### LINKED ISSUE #565: Core Data: optional scalar attributes
Currently we ignore optionality for scalar attributes, which isn't ideal.

To add support for this, and to do it correctly, we'll need something like this for each optional scalar attribute (based on mogenerator):
```swift
public var counter: Int16? {
  get {
    let key = "counter"
    willAccessValue(forKey: key)

    return primitiveValue(forKey: key) as? Int16
  }
  set {
    let key = "counter"
    willChangeValue(forKey: key)

    setPrimitiveValue(newValue, forKey: key)
  }
}
```

Note that:

- We can't use `#keyPath(counter)` because it isn't an obj-c property.
- We can't make it an `@objc` property because `Int16?` can't be represented in obj-c.
- We can't use Swift key paths, because we can't access the string representation (yet): https://bugs.swift.org/browse/SR-5220

Reference mogenerator template: https://github.com/rentzsch/mogenerator/blob/swift42/templates/machine.swift.motemplate
Article that explains why we can't just use `@NSManaged`:
http://aplus.rs/2017/handling-core-data-optional-scalar-attributes/

djbe:
Closes #565.

Implements the changes discussed in #565, to correctly handle optional scalar attributes in a swifty way.

### DIFF
--- a/Tests/Fixtures/Generated/CoreData/swift4-context-defaults.swift
+++ b/Tests/Fixtures/Generated/CoreData/swift4-context-defaults.swift
@@ -26,8 +26,8 @@ internal class AbstractEntity: NSManagedObject {
     return NSFetchRequest<AbstractEntity>(entityName: entityName())
   }
 
-  // swiftlint:disable implicitly_unwrapped_optional
-  // swiftlint:enable implicitly_unwrapped_optional
+  // swiftlint:disable implicitly_unwrapped_optional discouraged_optional_boolean
+  // swiftlint:enable implicitly_unwrapped_optional discouraged_optional_boolean
 }
 
 // MARK: - ChildEntity
@@ -46,8 +46,8 @@ internal class ChildEntity: MainEntity {
     return NSFetchRequest<ChildEntity>(entityName: entityName())
   }
 
-  // swiftlint:disable implicitly_unwrapped_optional
-  // swiftlint:enable implicitly_unwrapped_optional
+  // swiftlint:disable implicitly_unwrapped_optional discouraged_optional_boolean
+  // swiftlint:enable implicitly_unwrapped_optional discouraged_optional_boolean
 }
 
 // MARK: - MainEntity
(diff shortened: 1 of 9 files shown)


Tests/Fixtures/Generated/CoreData/swift4-context-defaults.swift
```diff
@@ -70,11 +70,56 @@ internal class MainEntity: NSManagedObject {
   @NSManaged internal var boolean: Bool
   @NSManaged internal var date: Date?
   @NSManaged internal var decimal: NSDecimalNumber?
-  @NSManaged internal var double: Double
+  internal var double: Double? {
```
Liquidsoul: If I understand the code correctly, the optional version is:
```swift
@NSManaged internal var double: Double?
```
This does not need the same custom getter/setter?

Maybe we should have `optional(Double|Int16|Int64)` attributes in the tests to show the corresponding generated code?

djbe: Note that I changed the model so that some of the scalar attributes are optionals (see https://github.com/SwiftGen/SwiftGen/pull/567/files#diff-a8d5cc54c3ee6120761c601a66578206L18).

As I understand it, custom setters/getters are needed for scalar optionals as objc can't represent them.

djbe: See http://aplus.rs/2017/handling-core-data-optional-scalar-attributes/

djbe:
@Liquidsoul I've made some more test attributes as requested + simplified the optional code check in the template.

djbe:
Seeing as there hasn't been much feedback here (or really in the related issue), and since it's a small improvement, I'm going to assume it's good to go and merge this soon.

Issue #547, opened 2018-10-08
SwiftGen config warnings are invisible in Xcode


Igor-Palaguta:
SwiftGen v6 warnings are not shown as warnings in Xcode 10. Probably warning message should start immediately with `warning:` and not with `swiftgen: warning:`

e.g. Xcode output without exclamation sign
<img width="629" alt="screen shot 2018-10-08 at 18 00 39" src="https://user-images.githubusercontent.com/800044/46616886-2da74c00-cb24-11e8-93ce-e15c7b1429da.png">

djbe:
Refs https://github.com/SwiftGen/SwiftGen/pull/533#discussion_r222101611.

AliSoftware:
The warnings generated by `swiftlint` for example start with the path to the file, with line and column if available, then `warning`, so `path:line:col: warning: message` or `path:line: warning: message` or `path: warning: message`

I'm not sure then why those don't show up in Xcode, as if I create a Script build phase to `echo` those same messages, they do show up correctly:

<img width="1620" alt="xcode-warnings-format" src="https://user-images.githubusercontent.com/216089/46626959-3c4c2e00-cb39-11e8-895b-ae83fa6e2316.png">

I think the issue might not be related to the line starting with `warning:` or not (as you see in my attached screenshot, the same output works in my example), but probably because we might print those messages in `stderr` (and not `stdout`) iirc?

djbe:
SwiftLint outputs config errors to `stderr`, and those will be invisible to Xcode. SwiftLint warnings are actually just normal output (they don't have deprecation notices afaik). 

We may want to change our warning loggint to `stdout`:
https://github.com/SwiftGen/SwiftGen/blob/master/Sources/SwiftGen/Logs.swift#L49

AliSoftware:
Yeah I agree. I think we even could improve our log enum to take optional associated values for file and line and if provided we can include it in the output.

Might even make that conform to `Swift.Error` so we could throw them idk 🤔

djbe:
Or we could check for environment variables to detect Xcode, and route everything to `stdout` in that case. From a quick check we could use the existence of something like `XCODE_VERSION_ACTUAL`.

AliSoftware:
```
struct LogMessage: Swift.Error {
  let file: String?
  let line: UInt?
  let kind: LogLevel
  let message: String

  init(kind: LogLevel, message: String, file: String? = nil, line: UInt? = nil) { … }
[...]

PR #570, opened 2019-01-01
Filter input files

### LINKED ISSUE #383: Add parameter to filter (include/exclude) files
Param to be able to filter files to parse in or out.

This would mitigate the problem about making `ib` scan XIBs and Storyboards for use cases when you only want storyboard scene constants and don't need to parse XIBs, but they would be parsed anyway and slightly impact performance on every build. Then `storyboards` would become an alias of `ib --filter '*.storyboards'`

The name of the argument is up for debate. `--filter` might be clearer than `--exclude`, because users might otherwise expect a `--include` parameter, which I don't know how we'd add.

djbe:
Fixes #383.

Config entries (or CLI invocations) now accept a `filter` flag (`-f`) that allows applying a regex to input paths (or the contents of directories). Each parser defines a default filter, which is also shown in the help output as a bonus.

One (almost fortunate) consequence of this PR is that every command now accepts a directory, and recurse into its contents as needed. Note that this PR partially replaces #371 and #32 (directory input for strings command). We may still want to add logic to detect multiple `lproj`s. 

Todo:
- [x] Write documentation about new `filter` config option
- [x] Write documentation about new `filter` CLI option
- [x] Tests maybe?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@ _None_
 * Expand environment variables in YAML files.  
   [Wolfgang Lutz](https://github.com/lutzifer)
   [#355](https://github.com/SwiftGen/SwiftGen/issues/#355)
+* Each command now accepts a `filter` option, which accepts a regular expression for filtering input paths. The filter is applied to individual paths as well as when the command recurses into directories.  
+  [David Jennes](https://github.com/djbe)
+  [#383](https://github.com/SwiftGen/SwiftGen/issues/383)
+  [#570](https://github.com/SwiftGen/SwiftGen/pull/570)
 
 ### Bug Fixes
 
--- a/Documentation/ConfigFile.md
+++ b/Documentation/ConfigFile.md
@@ -14,7 +14,8 @@ The configuration file is a YAML file structured like this (example):
 input_dir: Sources/Resources
 output_dir: Sources/Generated/
 strings:
-  inputs: Base.lproj/Localizable.strings
+  inputs: Base.lproj
+  filter: .*\.strings
   outputs:
     - templateName: structured-swift4
       output: L10n-Constants.swift
@@ -61,6 +62,7 @@ Each key corresponding to a SwiftGen subcommands (`colors`, `coredata`, `fonts`,
 | Subkey | Type | Description |
 |--------|------|-------------|
 | `inputs` | Path or Array of Paths | The file(s)/dir(s) to parse (e.g. the path to your assets catalog for the `xcassets` command, or your `Localizable.strings` file for the `strings` command, etc). |
+| `filter` | Regular Expression | The regular expression to apply to each input path, only paths matching the given filter will be used. Note that each command has a default built-in filter, which you can override with this option. |
 | `outputs` | Array | A list of output descriptions, composed of a template and a file output. |
 | `paths` | Path or Array of Paths | **Deprecated** The file(s)/dir(s) to parse (e.g. the path to your assets catalog for the `xcassets` command, or your `Localizable.strings` file for the `strings` command, etc). |
 | `templateName` | String | **Deprecated** The name of the template to use. If you provide a value for this, you shouldn't also provide a value for `templatePath`. |
--- a/Documentation/ConfigFile.md
+++ b/Documentation/ConfigFile.md
@@ -62,7 +62,7 @@ Each key corresponding to a SwiftGen subcommands (`colors`, `coredata`, `fonts`,
 | Subkey | Type | Description |
 |--------|------|-------------|
 | `inputs` | Path or Array of Paths | The file(s)/dir(s) to parse (e.g. the path to your assets catalog for the `xcassets` command, or your `Localizable.strings` file for the `strings` command, etc). |
-| `filter` | Regular Expression | The regular expression to apply to each input path, only paths matching the given filter will be used. Note that each command has a default built-in filter, which you can override with this option. |
+| `filter` | Regular Expression | The regular expression to apply to each input path, only paths matching the given filter will be used. Filters are applied to actual (relative) paths, not just the filename. Note that each command has a default built-in filter, which you can override with this option. |
 | `outputs` | Array | A list of output descriptions, composed of a template and a file output. |
 | `paths` | Path or Array of Paths | **Deprecated** The file(s)/dir(s) to parse (e.g. the path to your assets catalog for the `xcassets` command, or your `Localizable.strings` file for the `strings` command, etc). |
 | `templateName` | String | **Deprecated** The name of the template to use. If you provide a value for this, you shouldn't also provide a value for `templatePath`. |
@@ -70,7 +70,7 @@ Each key corresponding to a SwiftGen subcommands (`colors`, `coredata`, `fonts`,
 | `output` | Path | **Deprecated** The path of the output file to generate. _(Note: Any intermediate directory up to this file must already exist.)_ |
 | `params` | Dictionary | **Deprecated** Any optional parameter you want to pass to the template (similarly to `--param` in the CLI). |
 
-> 💡 Note: For custom filters, use `.+` to match multiple characters (at least one), and don't forget to escape the dot (`\.`) if you want to match a literal dot like for an extension. Add `$` at the end to ensure the path ends with the extension you want. For example, use `.+\.xib$` to match files with a `.xib` extension. Use a tool such as [RegExr](https://regexr.com) to ensure you're using a valid regular expression.
+> 💡 Note: For custom filters, use `.+` to match multiple characters (at least one), and don't forget to escape the dot (`\.`) if you want to match a literal dot like for an extension. Add `$` at the end to ensure the path ends with the extension you want. Regular expressions will be case sensitive by default, and not anchored to the start/end of a path. For example, use `.+\.xib$` to match files with a `.xib` extension. Use a tool such as [RegExr](https://regexr.com) to ensure you're using a valid regular expression.
 
 The `outputs` parameter accepts either a dictionary, or an array of dictionaries, with the keys described below. Each such "output" will take the input files, and use the output template to generate a file at the given output path. This allows you to generate multiple outputs for the same input files (which will only be parsed once).
 
--- a/Documentation/ConfigFile.md
+++ b/Documentation/ConfigFile.md
@@ -15,7 +15,7 @@ input_dir: Sources/Resources
 output_dir: Sources/Generated/
 strings:
   inputs: Base.lproj
-  filter: .*\.strings
+  filter: .+\.strings$
   outputs:
     - templateName: structured-swift4
       output: L10n-Constants.swift
@@ -36,7 +36,7 @@ xcassets:
 
 ℹ️ All _relative_ paths specified in the configuration file (`input_dir`, `output_dir`, `inputs`, `templatePath`, `output`) are relative to the location of the _configuration file_ itself.
 
-> 💡 We advise against using _absolute_ paths — starting with `/` — in the configuration file, so that they won't rely on where the project was cloned on your machine._
+> 💡 We advise against using _absolute_ paths — starting with `/` — in the configuration file, so that they won't rely on where the project was cloned on your machine.
 
 Here's a quick description of all the possible _root_ keys. All of them are optional.
 
@@ -69,6 +69,8 @@ Each key corresponding to a SwiftGen subcommands (`colors`, `coredata`, `fonts`,
 | `templatePath` | Path | **Deprecated** The path to the template to use. If you provide a value for this, you shouldn't also provide a value for `templateName`. |
 | `output` | Path | **Deprecated** The path of the output file to generate. _(Note: Any intermediate directory up to this file must already exist.)_ |
 | `params` | Dictionary | **Deprecated** Any optional parameter you want to pass to the template (similarly to `--param` in the CLI). |
+
+> 💡 Note: For custom filters, use `.+` to match multiple characters (at least one), and don't forget to escape the dot (`\.`) if you want to match a literal dot like for an extension. Add `$` at the end to ensure the path ends with the extension you want. For example, use `.+\.xib$` to match files with a `.xib` extension.
 
 The `outputs` parameter accepts either a dictionary, or an array of dictionaries, with the keys described below. Each such "output" will take the input files, and use the output template to generate a file at the given output path. This allows you to generate multiple outputs for the same input files (which will only be parsed once).
 
--- a/Documentation/ConfigFile.md
+++ b/Documentation/ConfigFile.md
@@ -70,7 +70,7 @@ Each key corresponding to a SwiftGen subcommands (`colors`, `coredata`, `fonts`,
 | `output` | Path | **Deprecated** The path of the output file to generate. _(Note: Any intermediate directory up to this file must already exist.)_ |
 | `params` | Dictionary | **Deprecated** Any optional parameter you want to pass to the template (similarly to `--param` in the CLI). |
 
-> 💡 Note: For custom filters, use `.+` to match multiple characters (at least one), and don't forget to escape the dot (`\.`) if you want to match a literal dot like for an extension. Add `$` at the end to ensure the path ends with the extension you want. For example, use `.+\.xib$` to match files with a `.xib` extension.
+> 💡 Note: For custom filters, use `.+` to match multiple characters (at least one), and don't forget to escape the dot (`\.`) if you want to match a literal dot like for an extension. Add `$` at the end to ensure the path ends with the extension you want. For example, use `.+\.xib$` to match files with a `.xib` extension. Use a tool such as [RegExr](https://regexr.com) to ensure you're using a valid regular expression.
 
 The `outputs` parameter accepts either a dictionary, or an array of dictionaries, with the keys described below. Each such "output" will take the input files, and use the output template to generate a file at the given output path. This allows you to generate multiple outputs for the same input files (which will only be parsed once).
 
--- a/README.md
+++ b/README.md
@@ -176,7 +176,8 @@ To use SwiftGen, simply create a `swiftgen.yml` YAML file to list all the subcom
 
 ```yaml
 strings:
-  inputs: Resources/Base.lproj/Localizable.strings
+  inputs: Resources/Base.lproj
+  filter: Localizable.*\.strings
   outputs:
     - templateName: structured-swift4
       output: Generated/strings.swift
@@ -223,6 +224,7 @@ Each subcommand generally accepts the same options and syntax, and they mirror t
 * `--templateName NAME` or `-n NAME`: define the Stencil template to use (by name, see [here for more info](Documentation/templates)) to generate the output.
 * `--templatePath PATH` or `-p PATH`: define the Stencil template to use, using a full path.
 * Note: you should specify one and only one template when invoking SwiftGen. You have to use either `-t` or `-p` but should not use both at the same time (it wouldn't make sense anyway and you'll get an error if you try)
+* `--filter REGEX` or `-f REGEX`: the filter to apply to each input path. Each command has a default filter that you can override with this option.
 * Each command supports multiple input files (or directories where applicable).
 * You can always use the `--help` flag to see what options a command accept, e.g. `swiftgen xcassets --help`.
 
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ To use SwiftGen, simply create a `swiftgen.yml` YAML file to list all the subcom
 ```yaml
 strings:
   inputs: Resources/Base.lproj
-  filter: Localizable.*\.strings
+  filter: .*\.strings
   outputs:
     - templateName: structured-swift4
       output: Generated/strings.swift
@@ -225,6 +225,7 @@ Each subcommand generally accepts the same options and syntax, and they mirror t
 * `--templatePath PATH` or `-p PATH`: define the Stencil template to use, using a full path.
 * Note: you should specify one and only one template when invoking SwiftGen. You have to use either `-t` or `-p` but should not use both at the same time (it wouldn't make sense anyway and you'll get an error if you try)
 * `--filter REGEX` or `-f REGEX`: the filter to apply to each input path. Each command has a default filter that you can override with this option.
+* Note: use `.*` to match any characters, and don't forget to escape the dot (`\.`) if you want to match a literal dot like for an extension. For example, use `.*\.xib` to match files with a `.xib` extension.
 * Each command supports multiple input files (or directories where applicable).
 * You can always use the `--help` flag to see what options a command accept, e.g. `swiftgen xcassets --help`.
 
--- a/README.md
+++ b/README.md
@@ -225,7 +225,7 @@ Each subcommand generally accepts the same options and syntax, and they mirror t
 * `--templatePath PATH` or `-p PATH`: define the Stencil template to use, using a full path.
 * Note: you should specify one and only one template when invoking SwiftGen. You have to use either `-t` or `-p` but should not use both at the same time (it wouldn't make sense anyway and you'll get an error if you try)
 * `--filter REGEX` or `-f REGEX`: the filter to apply to each input path. Each command has a default filter that you can override with this option.
-* Note: use `.+` to match multiple characters (at least one), and don't forget to escape the dot (`\.`) if you want to match a literal dot like for an extension. Add `$` at the end to ensure the path ends with the extension you want. For example, use `.+\.xib$` to match files with a `.xib` extension.
+* Note: use `.+` to match multiple characters (at least one), and don't forget to escape the dot (`\.`) if you want to match a literal dot like for an extension. Add `$` at the end to ensure the path ends with the extension you want. For example, use `.+\.xib$` to match files with a `.xib` extension. Use a tool such as [RegExr](https://regexr.com) to ensure you're using a valid regular expression.
 * Each command supports multiple input files (or directories where applicable).
 * You can always use the `--help` flag to see what options a command accept, e.g. `swiftgen xcassets --help`.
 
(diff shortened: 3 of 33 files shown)


CHANGELOG.md
```diff
@@ -22,6 +22,10 @@ _None_
 * Expand environment variables in YAML files.  
   [Wolfgang Lutz](https://github.com/lutzifer)
   [#355](https://github.com/SwiftGen/SwiftGen/issues/#355)
+* Each command now accepts a `filter` option, which accepts a regular expression for filtering input paths. The filter is applied to individual paths as well as when the command recurses into directories.
```
  <tbody>
    <tr>
      <td>:warning:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/SwiftGen/SwiftGen/blob/10765608d4328beb0bb5d658362dd7eb6e4d6d49/CHANGELOG.md#L25"></span>Line describing your entry should end with a period and 2 spaces.</td>
    </tr>
  </tbody>
</table>

SwiftGen-Eve:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Found some warnings in CHANGELOG.md</td>
    </tr>
  </tbody>
</table>

---

</p>

CHANGELOG.md
djbe: Good bot!

README.md
```diff
@@ -176,7 +176,8 @@ To use SwiftGen, simply create a `swiftgen.yml` YAML file to list all the subcom
 
 ```yaml
 strings:
-  inputs: Resources/Base.lproj/Localizable.strings
+  inputs: Resources/Base.lproj
+  filter: Localizable.*\.strings
```
AliSoftware: This filter example is quite confusing. Either you want to select all the `.strings` files, so use `.*\.strings`, or you want to select all `Localizable.strings` files, so use `Localizable\.strings`… But it's quite uncommon to have `LocalizableFoo.strings` and `LocalizableBar.string` and `Localizable.*\.strings` here, so maybe not the best example here. Maybe use the same example as you did in `ConfigFile.md`?

README.md
```diff
@@ -223,6 +224,7 @@ Each subcommand generally accepts the same options and syntax, and they mirror t
 * `--templateName NAME` or `-n NAME`: define the Stencil template to use (by name, see [here for more info](Documentation/templates)) to generate the output.
 * `--templatePath PATH` or `-p PATH`: define the Stencil template to use, using a full path.
 * Note: you should specify one and only one template when invoking SwiftGen. You have to use either `-t` or `-p` but should not use both at the same time (it wouldn't make sense anyway and you'll get an error if you try)
+* `--filter REGEX` or `-f REGEX`: the filter to apply to each input path. Each command has a default filter that you can override with this option.
```
AliSoftware: Not sure it's worth it, and people are supposed to know how RegExes work so it's not our job to explain it to them, but given in those filters it will be pretty common to filter by extension, it could be interesting here to just add a quick reminder sentence about "e.g. use `.*` to match any characters, and don't forget to escape the dot (`\.`) if you want to match a literal dot like for an extension. For example, use `.*\.xib` to match files with a `.xib` extension."

This could be added as a `* Note:` similarly to what we have for `--templateName` vs `--templatePath` just above

Sources/SwiftGen/Commander/ParserCLICommands.swift
```diff
@@ -43,24 +43,35 @@ private let paramsOption = VariadicOption<String>(
...
+    ) { output, oldTemplateName, templateName, templatePath, parameters, filter, paths in
       try ErrorPrettifier.execute {
         let parser = try self.parserType.init(options: [:]) { msg, _, _ in
           logMessage(.warning, msg)
         }
-        try parser.parse(paths: paths)
+        let filter = try Filter(pattern: filter)
+        try parser.searchAndParse(paths: paths, filter: filter)
 
-        let resolvedlTemplateName = templateName.isEmpty ? oldTemplateName : templateName
```
AliSoftware: 😅

Sources/SwiftGen/Config/ConfigEntry.swift
```diff
@@ -15,6 +15,7 @@ import PathKit
 struct ConfigEntry {
   enum Keys {
     static let inputs = "inputs"
+    static let filter = "filter"
     static let outputs = "outputs"
 
     // Legacy: remove this once we stop supporting the output key at subcommand level
```
AliSoftware: Unrelated to this PR, but where are we on versioning, will the next version be a minor or a major one? Could be the occasion with the new CoreData support and since the deprecation has been present for a while now to cleanup those…?

Sources/SwiftGen/Config/ConfigOutput.swift
```diff
@@ -52,21 +52,23 @@ extension ConfigEntryOutput {
 /// Convert to CommandLine-equivalent string (for verbose mode, printing linting info, …)
 ///
 extension ConfigEntryOutput {
-  func commandLine(forCommand cmd: String, inputs: [Path]) -> String {
+  func commandLine(forCommand cmd: String, inputs: [Path], filter: String?) -> String {
     let tplFlag: String = {
```
AliSoftware: Occasion to rename `s/tplFlag/templateFlag/`? That short name just felt strange when re-reading this code…

Sources/SwiftGen/ParserCLI.swift
```diff
@@ -55,7 +55,7 @@ let allParserCommands: [ParserCLI] = [
     pathDescription: "Directory to scan for .storyboard files. Can also be a path to a single .storyboard"
   ),
   .init(
-    parserType: Yaml.Parser.self,
+    parserType: JSON.Parser.self,
```
AliSoftware: Was there some legacy error there? (I honestly couldn't remember if we have two really distinct parsers for JSON vs Yaml now or only one for both)

Sources/SwiftGenKit/Parsers/Colors/ColorsParser.swift
```diff
@@ -34,23 +34,33 @@ public enum Colors {
...
     public init(options: [String: Any] = [:], warningHandler: Parser.MessageHandler? = nil) {
       self.warningHandler = warningHandler
 
-      register(parser: CLRFileParser.self)
-      register(parser: JSONFileParser.self)
-      register(parser: TextFileParser.self)
-      register(parser: XMLFileParser.self)
+      for parser in Parser.subParsers {
+        register(parser: parser)
+      }
```
AliSoftware: I don't know why, but I'd have seen a `subParsers.forEach` there instead… matter of taste I guess 😛

Sources/SwiftGenKit/Parsers/Colors/ColorsParser.swift
```diff
@@ -34,23 +34,33 @@ public enum Colors {
...
-    public func parse(path: Path) throws {
+    public static var defaultFilter: String {
+      let extensions = Parser.subParsers.flatMap { $0.extensions }.sorted().joined(separator: "|")
+      return ".*\\.(?i:\(extensions))"
+    }
+
+    public func parse(path: Path, relativeTo parent: Path) throws {
       guard let parserType = parsers[path.extension?.lowercased() ?? ""] else {
-        throw ParserError.unsupportedFileType(path: path, supported: Array(parsers.keys))
+        throw ParserError.unsupportedFileType(path: path, supported: Array(parsers.keys.sorted()))
```
AliSoftware: I wonder if `.sorted()` on a KeysCollection doesn't already return an `Array`?

Sources/SwiftGenKit/Parsers/CoreData/CoreDataParser.swift
```diff
@@ -31,42 +39,40 @@ public enum CoreData {
...
+        try parse(path: modelPath, relativeTo: parent)
+      } else if path.extension == Constants.modelExtension {
         try addModel(path: path)
       } else {
-        let dirChildren = path.iterateChildren(options: [.skipsHiddenFiles, .skipsPackageDescendants])
-
-        for file in dirChildren where file.extension == "xcdatamodeld" || file.extension == "xcdatamodel" {
-          try parse(path: file)
-        }
+        throw ParserError.invalidFormat(reason: "Unknown file type for \(path)")
```
AliSoftware: What happens if the user overrides the defaultFilter with their own and their filter doesn't limit to only those two extensions, like use `UserModel.*`, and in the directory they scan they have `UserModel1.xcdatamodel`, `UserModel2.xcdatamodel`, but also `UserModel3.json` and `UserModel4.xcdatamodel`? I mean, just wanted to be sure that the `ParserError` only interrupts the parsing of an unsupported file like `UserModel3.json` but doesn't interrupt the whole scan of the rest of the directory and that `UserModel4.xcdatamodel` still gets a chance to be parsed :wink:

Sources/SwiftGenKit/Parsers/Filter.swift
```diff
@@ -0,0 +1,45 @@
...
+  }
+}
+
+extension Path {
+  /// Checks if this path matches the given filter (regular expression).
+  ///
+  /// - Parameter filter: The regular expression to match
+  /// - Returns: true if it matches
+  func matches(filter: Filter) -> Bool {
+    let range = NSRange(location: 0, length: string.utf16.count)
```
AliSoftware: Prefer using `NSRange(string.startIndex..<string.endIndex, in: string)`

Sources/SwiftGenKit/Parsers/Parser.swift
```diff
@@ -23,9 +25,23 @@ public protocol Parser {
...
 public extension Parser {
-  func parse(paths: [Path]) throws {
+  func searchAndParse(paths: [Path], filter: Filter) throws {
     for path in paths {
-      try parse(path: path)
+      try searchAndParse(path: path, filter: filter)
+    }
+  }
+
+  func searchAndParse(path: Path, filter: Filter) throws {
```
AliSoftware: The fact that those two methods are not declared in the protocol too means that it will use static dispatch and not dynamic dispatch. I think that there's something with confusing situations arising in such cases, like If one day we have a parser that is a subclass of another parser, it might take the implementation of the superclass (if called on a variable declared as the superclass at compile-time) and not the child class (real type of that variable at runtime)…

Not sure this will really happen, and not sure I remember that thing and behaviour properly, but both something to keep in mind and something to double-check

Sources/SwiftGenKit/Parsers/Parser.swift
```diff
@@ -23,9 +25,23 @@ public protocol Parser {
...
+    if path.matches(filter: filter) {
+      let parentDir = path.absolute().parent()
+      try parse(path: path, relativeTo: parentDir)
+    } else {
+      let dirChildren = path.iterateChildren(options: [.skipsHiddenFiles, .skipsPackageDescendants])
+      let parentDir = path.absolute()
+
+      for path in dirChildren where path.matches(filter: filter) {
+        try parse(path: path, relativeTo: parentDir)
```
AliSoftware: In which case can `relativeTo:` not be the parent of the file passed to `path:`? I mean, if it's always the parent, then why pass it as a parameter instead of having the implementations of `parse(path:)` which  might need to know the parent be the one calling `path.absolute().parent()` to get it?

Sources/SwiftGenKit/Parsers/Plist/PlistParser.swift
```diff
@@ -29,27 +29,10 @@ public enum Plist {
...
-      } else {
-        let dirChildren = path.iterateChildren(options: [.skipsHiddenFiles, .skipsPackageDescendants])
-        let parentDir = path.absolute()
-
-        for file in dirChildren where SupportedTypes.supports(extension: file.extension ?? "") {
-          files.append(try File(path: file, relativeTo: parentDir))
-        }
-      }
+    public func parse(path: Path, relativeTo parent: Path) throws {
+      files.append(try File(path: path, relativeTo: parent))
```
AliSoftware: Wait, what? That's the only thing this parser does? 😅

Sources/SwiftGenKit/Parsers/Yaml/YamlParser.swift
```diff
@@ -22,37 +22,30 @@ public enum Yaml {
 
   // MARK: Yaml File Parser
 
-  public final class Parser: SwiftGenKit.Parser {
+  public class Parser: SwiftGenKit.Parser {
```
AliSoftware: Why no more `final`?

Sources/SwiftGen/Config/ConfigEntry.swift
djbe: 6.1.0, but it's up to us when to remove deprecated stuff. I'd say remove them in one of the next minor versions, or the next major (which might be sooner than later).

Sources/SwiftGenKit/Parsers/Colors/ColorsParser.swift
djbe: Of all the functional stuff in swift (`map`, `filter`, ...), `forEach` is one I never use. 🤷‍♂️

Sources/SwiftGen/ParserCLI.swift
djbe: Nah, it used to be the same parser.

Now, because they define different default filters, the JSON one is a subclass of the YAML parser that only overrides that default filter.

Sources/SwiftGenKit/Parsers/Colors/ColorsParser.swift
djbe: I probably at first had it without the sorted, and added it for the tests (for consistent output)

Sources/SwiftGenKit/Parsers/CoreData/CoreDataParser.swift
djbe: Well, you're feeding invalid data into the **core data** parser, so if you want to override the default filter, maybe ensure you're only passing in valid data?

This is something that applies to all the parsers.

Sources/SwiftGen/ParserCLI.swift
AliSoftware: Haw, which explains why it's not `final` anymore (comment below). Missed that. :+1:

Sources/SwiftGenKit/Parsers/Parser.swift
djbe: Hmmm, the `searchAndParse` will be static, but `parse(path:` is defined in the protocol, so that should be dispatched dynamically (I think). If not, that'd seem like a swift bug to me? 😕

Sources/SwiftGenKit/Parsers/CoreData/CoreDataParser.swift
AliSoftware: Yeah, of course, but what I meant was, we just need to make sure that a failure when parsing one file doesn't stop the parsing of the rest of the files. If the config file in SwiftGen asks to parse (a directory containing) 5 files, and file 3 is corrupted or errors, we should print the error but still continue parsing files 4 and 5. I think I remember in a previous version of SwiftGen that this was an issue (SwiftGen stopped on the first error it encountered and didn't continue on other files), instead of reporting the error and just skipping the problematic file as intended but continuing on the other files… so just wanted to be sure that behavior was still working properly as expected.

Sources/SwiftGenKit/Parsers/Yaml/YamlParser.swift
djbe: See comment about the JSON parser (defaultFilter) above.

Sources/SwiftGenKit/Parsers/Parser.swift
AliSoftware: Yeah, but I think there is (that bug) indeed. Anyway, worth checking. And tbh, this doesn't hurt much to put them in the protocol anyway, making those implementations in extensions "default implementations of the protocol" (instead of being seen as "additional static methods added to all conforming types" for cases like here when it's not declared as a requirement on the protocol)

Sources/SwiftGenKit/Parsers/Parser.swift
djbe: Take the fonts case. Let's say I `searchAndParse` my `Resources/` folder. My fonts are located inside `Resources/Stuff/Fonts/`, and I'm copying those fonts into my app at that specific path.

The parser would need to preserve paths relative to the `Resources` folder, not the parent of `Fonts` (which'd be `Stuff`).

Sources/SwiftGenKit/Parsers/Parser.swift
AliSoftware: Ok this seems that there's no bug after all, false alarm… I think?

```swift
protocol Parser {
//  func parse(path: String)
}

extension Parser {
  func parse(path: String) {
    print("Parser parsing \(path)")
  }
  func searchAndParse(path: String) {
    print("Search >>>")
    parse(path: path)
    print("<<< Search")
  }
}

class Parent: Parser {
  func parse(path: String) {
    print("Parent parsing \(path)")
  }
}

class Child: Parent {
  override func parse(path: String) {
    print("Child parsing \(path)")
  }
}

let parser: Parent = Child()
parser.parse(path: "path1")
parser.searchAndParse(path: "path2")
```

Output if I DO declare "func parse(path: String)" in the protocol (uncomment it):
```
Child parsing path1
Search >>>
Child parsing path2
<<< Search
```

Output if I DON'T declare "func parse(path: String)" in the protocol (keep it commented):
```
/*
Child parsing path1
Search >>>
Parser parsing path2
<<< Search
*/
```
So in this case, `parseAndSearch` calls the implementation on `Parser`… which calls the `parse` defined on `Parser`, so despite the `parser` variable being a `Child()` and declared as `: Parent`, it calls neither the `parse()` on `Child` nor the one on `Parent`, but the one on the protocol…

But as long as `parse(path: )` is declared as a requirement on the `protocol`, at least this one will be dynamic it seems, so… 😅:+1: ?

README.md
```diff
@@ -225,6 +225,7 @@ Each subcommand generally accepts the same options and syntax, and they mirror t
 * `--templatePath PATH` or `-p PATH`: define the Stencil template to use, using a full path.
 * Note: you should specify one and only one template when invoking SwiftGen. You have to use either `-t` or `-p` but should not use both at the same time (it wouldn't make sense anyway and you'll get an error if you try)
 * `--filter REGEX` or `-f REGEX`: the filter to apply to each input path. Each command has a default filter that you can override with this option.
+* Note: use `.*` to match any characters, and don't forget to escape the dot (`\.`) if you want to match a literal dot like for an extension. For example, use `.*\.xib` to match files with a `.xib` extension.
```
AliSoftware: I just realised: shouldn't we end all of our `defaultFilter` with `$`? (and maybe start it with `^`… or even simpler, don't even include `.*`?

Because our RegExes are not anchored to the whole file name when we match them, right? So it will match if part of the path matches, even if it's in the middle…

For example the `.*\.strings` regex actually matches `Loc.stringsdict` because `.strings` exists right in the middle.
While what we probably want is a RegEx like `.strings$` instead… so maybe we should rewrite all the `defaultFilters`… and also suggest that instead in that `Note` too…

AliSoftware: Also, didn't re-check but, does the `RegEx` match the whole path when filtering, on only the lastPathComponent (e.g. `basename`)? It might be important to decide what we prefer (filtering the basename is more simple and might avoid surprises, might also be less powerful, but regex on the whole path might not be worth it anyway) and specify which we chose in the docs.

Sources/SwiftGenKit/Parsers/Filter.swift
```diff
@@ -0,0 +1,45 @@
...
+}
+
+extension Path {
+  /// Checks if this path matches the given filter (regular expression).
+  ///
+  /// - Parameter filter: The regular expression to match
+  /// - Returns: true if it matches
+  func matches(filter: Filter) -> Bool {
+    let range = NSRange(string.startIndex..<string.endIndex, in: string)
+    return filter.regex.firstMatch(in: string, options: [], range: range) != nil
```
AliSoftware: I think it would be less surprising if we matches only the `basename` of the Path, instead of matching on its full absolute path?

Also, should we anchor the RegEx (there's an `options: […]` for that)? I personally think we shouldn't, and let the user anchor it using `^` and `$` if they need, and that will make easier RegExes for matching extensions using `\.xib$`, but then we should specify that in the docs on the `--filter` param to be precise enough to our users.

README.md
djbe: It matches the whole (relative) path. I'd say keep it, so that users can filter out specific folders if they really want.

You're right that I should add the end token, to ignore `stringsdict` (for now). I'd keep the `.*` to ensure we're matching a filename with that extension (and not something with no name, which could lead to other weird cases).

AliSoftware: Ok but if we match the whole (relative) path and not just the `basename`, that's indeed more powerful, but means that `^` will mean a different thing… and then how do you match "all files beginning with `SG`" for example? You can't use a RegEx like `^SG` because a file with relative path `foo/bar/SGBaz.json` won't match when you match the whole path. You can't use `/SG` either, because files at the root of the scanned directory will be `SGQux.json` and won't have the `/` then…

djbe: `SG[^\/]*\.json` or something like that?

AliSoftware: Nope, that will match "paths containing `SG` followed by anything but a slash, followed by anything, following by `.json`"… so a totally different thing. For example your pattern will match `foo/bar/bazSGqux.json`…

djbe: `^(.+\/|)SG[^\/]*\.json$` then? I dunno, it's late 😆

AliSoftware:
Shouldn't we add tests for `Filter` btw? Not just the lines added in the tests for parsers but tests on its own right, checking matches against some test `Path` to ensure that we don't break if one day we change the regex engine or choose to change between matching the basename vs full pay or change the regex options like case insensitivity, etc?

djbe:
Sure. I wouldn’t go too far though, some tests for full path and case would be good. We don’t want to test the actual regex engine 😄

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Sources/SwiftGenKit/Parsers/CoreData/CoreDataParser.swift
djbe: Right, so as discussed in Slack, we're going to leave the "fail early" system as it is now (unchanged from before this PR), and we'll take a better look at this in a separate issue. Refs #547.

README.md
djbe: Keep in mind that most people probably won't need more than the default provided filters, and if they ever need a custom filter, will probably have enough with some simple regexes with for example fixed folder names. For those one or 2 users that need something really specific, they can have some fun with regexes and ask a question here 🤷‍♂️ 

Let's also not forget that we allow multiple input paths, so that'll also cover a bunch of edge cases.

Sources/SwiftGenKit/Parsers/Filter.swift
djbe: Yeah, keep it unanchored, so that users can go loco if they want.

We're also not setting anything about case (in)sensitivity. Some of our default filters do match case insensitively with the `i` flag: `stuff\.(?i:json)$` matches `stuff.json`, `stuff.Json`, ...

Documentation/ConfigFile.md
```diff
@@ -70,6 +70,8 @@ Each key corresponding to a SwiftGen subcommands (`colors`, `coredata`, `fonts`,
 | `output` | Path | **Deprecated** The path of the output file to generate. _(Note: Any intermediate directory up to this file must already exist.)_ |
 | `params` | Dictionary | **Deprecated** Any optional parameter you want to pass to the template (similarly to `--param` in the CLI). |
 
+> 💡 Note: For custom filters, use `.*` to match any characters, and don't forget to escape the dot (`\.`) if you want to match a literal dot like for an extension. Add `$` at the end to ensure the path ends with the extension you want. For example, use `.*\.xib$` to match files with a `.xib` extension.
```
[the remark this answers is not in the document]
djbe: I've changed it to `.+`, as it better represents what we're trying to achieve here (match a file extension, with a non-empty filename). We always (?) need a non-empty filename, and this is "general" enough to cover 99% of the cases.

If someone really wants to have a `A/B/C/.json`, they can try (probably won't work, depends on the parser).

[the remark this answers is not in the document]
AliSoftware: But again, `.+\.json$` will only be guaranteed to match non-empty filenames if we match on basenames :wink: Again, if we match on the whole path, `.+` will match `A/B/C/` in your example above and `A/B/C/.json` will match the filter anyway… So that `.+` would be enough either. You probably want `[^/]\.json$` instead, so a `.json` at the end of a string preceded by one character there is not a `/` (there might be more characters before, but we don't really care, as long as we match the last character before the extension, we know it's a non-empty filename, so no need to make the RegEx engine backtrack with a `+` or `*` for nothing)

Tests/SwiftGenKitTests/FilterTests.swift
```diff
@@ -0,0 +1,49 @@
...
+    let filter = try Filter(pattern: ".+\\.(?i:json)$")
+
+    XCTAssertTrue(Path("foo.json").matches(filter: filter))
+    XCTAssertTrue(Path("foo.Json").matches(filter: filter))
+    XCTAssertTrue(Path("foo.JSON").matches(filter: filter))
+  }
+
+  func testComplex() throws {
+    // match json files with the filename beginning with "SG"
+    let filter = try Filter(pattern: "^(.+\\/|)SG[^\\/]*\\.json$")
```
AliSoftware: Quite unusual in a RegEx to see `(x|)` to match either x or… nothing. We usually just use `x?` instead to match "maybe x"… So `^(.+/)?SG[^/]*\\.json$` should be enough (btw, why did you escape the `/` in those RegExes?)

djbe: No idea why I escaped them 😆, it was late and I assumed it was some special character for some reason.

djbe: I can't remember if it works correctly without the `(x|)`, will give it a quick test.

djbe:
There, those extra notes should cover it, I think?

PR #573, opened 2019-01-05
Strings: Allow specifying localization macro

### LINKED ISSUE #426: Allow runtime change of locale/language
I am allowing users to change the language of my app.

But if I use the L10n enum. The string will be loaded the first time it's needed, which is as it should be.
However, when the user changes the language of the app, the L10n enum should also be reset.

Would that be possible?

### LINKED ISSUE #468: Strings: Parameter to change `NSLocalizationString` usage with custom mechanism
Hi! I see that by default for localized strings we get this extension:
```
extension L10n {
    let format = NSLocalizedString(key, tableName: table, bundle: Bundle(for: BundleToken.self), comment: "")
    return String(format: format, locale: Locale.current, arguments: args)
  }
}
```
It would be really nice if there was a way to customize like:
```
extension L10n {
    let format = myCustomFunction(key, tableName: table, bundle: Bundle(for: BundleToken.self), comment: "")
    return String(format: format, locale: Locale.current, arguments: args)
  }
}
```
On one of the projects we use plist with XML strings instead and it doesn't work with NSLocalizedString, but we have a custom getter to get the string by the key.

StevenSorial:
Closes #426, Closes #468.

I've added a parameter to the strings templates to allow for runtime changes.

### DIFF
--- a/Documentation/templates/strings/flat-swift3.md
+++ b/Documentation/templates/strings/flat-swift3.md
@@ -22,7 +22,7 @@ You can customize some elements of this template by overriding the following par
 | `enumName` | `L10n` | Allows you to change the name of the generated `enum` containing all string tables. |
 | `noComments` | N/A | Setting this parameter will disable the comments describing the translation of a key. |
 | `publicAccess` | N/A | If set, the generated constants will be marked as `public`. Otherwise, they'll be declared `internal`. |
-| `localizeFunction` | `NSLocalizedString` | Allows you to set your own custom localization function. Your custom function must have the same signature as the one provided by `Foundation`, i.e. `NSLocalizedString(_:tableName:bundle:comment:)` |
+| `localizeFunction` | `NSLocalizedString` | Allows you to set your own custom localization function. Your custom function must have the same signature as the one provided by `Foundation`, i.e. `yourFunctionName(_:tableName:bundle:comment:)` |
 
 ## Generated Code
 
--- a/Documentation/templates/strings/flat-swift3.md
+++ b/Documentation/templates/strings/flat-swift3.md
@@ -22,6 +22,7 @@ You can customize some elements of this template by overriding the following par
 | `enumName` | `L10n` | Allows you to change the name of the generated `enum` containing all string tables. |
 | `noComments` | N/A | Setting this parameter will disable the comments describing the translation of a key. |
 | `publicAccess` | N/A | If set, the generated constants will be marked as `public`. Otherwise, they'll be declared `internal`. |
+| `localizeFunction` | `NSLocalizedString` | Allows you to set your own custom localization function. Your custom function must have the same signature as the one provided by `Foundation`, i.e. `NSLocalizedString(_:tableName:bundle:comment:)` |
 
 ## Generated Code
 
--- a/Documentation/templates/strings/flat-swift4.md
+++ b/Documentation/templates/strings/flat-swift4.md
@@ -21,7 +21,7 @@ You can customize some elements of this template by overriding the following par
 | `enumName` | `L10n` | Allows you to change the name of the generated `enum` containing all string tables. |
 | `noComments` | N/A | Setting this parameter will disable the comments describing the translation of a key. |
 | `publicAccess` | N/A | If set, the generated constants will be marked as `public`. Otherwise, they'll be declared `internal`. |
-| `localizeFunction` | `NSLocalizedString` | Allows you to set your own custom localization function. Your custom function must have the same signature as the one provided by `Foundation`, i.e. `NSLocalizedString(_:tableName:bundle:comment:)` |
+| `localizeFunction` | `NSLocalizedString` | Allows you to set your own custom localization function. Your custom function must have the same signature as the one provided by `Foundation`, i.e. `yourFunctionName(_:tableName:bundle:comment:)` |
 
 ## Generated Code
 
--- a/Documentation/templates/strings/flat-swift4.md
+++ b/Documentation/templates/strings/flat-swift4.md
@@ -21,6 +21,7 @@ You can customize some elements of this template by overriding the following par
 | `enumName` | `L10n` | Allows you to change the name of the generated `enum` containing all string tables. |
 | `noComments` | N/A | Setting this parameter will disable the comments describing the translation of a key. |
 | `publicAccess` | N/A | If set, the generated constants will be marked as `public`. Otherwise, they'll be declared `internal`. |
+| `localizeFunction` | `NSLocalizedString` | Allows you to set your own custom localization function. Your custom function must have the same signature as the one provided by `Foundation`, i.e. `NSLocalizedString(_:tableName:bundle:comment:)` |
 
 ## Generated Code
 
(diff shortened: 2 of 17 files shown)


SwiftGen-Eve:

---

</p>

Documentation/templates/strings/flat-swift4.md
```diff
@@ -21,6 +21,7 @@ You can customize some elements of this template by overriding the following par
 | `enumName` | `L10n` | Allows you to change the name of the generated `enum` containing all string tables. |
 | `noComments` | N/A | Setting this parameter will disable the comments describing the translation of a key. |
 | `publicAccess` | N/A | If set, the generated constants will be marked as `public`. Otherwise, they'll be declared `internal`. |
+| `runtime` | N/A | If set to true, the generated strings will be computed instead of static. |
```
AliSoftware: The term "static" here is probably not the most appropriate 😉 they are still `static` in the sense `static var` (properties of the _type_ not the _instance_), they are just recomputed at each call instead of constants lazily computed only once.

AliSoftware (commented):
I'm not sure I understand how you would use that template, how making them computed variables, so re-evaluated every time they are accessed, is sufficient to support runtime change of the locale

If the user manually changes the locale of his or her phone in the system preferences your app is killed and rebooted anyway
And if you meant changing the current Locale programmatically or basing your UI on a locale chosen inside the app that might be one different from the system-wide currentLocale, all the solutions I've seen to do that involve also using a custom translation function instead of `NSLocalizedString` (which uses the system-wide currentLocale) we're currently using, right?

AliSoftware:
Ok I've re-read the associated issue and saw https://github.com/SwiftGen/SwiftGen/issues/426#issuecomment-387453860

So this means that:
1) this template will also need people to know about this additional gist to add to their project for this to work
2) this is based on a hack and manipulating the ObjC runtime to replace the NSBundle class

If you really need such a feature I'd instead recommend to get rid of your custom gist hacking the runtime, and instead allow the template to use a custom function instead of `NSLocalizedString` (Using a `param` as well, or maybe regrouping the feature into only one param that would give the name of the function to use instead of `NSLocalizedString` and if it's not empty, then use computed variables too?)

Then in your app you'd be able to use whatever function for localizing strings, in your case you can just pass `BundleEx.main.localizedString(_:table:…)` so that SwiftGen uses this instead of the main Bundle's one, and you'd be able to totally get rid of your gist and your ObjC runtime hack switching the classes.

StevenMagdy:
perfect suggestion, I got what you mean but I don't know how to implement it correctly.

is that what you had in mind?
[template](https://gist.github.com/StevenMagdy/2dd7344654f8801b80acd61154ac4995)
config:
```
strings:
  inputs: App/Base.lproj/Localizable.strings
  outputs:
    - templatePath: structured-swift4.stencil
      output: Strings.swift
      params:
          localizeFunction: "BundleEx.main.localizedString(forKey: key, value: nil, table: table)"
```

AliSoftware:
Yep indeed that's what I meant 👍

In practice for readability though I think the `localizeFunction` would be the name of a global to call, defaulting to `"NSLocalizedStringFromTable"`. In your case you'd instead set it to `CustomLocStr` and define in your code somewhere a global function like:
```
func CustomLocStr(key: String, table: String) -> String {
  return … // Do your stuff with BundleEx here
}
```

And in the template you'll replace the hard coded "NSLocalizedStringFromTable" in `func tr` with `{{ param.localizeFunction | default:"NSLocalizedStringFromTable" }}` 😉


_Disclaimer: I'm on my phone so not easy for me to browse the existing code and template, I don't exactly remember how our `func tr` is written, but you see the idea I hope_

djbe:
Keep in mind that Xcode allows defining your own `NSLocalizedString` replacement with the setting `LOCALIZED_STRING_MACRO_NAMES`:
> The base names for NSLocalizedString-like macros or functions used to produce localized strings in source code. The default base names of NSLocalizedString and CFLocalizedString are always considered, even if this setting is empty.

We may want to use a similar(-ish) parameter? And to mirror that functionality, only accept a "base" name such as `"MyLocalizedString"`, which we'd use as:
```swift
let format = {{ param.localizedStringMacro|default"NSLocalizedString" }}(key, tableName: table, bundle: Bundle(for: BundleToken.self), comment: ""))
return String.localizedStringWithFormat(format: format, arguments: args)
```

Note that I used [String.localizedStringWithFormat](https://developer.apple.com/documentation/swift/string/1414192-localizedstringwithformat) which Apple seems to recommend.

StevenMagdy:
> Keep in mind that Xcode allows defining your own `NSLocalizedString` replacement with the setting `LOCALIZED_STRING_MACRO_NAMES`:
> 
> > The base names for NSLocalizedString-like macros or functions used to produce localized strings in source code. The default base names of NSLocalizedString and CFLocalizedString are always considered, even if this setting is empty.
> 
> We may want to use a similar(-ish) parameter? And to mirror that functionality, only accept a "base" name such as `"MyLocalizedString"`, which we'd use as:
> 
> ```swift
> let format = {{ param.localizedStringMacro|default"NSLocalizedString" }}(key, tableName: table, bundle: Bundle(for: BundleToken.self), comment: ""))
> return String.localizedStringWithFormat(format: format, arguments: args)
> ```
> 
> Note that I used [String.localizedStringWithFormat](https://developer.apple.com/documentation/swift/string/1414192-localizedstringwithformat) which Apple seems to recommend.

Now, this is getting too advanced for me 😅.
This is the best(and the least ugly) I could come up with:
[template](https://gist.github.com/StevenMagdy/2dd7344654f8801b80acd61154ac4995)
config:
```
strings:
  inputs: App/Base.lproj/Localizable.strings
  outputs:
    - templatePath: structured-swift4.stencil
      output: Strings.swift
      params:
          localizeFunction: "MyLocalizedString(mykeylabel:mytablelabel:)"
```

StevenMagdy:
@AliSoftware I've pushed discussed changes.

djbe:
This has lingered for quite a while. I've updated it to work with all the recent changes, and tweaked it a bit to be simpler & easier to use.

djbe (approved):

Documentation/templates/strings/flat-swift3.md
```diff
@@ -22,6 +22,7 @@ You can customize some elements of this template by overriding the following par
 | `enumName` | `L10n` | Allows you to change the name of the generated `enum` containing all string tables. |
 | `noComments` | N/A | Setting this parameter will disable the comments describing the translation of a key. |
 | `publicAccess` | N/A | If set, the generated constants will be marked as `public`. Otherwise, they'll be declared `internal`. |
+| `localizeFunction` | `NSLocalizedString` | Allows you to set your own custom localization function. Your custom function must have the same signature as the one provided by `Foundation`, i.e. `NSLocalizedString(_:tableName:bundle:comment:)` |
```
AliSoftware: I'd use "i.e. `yourFunctionName(_:tableName:bundle:comment:)`" here to avoid confusion

templates/strings/flat-swift3.stencil
```diff
@@ -34,6 +34,8 @@ import Foundation
   {{accessModifier}} static func {{string.key|swiftIdentifier:"pretty"|lowerFirstWord|escapeReservedKeywords}}({% call parametersBlock string.types %}) -> String {
     return {{enumName}}.tr("{{table}}", "{{string.key}}", {% call argumentsBlock string.types %})
   }
+  {% elif param.localizeFunction %}
```
AliSoftware, line 37: Worth a stencil comment `{#  custom localization function is mostly used for in-app lang selection,  so we want the loc to be recomputed at each call for those (hence  the computed var) #}`?

AliSoftware (approved):
Aside from small nitpicks on the doc, LGTM

PR #575, opened 2019-01-10
CoreData: add parameter to generate `@objc(...)` annotation

djbe:
This PR does a few things:
- Disables the generation of the `@objc(...)` annotation by default, it isn't really needed for most cases.
- Adds a `generateObjcName` parameter to enable this behaviour again for users that may want it.

We also ensure the module is stripped from the class name, as there's no good way to correctly generate code for this. In code, we have no control over the module in which we'll be defined.

If we keep it, the template would generate `class SwiftGen.ChildEntity: MainEntity {`, which is invalid swift code. And depending on what the user enters in the "Class Name" field (something with a '.') and provide a module, we'd end up with "Foo.Bar.Baz".


SwiftGen-Eve:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

---

</p>

Documentation/templates/coredata/swift3.md
```diff
@@ -19,14 +19,14 @@ You can customize some elements of this template by overriding the following par
 
 | Parameter Name | Default Value | Description |
 | -------------- | ------------- | ----------- |
+| `generateObjcName`| N/A | If set, the generated types will have an `@objc(...)` with their class name. |
```
AliSoftware: an `@objc(…)` **annotation**

templates/coredata/swift4.stencil
```diff
@@ -8,18 +8,24 @@ import Foundation
...
-{% set override %}{% if superclass != "NSManagedObject" %}override {% endif %}{% endset %}
-// MARK: - {{ entityClassName }}
+{% set superclass %}{{ model.entities[entity.superEntity].className|default:"NSManagedObject" }}{% endset %}
+{% set entityClassName %}{{ entity.className|default:"NSManagedObject" }}{% endset %}
+
+// MARK: - {{ entity.name }}
 
+{% if param.generateObjcName %}
+{% if entityClassName|contains:"." %}
+// warning: {{ entityClassName }} cannot be an objc name.
```
AliSoftware: Didn't look at everything but so you have a casr, among the different fixtures, which contain an entityClassName with a dot so we see the warning comment being generated as expected here in those cases?

djbe: No, this is just an extra just in case. The generated code won't even work if this happens, you'd get something like:
```swift
class Foo.Bar: NSManagedObject {
}
```

Which won't work. If the user is trying to create a namespaced type, it'd have to be something like:
```swift
extension Foo
  class Bar: NSManagedObject {
  }
}
```
But we have no way of knowing what the user would even be trying to achieve as we have no info about the type `Foo` (or even deeper levels).

djbe: We could completely disable code generation for these cases, and just generate that comment.

djbe: We now generate a warning for such types that we cannot represent:
> // Warning: 'We.Cannot.Handle.ImpossibleType' cannot be a valid type name, skipping code generation.

djbe:
Well I actually needed this for some projects, so pretty sure it works 😆 
The optional scalars PR changed more, but I also tested that one in those same projects without issues.

We'll see what feedback comes in after release, when there's more widespread use.

PR #581, opened 2019-01-26
CoreData: playgrounds demo page

djbe:


SwiftGen-Eve:

---

</p>

djbe:
True. It's not like we add a new command every day though 🤷‍♂️

AliSoftware:
Yeah, I didn't mean for not forgetting to add a page for new commands, I meant more for keeping the code in them up-to-date when you decide to update the templates and the generated code :wink:

SwiftGen.playground/Pages/CoreData-Demo.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,72 @@
...
+  @NSManaged internal var transient: String?
+  @NSManaged internal var uri: URL?
+  @NSManaged internal var uuid: UUID?
+  // swiftlint:enable implicitly_unwrapped_optional
+}
+
+//: #### Usage Example
+
+let container = createContainer()
+let context = container.viewContext
```
AliSoftware: Maybe add type annotations there since those are methods hidden away for keeping that page simple at least it would help when reading the demo to understand at first sight if the variables are explicitly typed `: NSPersistentContainer` and `: NSManagedObjectContext`

SwiftGen.playground/contents.xcplayground
```diff
@@ -2,6 +2,7 @@
 <playground version='6.0' target-platform='ios' display-mode='rendered'>
     <pages>
         <page name='Colors-Demo'/>
+        <page name='CoreData-Demo'/>
```
AliSoftware: _(commenting here as I can't comment on binary files)_

Is it normal / useful that the `.momd` in the Playground have 2 `.mom` inside it, I'm guessing 2 versions of the DataModel (`Model.mom` and `Model 2.mom`)? Not sure it's worth it for the demo in the playground, but I'm guessing we end up with that because you used the `rake` task to copy the compiled resources in the playground and picked one from the fixtures that had 2 versions… (and I didn't check if we had one in the fixtures with only one version to be used instead). Anyway, not a big deal, just wanted to check if that was normal or not.

Also, I don't remember all the internals of `.momd` but even if there are 2 `.mom` why is there only one `.omo` (I don't remember what that extension/file type is for)

djbe: We only have 1 model file with 2 versions (to ensure we respect the version set in the model).

The rake task doesn't copy anything, this is the output of `momc`, the compiler for core data model files. TBH, we shouldn't touch that output, no idea what breaks if we do.

AliSoftware: Yeah sure. Just surprised there's an `.omo` file for Model 2 only though, but 🤷‍♂️

PR #585, opened 2019-01-30
Fix CLI args in scheme & docs #trivial

AliSoftware:
Some small fixes regarding `-t` => `--templateName`, command line arguments saved in scheme, and other nitpicks

### DIFF
--- a/Documentation/Articles/Xcode-Integration.md
+++ b/Documentation/Articles/Xcode-Integration.md
@@ -48,16 +48,16 @@ Note: Those script phase examples above all assume that you're [using a `swiftge
 
 When adding SwiftGen as a build phase in your project, if instead of using a Config file you invoked subcommands from the command line, then sometimes builds can cancel because Xcode doesn't like source code changes happening mid-build. It can also cause issues with `IBDesignable` views triggering looping builds when opening storyboards.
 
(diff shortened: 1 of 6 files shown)


SwiftGen-Eve:

---

</p>

djbe:
Does the `static` bit not work? 😕 Or is it just style?

Side Note: I'd consider dropping CLI interface (for parsers) for SwiftGen 7.0.

AliSoftware:
This is just style, indeed. At first I wanted to make them all `class let` or `class var` to be consistent. Then Xcode told me that classes can't have `class let`, only `static let` or `class var`. So I could have gone all the way to change them to `class var defaultFilter { … }` computed var at that point, but those classes being `final` I didn't see the point and stopped there.

So you're right that since `JSON` is a  `final` class, `static` vs `class` should be similar there, but I felt it was more consistent to make it a `class var` like the property it was overriding 😄

I agree that CLI interface for parsers might go away at some point. Not sure how we'd write tests in the homebrew formula without it, but we could figure out a solution if it's just for that :wink:

djbe:
Eh, have homebrew use a config file? 😄

djbe (approved):

AliSoftware:
😂 yeah at first I thought "mmmh we're gonna have to replace each test we do in the brew formula, so each invocation with CLI params, with an invocation using their own specific config file, that's gonna need as many config files as invocations we test for in the formula, a lot of fixtures to add just for that, not cool" but then I realized we could just use a single config file generating everything, replacing all the N single invocations with only one 🤦‍♂️ duh

PR #587, opened 2019-02-03
Parser Options

djbe:
Makes the necessary config/cli changes to pass along an `options` dictionary to the parser.

Note: 1st step for #576 (and others).

PR #588, opened 2019-02-04
Strings: custom separator option

### LINKED ISSUE #576: Strings: Structured strings split by '__' instead of '.'
Hi!

is it possible to split the strings by '__' instead of '.'? 
e.g.: "screen_name__header__title" -> L10n.ScreenName.Header.title

We're trying to use the same keys as our Android peers. They say they can't use dots in the keys...

Thanks!

djbe:
Fixes #576.

Couple of things I'm not 100% happy with (from #587):
- Might be easier to work with if options was a `[String: String]` (instead of `[String: Any]`). Couple of ways to achieve this: add another parser to SSK or in SwiftGen, or just downcast the result to `[String: String]` and throw an error if needed.
- Should we throw an error if we get an option we don't support? To achieve this, we'd need some list of acceptable options (keys) for a parser.


SwiftGen-Eve:

---

</p>

Tests/SwiftGenKitTests/StringsTests.swift
```diff
@@ -75,4 +75,12 @@ class StringsTests: XCTestCase {
       XCTFail("Unexpected error occured while parsing: \(error)")
     }
   }
+
+  func testCustomSeparator() throws {
+    let parser = Strings.Parser(options: ["separator": "_"])
```
AliSoftware: I think it could be worth using a multi-character separator for this test, like `["separator": "__"]` (which is the use case that was requested in #576 btw)

AliSoftware (commented):
I think it would still make sense to have `options` be the same type as the `params` for templates, after all here the `separator` option is a string, but for other use cases some options might be `Int` or `Bool`, we don't know yet the use cases we'll have in the future.

But I agree it would be nicer to have options be type-safe, and decide what we do if we encounter an unsupported option. I'd be tempted to say that we `throw` if we encounter an unrecognised option as well as if one of the option doesn't have the expected type.

Instead of having a `[String: Any]` type for options, we could imagine maybe an array of cases from an `enum` with associated values, like `case separator(String)`, but then how are we gonna build those in a generic way from a list of string options (from the CLI or the config files)?

Would be also nice to have a way to list all the supported options and their expected type during `--help`… it's too late now for me to think of a proper solution, but maybe tomorrow I'll think of something.

Tests/SwiftGenKitTests/StringsTests.swift
djbe: Rebased with some test keys in Localizable (some with `__` and some with the old `_`)

djbe:
I know I started the discussion about it, but maybe we should keep the discussion about option-safety to a separate issue/PR.

Documentation/Commands/strings.md
```diff
@@ -38,6 +38,10 @@ You can provide a custom filter using the `filter` option, it accepts any valid
 
 This parser currently doesn't accept any options.
 
+| Option Name | Default Value | Description |
+| -------------- | ------------- | ----------- |
+| `separator` | `.` | Allows you to change the string used to separate keys into structured components (see the explanation above). |
+
```
AliSoftware, line 44: Should we adapt that "explanation above" too?

> The strings file will be converted into a structured tree version, where each string is separated into components by the `.` character (note: you can choose another separator if you need, using the `separator` option, see [Customization] (#customization) below). We call this the `dot syntax`, each component representing a level.

AliSoftware (approved):
Assuming we postpone the refactoring for "safe option parsing + parser options listing in --help" for a separate PR, LGTM. I'd just nitpick the explanation about the dot-syntax to note the `separator` option to make things more clear.

PR #589, opened 2019-02-04
Assets: improve colors performance

### LINKED ISSUE #578: Asset Catalog: default template is very CPU intensive on lookup
Assume I have a color named `myColor`. Accessing `Asset.myColor.color` from the generated assets swift file is very expensive, since this will always look up the color and is not lazily evaluated.

I profiled it and if the color is accessed frequently this incurs a heavy performance penalty.

djbe:
Fixes #578.

To use `lazy var`, had to switch to `final class` with our own `init`. This is better than parsing the color during `init` as we can keep the `@availability` annotation for the `color` property.

### DIFF
--- a/Tests/Fixtures/Generated/XCAssets/swift3-context-all-allValues.swift
+++ b/Tests/Fixtures/Generated/XCAssets/swift3-context-all-allValues.swift
@@ -95,15 +95,17 @@ internal enum Asset {
 
 // MARK: - Implementation Details
 
-internal struct ColorAsset {
+internal final class ColorAsset {
   internal fileprivate(set) var name: String
 
   #if swift(>=3.2)
   @available(iOS 11.0, tvOS 11.0, watchOS 4.0, OSX 10.13, *)
-  internal var color: AssetColorTypeAlias {
-    return AssetColorTypeAlias(asset: self)
-  }
+  internal fileprivate(set) lazy var color: AssetColorTypeAlias = AssetColorTypeAlias(asset: self)
   #endif
+
+  fileprivate init(name: String) {
+    self.name = name
+  }
 }
 
 internal extension AssetColorTypeAlias {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,9 +28,13 @@ _None_
 
 ### Bug Fixes
 
+* Assets: improved the performance for color assets by caching the resolved colors.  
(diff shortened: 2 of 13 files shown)


SwiftGen-Eve:

---

</p>

Tests/Fixtures/Generated/XCAssets/swift3-context-all-allValues.swift
```diff
@@ -95,15 +95,17 @@ internal enum Asset {
...
-internal struct ColorAsset {
+internal final class ColorAsset {
   internal fileprivate(set) var name: String
 
   #if swift(>=3.2)
   @available(iOS 11.0, tvOS 11.0, watchOS 4.0, OSX 10.13, *)
-  internal var color: AssetColorTypeAlias {
-    return AssetColorTypeAlias(asset: self)
-  }
+  internal lazy var color: AssetColorTypeAlias = AssetColorTypeAlias(asset: self)
```
AliSoftware: `private(set)`?

AliSoftware (changes requested):

Tests/Fixtures/Generated/XCAssets/swift3-context-all-allValues.swift
djbe: Good idea.

djbe:
Rebased with changes from feedback.

AliSoftware (approved):
LGTM but I'd love for @eaigner to weight in and confirm it fixes the performance issue they observed.

eaigner:
While I can't speak to the performance aspect right now (the code was changed since i reported the issue), the stencil works.

djbe:
Could you try it with your local generated code, replacing the `struct` with a `final class`, and the computed `var` by a `lazy var`?

eaigner:
I have. But I did not benchmark performance. But since its lazy now, should not be a problem, right?

djbe:
I think it's safe to assume that the `lazy var`s solve the issue. Will merge this PR this evening unless there's more feedback.

eaigner:
When is 6.2 scheduled to release? Would be nice to have this out there. It's almost 2 months now.

djbe:
I've been swamped with work these last few weeks, big upcoming release at work.

TBH, things are ready for release, just some final checks and changelog/docs updates (see #621). Maybe this weekend, but I can't promise anything.

PR #590, opened 2019-02-04
Core Data: `entityName` should be a `class var`

djbe:
Apparently we based ourselves a bit too much on old Mogenerator code 😅

This goes from:
```swift
class func entityName() -> String {
```

To:
```swift
class var entityName: String {
```

I'd honestly consider this a "bug fix", although it should be considered a breaking change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,9 @@ _None_
 
 ### Bug Fixes
 
-_None_
+* Core Data: `entityName` is now correctly a `class var` instead of a `class func`.  
+  [David Jennes](https://github.com/djbe)
+  [#589](https://github.com/SwiftGen/SwiftGen/pull/589)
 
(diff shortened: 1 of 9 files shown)


SwiftGen-Eve:

---

</p>

AliSoftware:
Yeah, it's indeed tempting to only release it as a bugfix version (especially since not many people will have started to use the CoreData parser yet?)

Even after that change there are still some places in the rest of the generated code where `entityName()` is used, hence the compile-output errors.

djbe:
Gah, I was too eager 😆 

Yeah, considering the recent release, this not being "swifty" code, and most people not really using `entityName` directly (we provide a fetchRequest and description getters for this purpose)

djbe:
Rebased with the needed changes.

AliSoftware (approved):

PR #593, opened 2019-02-06
CoreData: enum attributes

### LINKED ISSUE #566: Core Data: enum attributes
Similar to #565, it'd be nice to support `RawRepresentable` enums using:

```swift
public var state: State? {
  get {
    let key = "state"
    willAccessValue(forKey: key)

    guard let primitiveValue = primitiveValue(forKey: key) as? State.RawValue else { return nil }
    return State(rawValue: primitiveValue)
  }
  set {
    let key = "state"
    willChangeValue(forKey: key)

    setPrimitiveValue(newValue?.rawValue, forKey: key)
  }
}
```

Mogenerator uses an "User Info" key for this "attributeRawValueEnumType". The name of such a key is up to us to decide. Gyro uses the key "enumType", although it goes a step further and also generates the enum definition, which is unnecessary (I think).

Reference mogenerator template: https://github.com/rentzsch/mogenerator/blob/swift42/templates/machine.swift.motemplate

djbe:
Fixes #566.

I'm a bit unsure on where to document the "User Info" key `enumType`. Pretty sure it should go in the templates documentation, but under what section, and which format?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,10 @@ _None_
   [David Jennes](https://github.com/djbe)
   [#591](https://github.com/SwiftGen/SwiftGen/issue/591)
   [#592](https://github.com/SwiftGen/SwiftGen/pull/592)
+* Core Data: the built-in templates now support `enum` attributes. They'll check the "User Info" of an attribute for a `enumType` key, which should be set to the enum type name you want to use for that attribute.  
+  [David Jennes](https://github.com/djbe)
+  [#566](https://github.com/SwiftGen/SwiftGen/issue/566)
+  [#593](https://github.com/SwiftGen/SwiftGen/pull/593)
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,15 +13,15 @@ _None_
 * Each command now accepts an `options` dictionary, with which you can set internal parser settings to change its behaviour. See the commands' specific documentation for available options.  
   [David Jennes](https://github.com/djbe)
   [#587](https://github.com/SwiftGen/SwiftGen/pull/587)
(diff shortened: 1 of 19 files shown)


SwiftGen-Eve:

---

</p>

AliSoftware:
I think adding an ad-hoc/dedicated section between "Customisation" and "Generated Code" in the template documentation would be appropriate.

In `gyro` we use `UserInfo` attributes for many things, including supporting enums, but also other stuff, so I think similarly we might have other usages for UserInfo keys in the future. So the section title should probably be "UserInfo Keys" with a table of known userInfos (only `enumType` today), more than having the section dedicated on "Handling Enums"?

```md
## UserInfo Keys

This template also make use of UserInfo keys that you can set on your dataModel's entities or attributes:

| Scope | UserInfo Key | Description |
|-------|--------------|-------------|
| Attribute | `enumType` | The name of the `enum` to associate to this attribute's type. The `enum` in question is expected to be declared in your code (SwiftGen won't generate it), must be `RawRepresentable` (e.g. `enum Foo: Int` or `enum Foo: String`) and its `RawValue` must match the attribute's type in your xcdatamodel. |
```

AliSoftware:
Another way to present that instead of a table, if you feel you'll need more space to document the feature (maybe a screenshot of the UserInfo's table with that `enumType` set on an attribute + some code extract to explain that more in details?!) is to have sub-sections

```md
## UserInfo Keys

This template also make use of UserInfo keys that you can set on your dataModel's entities or attributes

### Handling Enums

Blahblah explanation capture yada yada
```

And we'd add more `###` sub-sections if we end up adding more UserInfo Keys for other usages in the future.

AliSoftware (approved):

PR #596, opened 2019-02-08
Document dependencies #trivial

djbe:
Saw this blog post and thought it'd be a good idea to document the project's dependencies: http://blog.mazur.me/DEPENDENCIES.md


SwiftGen-Eve:

---

</p>

DEPENDENCIES.md
```diff
@@ -0,0 +1,39 @@
...
+- License: [MIT](https://github.com/SwiftGen/StencilSwiftKit/blob/master/LICENCE)
+
+### SwiftLint
+
+- Added to enforce a code style throughout the project, and avoid some bad coding decisions.
+- License: [MIT](https://github.com/realm/SwiftLint/blob/master/LICENSE)
+
+### Yams
+
+- Added because there's no built-in YAML parser, used for configuration file parsing.
```
AliSoftware: used for parsing the configuration file and for the YAML parser

AliSoftware (approved):

AliSoftware:
mmmh wait, I might have approved too quickly 😅 Shouldn't we use the occasion to add a Danger rule that would check if we added (or removed) a Dependency in our Podfile and/or Package.swift, and suggest to update that file in those occasions? Otherwise I fear that we might easily forget to keep that DEPENDENCIES.md up-to-date.

djbe:
You wanna tackle that one? Your ruby-fu is stronger than mine 😉 

Note: options safety PR coming soon.

AliSoftware:
Sure. Going to bed rn though and tomorrow will probably be quite busy, will try and take that on Sunday 👍

AliSoftware (changes requested):
- [x] Danger rule

Dangerfile
```diff
@@ -54,6 +55,13 @@ if podfile_changed ^ package_changed
   need_fixes << warn("You should make sure that `Podfile.lock` and `Package.resolved` are changed in sync")
 end
 
+# Check if DEPENDENCIES needs changes
+podfile_dependencies = YAML::load_file('Podfile.lock')['SPEC CHECKSUMS'].keys.sort
```
AliSoftware: I think `SPEC CHECKSUMS` might contain checksums of _all_ dependencies, including transitive ones.

In our current state, we don't have many transitive dependencies, and the ones we have (mainly `PathKit` coming from `Stencil` and `Stencil` coming from `StencilSwiftKit`), but imagine if `Kanna` had a dependency to `libXML` and `YAMS` had a dependency to hypothetical `Scanner`, `libYAML` and `Colorize` pods… we shouldn't require to document those, as they'd not be dependencies we chose to add directly, but transitive one coming as dependencies from other pods.

What we might want to document though:

* Dependencies explicitly added in our `Podfile` for SwiftGen-the-tool: we could parse those from the `PODS` key of the `Podfile.lock`, trimming away the version in parentheses to only keep the pod name (another idea would be to run `bundle exec pod ipc podfile Podfile` and parse all the values under any `dependencies` keys in that tree, recursively. But not sure that's worth the effort)
* Dependencies listed in `SwiftGenKit.podspec` as SwiftGenKit is also hosted in this repo and is the core of SwiftGen. We can parse those using `bundle exec pod pic spec SwiftGenKit.podspec` and extract the `dependencies` from the resulting JSON output.

AliSoftware (changes requested):

Dangerfile
AliSoftware: Also given that the base of those checks are not dependent on GIT and changes made in a PR, but could be run at any time as they compare existing file contents, we might want to move that portion of the code (listing missing dependencies in `DEPENDENCIES.md`) in a `rake` task instead, and in the `Dangerfile` just call that task — but only if the `Podfile.lock` has changed.

djbe:
👍 

Can't approve 'cause it's my own PR 🤦‍♂️

AliSoftware (approved):

PR #597, opened 2019-02-09
Parser Options - Part 2: Safety

djbe:
As discussed in #588 and in Slack, this PR adds better safety checks for parser options.
Initial implementation by @AliSoftware, added some code for checking if options match those we know.

Still need to:
- [x] Add some tests
- [ ] Maybe some way of displaying available options for config users?

### DIFF
--- a/Sources/SwiftGenKit/Parsers/Colors/ColorsFileTypeParser.swift
+++ b/Sources/SwiftGenKit/Parsers/Colors/ColorsFileTypeParser.swift
@@ -16,9 +16,11 @@ extension Colors {
 }
 
 protocol ColorsFileTypeParser: AnyObject {
+  init(options: [String: Any])
+
+  static var allOptions: ParserOptionList { get }
   static var extensions: [String] { get }
 
-  init()
   func parseFile(at path: Path) throws -> Colors.Palette
 }
 
--- a/Sources/SwiftGenKit/Parsers/Colors/ColorsFileTypeParser.swift
+++ b/Sources/SwiftGenKit/Parsers/Colors/ColorsFileTypeParser.swift
@@ -24,6 +24,12 @@ protocol ColorsFileTypeParser: AnyObject {
   func parseFile(at path: Path) throws -> Colors.Palette
 }
 
+extension ColorsFileTypeParser {
+  static var allOptions: ParserOptionList {
+    return []
+  }
+}
+
 // MARK: - Private Helpers
 
 extension Colors {
--- a/Sources/SwiftGenKit/Parsers/Colors/ColorsFileTypeParser.swift
+++ b/Sources/SwiftGenKit/Parsers/Colors/ColorsFileTypeParser.swift
@@ -16,7 +16,7 @@ extension Colors {
 }
 
 protocol ColorsFileTypeParser: AnyObject {
-  init(options: [String: Any])
+  init(options: ParserOptionValues)
 
   static var allOptions: ParserOptionList { get }
   static var extensions: [String] { get }
--- a/Sources/SwiftGenKit/Parsers/Parser.swift
+++ b/Sources/SwiftGenKit/Parsers/Parser.swift
@@ -40,17 +40,6 @@ public protocol Parser {
 }
 
 public extension Parser {
-  /// Factory method for safely creating a parser instance. Options will be checked against the list of
-  /// supported options (in `allOptions`).
-  ///
-  /// - Parameter options: Dictionary of options, checked against `allOptions`.
-  /// - Parameter warningHandler: Callback for logging issues.
-  /// - Returns: An instance of the parser.
-  static func createWith(options: [String: Any], warningHandler: MessageHandler?) throws -> Self {
-    try allOptions.check(options: options)
-    return try Self.init(options: options, warningHandler: warningHandler)
-  }
-
   /// Recursively search through the given paths, parsing any file or folder that matches the given filter.
   ///
   /// - Parameter paths: The paths to search recursively through.
--- a/Sources/SwiftGenKit/Parsers/Parser.swift
+++ b/Sources/SwiftGenKit/Parsers/Parser.swift
@@ -8,6 +8,11 @@ import Foundation
 import PathKit
 
 public protocol Parser {
+  /// Parser initializer. Please use the `createWith(options:warningHandler:)` factory method instead,
+  /// it'll ensure the safety of the provided options.
+  ///
+  /// - Parameter options: Dictionary of options, used to customize processing.
+  /// - Parameter warningHandler: Callback for logging issues.
   init(options: [String: Any], warningHandler: MessageHandler?) throws
 
   /// Regex for the default filter
@@ -35,12 +40,31 @@ public protocol Parser {
 }
 
 public extension Parser {
+  /// Factory method for safely creating a parser instance. Options will be checked against the list of
+  /// supported options (in `allOptions`).
+  ///
+  /// - Parameter options: Dictionary of options, checked against `allOptions`.
+  /// - Parameter warningHandler: Callback for logging issues.
+  /// - Returns: An instance of the parser.
+  static func createWith(options: [String: Any], warningHandler: MessageHandler?) throws -> Self {
+    try allOptions.check(options: options)
+    return try Self.init(options: options, warningHandler: warningHandler)
+  }
+
+  /// Recursively search through the given paths, parsing any file or folder that matches the given filter.
+  ///
+  /// - Parameter paths: The paths to search recursively through.
+  /// - Parameter filter: The filter to apply to each path.
   func searchAndParse(paths: [Path], filter: Filter) throws {
     for path in paths {
       try searchAndParse(path: path, filter: filter)
     }
   }
 
+  /// Recursively search through the given path, parsing any file or folder that matches the given filter.
+  ///
+  /// - Parameter path: The path to search recursively through.
+  /// - Parameter filter: The filter to apply to each path.
   func searchAndParse(path: Path, filter: Filter) throws {
     if path.matches(filter: filter) {
       let parentDir = path.absolute().parent()
--- a/Sources/SwiftGenKit/Parsers/Parser.swift
+++ b/Sources/SwiftGenKit/Parsers/Parser.swift
@@ -79,3 +79,9 @@ public extension Parser {
     }
   }
 }
+
+public extension Parser {
+  static var allOptions: ParserOptionList {
+    return []
+  }
+}
--- a/Sources/SwiftGenKit/Parsers/Parser.swift
+++ b/Sources/SwiftGenKit/Parsers/Parser.swift
@@ -10,11 +10,21 @@ import PathKit
 public protocol Parser {
   init(options: [String: Any], warningHandler: MessageHandler?) throws
 
-  // regex for the default filter
+  /// Regex for the default filter
   static var defaultFilter: String { get }
 
-  // Parsing and context generation
+  /// List of supported options
+  static var allOptions: ParserOptionList { get }
+
+  /// Try to parse the file (or directory) at the given path, relative to some parent path.
+  ///
+  /// - Parameter path: The file/directory to parse
+  /// - Parameter parent: The parent path (may not be the direct parent)
   func parse(path: Path, relativeTo parent: Path) throws
+
+  /// Generate the stencil context from the parsed content.
+  ///
+  /// - Returns: A dictionary representing the parsed data 
   func stencilContext() -> [String: Any]
 
   /// This callback will be called when a Parser want to emit a diagnostics message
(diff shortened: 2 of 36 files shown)


SwiftGen-Eve:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

---

</p>

djbe:
A thing I'm not too happy with is that the check for dictionary <-> available options is done outside the `init`. The burden is placed outside SGK, it must be done in all places where the Parser `init` would be called.

I initially did this because I wanted to avoid having to perform the check in each `init`, for each parser. Maybe a factory function's better in this case.

Sources/SwiftGenKit/Parsers/Parser.swift
```diff
@@ -25,12 +40,31 @@ public protocol Parser {
...
 
 public extension Parser {
+  /// Factory method for safely creating a parser instance. Options will be checked against the list of
+  /// supported options (in `allOptions`).
+  ///
+  /// - Parameter options: Dictionary of options, checked against `allOptions`.
+  /// - Parameter warningHandler: Callback for logging issues.
+  /// - Returns: An instance of the parser.
+  static func createWith(options: [String: Any], warningHandler: MessageHandler?) throws -> Self {
+    try allOptions.check(options: options)
```
AliSoftware: Not sure I like the factory method idea. I mean, what's the harm in doing that call in `init` and let consumers call `init` (instead of forgetting there's a factory method `createWith` they'd have to call instead of `init`)?

I don't think checking the options in `init` will have any significant performance effect or whatever, so not sure why you were reluctant to do it there, did I miss something?

Sources/SwiftGenKit/Parsers/ParserOptionList.swift
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  /// Check if the given dictionary of key values are all supported options.
+  ///
+  /// - Parameter options: a dictionary of options
+  func check(options: [String: Any]) throws {
+    let known = Set(self.options.map { $0.key })
+    for option in options where !known.contains(option.key) {
+      throw Error.unknownOption(key: option.key, value: option.value)
+    }
```
AliSoftware, line 32: The method will exit at the first `throw`, so looping with that `for` is useless here imho.

You might replace it with:
```swift
    let knownKeys = Set(self.options.map { $0.key })
    if let badOption = options.first(where: { !knownKeys.contains($0.key) }) {
      throw Error.unknownOption(key: badOption.key, value: badOption.value)
    }
```

Sources/SwiftGenKit/Parsers/ParserOptionList.swift
```diff
@@ -0,0 +1,44 @@
...
+public struct ParserOptionList {
+  enum Error: Swift.Error {
+    case unknownOption(key: String, value: Any)
+  }
+
+  private let options: [AnyParserOption]
+
+  init(options: [AnyParserOption] = []) {
+    self.options = options
+  }
```
AliSoftware, line 20: Bonus points: make `ParserOptionList: ExpressibleByArrayLiteral`.

That way we'd then be able to write:

```swift
static let allOptions: ParserOptionList = [
  Option.separator
]
```
And if we end up having more and more options, this would make it align and read nicely 😜

Sources/SwiftGenKit/Parsers/Parser.swift
```diff
@@ -8,13 +8,28 @@ import Foundation
...
+  /// - Parameter warningHandler: Callback for logging issues.
   init(options: [String: Any], warningHandler: MessageHandler?) throws
 
-  // regex for the default filter
+  /// Regex for the default filter
   static var defaultFilter: String { get }
 
-  // Parsing and context generation
+  /// List of supported options
+  static var allOptions: ParserOptionList { get }
```
AliSoftware, line 22: Should we provide a default value of `[]` (no option) for this one?

Sources/SwiftGenKit/Parsers/Strings/StringsParser.swift
```diff
@@ -29,20 +29,25 @@ public enum Strings {
   }
 
   private enum Option {
```
AliSoftware: Wondering if we shouldn't keep that `public`, so that other apps using SwiftGenKit from code could do:
```swift
let parser = Strings.Parser(options: [Strings.Parser.Option.separator.key: "__"])
```
to allow them to avoid hard-coded strings in their code… (well, if we do that, we might want to check if some other properties on `AnyParserOption` & co would also need to be public too)

AliSoftware (changes requested):

Sources/SwiftGenKit/Parsers/Parser.swift
djbe: The thing is we'd need to repeat this (and not forget it) in each and every parser init.

Sources/SwiftGenKit/Parsers/Parser.swift
djbe: Dunno. We don't for filter (although that'd be nonsensical, and without the default, it's clear in each parser what it can & can't do.

djbe: Although less code is always nice.

Sources/SwiftGenKit/Parsers/Strings/StringsParser.swift
```diff
@@ -29,20 +29,25 @@ public enum Strings {
...
   public final class Parser: SwiftGenKit.Parser {
     var tables = [String: [Entry]]()
     public var warningHandler: Parser.MessageHandler?
     private let keyStructureSeparator: String
 
-    public init(options: [String: Any] = [:], warningHandler: Parser.MessageHandler? = nil) {
+    public init(options: [String: Any] = [:], warningHandler: Parser.MessageHandler? = nil) throws {
       self.warningHandler = warningHandler
-      self.keyStructureSeparator = (options[Option.separator] as? String) ?? Entry.defaultSeparator
+      self.keyStructureSeparator = try Option.separator.get(from: options)
```
AliSoftware: Re: https://github.com/SwiftGen/SwiftGen/pull/597#discussion_r255347539

What about replacing `keyStructureSeparator: String` by `options: ParserOptionsList`, then in the init do `self.options = self.allOptions.extract(options)` — `extract(_ dict: [String: Any])` being a method we'd add on `ParserOptionList`, that would both validate that `dict` doesn't contain any key not in `self` (allOptions here), and then only extract the subset of keys found in that dict.

Then in places where we need to access that separator in the `Strings.Parser` we would do `self.options[.separator]` and get the string back

Multiple advantages of this approach:
* It allows to only have one property to store all options (if one day we don't have only "separator" but 5 more options, at least here we'll not need to add 5 more properties to `Strings.Parser` and parse them all one by one in `init`)
* It will force us to do the check in the `init` — no risk to forget to check if all options are valid, like you did with `createWith(…)` calling `check`, as if our parser has at least one option, we'll be forced to have a `let options` property in that Parser to hold them, and forced to initialise that non-optional value during `init` via `self.options = self.allOptions.extract(options)`. No risk to forget then
* And in the end this would avoid having to have that `createWith` factory method that we'll forget to call as likely as we'd have forgotten to call `check` in our `init` implementations

Sources/SwiftGenKit/Parsers/Parser.swift
AliSoftware: Suggestion: https://github.com/SwiftGen/SwiftGen/pull/597#discussion_r255347946

Sources/SwiftGenKit/Parsers/Colors/ColorsFileTypeParser.swift
```diff
@@ -24,6 +24,12 @@ protocol ColorsFileTypeParser: AnyObject {
   func parseFile(at path: Path) throws -> Colors.Palette
 }
 
+extension ColorsFileTypeParser {
+  static var allOptions: ParserOptionList {
+    return []
+  }
```
AliSoftware, line 30: I think the default value can be expressed as `static let allOptions: ParserOptionList = []` while still conforming to the protocol? (instead of a computed var)

Sources/SwiftGenKit/Parsers/Parser.swift
```diff
@@ -79,3 +79,9 @@ public extension Parser {
     }
   }
 }
+
+public extension Parser {
+  static var allOptions: ParserOptionList {
+    return []
+  }
```
AliSoftware, line 75: Same, `static let allOptions: ParserOptionList = []` if possible

Sources/SwiftGenKit/Parsers/Colors/ColorsFileTypeParser.swift
djbe: It doesn't.

> Static stored properties not supported in protocol extensions

Sources/SwiftGenKit/Parsers/Parser.swift
djbe: See above, it doesn't.

AliSoftware: 😢

Sources/SwiftGenKit/Parsers/Strings/StringsParser.swift
djbe: Thing is:
- Only 1 command has options for now (not to bad, we can add empty options to the other ones). Problem isn't now, but some future point where we have to remember that the init needs to parse/check the options.
- ParserOptionsList doesn't store values, it's a list of available options. We'd need an extra internal structure to store these.

AliSoftware: Ok, maybe `ParserOptionsList` wasn't the right type, but other than that, my whole point is if we implement such a think like in my comment, there's no way we'd risk avoiding checking options, because that check would have to be done if we want to extract those options from the dict in the first place


My point is, I really don't like the need of having that factory method, and having to remember to call it.

Tbh, I think there's way less risk that we forget to `check` the options during `init` — that could only happen if we implement a new parser one day, as I'd suggest we'd already add the call to `check` in the `init` of the existing parsers in that PR even the ones for which `allOptions = []`, after all we have to, to ensure they're not called with any option as they don't support any) — than the risk of not knowing / forgetting to call the `createWith` factory method (both for us in the various parts of SwiftGen & SwiftGenKit & SwiftGenKitTests where we have to create those parsers, than for any other app or framework that would want to use SwiftGenKit from their Swift code).

How many times do we create a new parser, for which there's that risk of forgetting to call `check`? And when we create new parsers we'll likely, if not copy/paste, at least get inspiration from existing Parsers. While for instantiating parsers, nobody would think that we shouldn't call `init` and will likely not read the doc-comment about it anyway, as long as autocomplete suggests a public init they'll call it without any 2nd thoughts and the risk of not using `createWith` like they'd be supposed to do is way more likely imho.

djbe:
I've made the requested changes, but purposefully left in a compile error where a new type of issue arises. The problem is that:
- The subscript needs to return a `T` instead of a `T?` --> needs access to the default value
- Add default value to `AnyParserOption` --> need to add associated type `Value` to protocol
- PAT --> we can no longer have a `ParserOptionList` with an array of `AnyParserOption`
- ...
- No idea

djbe:
Right, so that solves the compilation issue. Now the annoying thing is that we don't have a compile time check if the option is available, only a runtime `fatalError`. Ideas to improve this are always welcome 😄 

Now, what would be the best way to handle config `lint` / listing available options? For CLI we show available options in the command description. With config, we have no such option. @AliSoftware thoughts?

Tests/SwiftGenKitTests/AssetCatalogTests.swift
```diff
@@ -5,47 +5,60 @@
...
   }
+
+  // MARK: - Custom options
+
+  func testUnknownOption() throws {
+    do {
+      _ = try AssetsCatalog.Parser(options: ["SomeOptionThatDoesntExist": "foo"])
+      XCTFail("Parser successfully created with an invalid option")
+    } catch ParserOptionList.Error.unknownOption {
+      // That's the expected exception we want to happen
```
AliSoftware, line 59: Maybe check that the associated values of the thrown error are what's expected?

AliSoftware (approved):
After that addition on the test for bad options, should be ok to merge.

Maybe in the future we might find an even better solution/API for that, like making `ParserOptionValues` generic over an associated type of the Parser protocol (that associated type being the `enum Option` listing all available options), allowing the subscript to be bound to that type and access to an option to be shortcut to `options[.separator]` instead of `options[Option.sepatator]`… but I think at that point we'll have time to experiment with that more in the future otherwise we'll likely never find a satisfying solution to begin with

PR #600, opened 2019-02-10
Swift 5 templates

### LINKED ISSUE #595: Swift 5 templates
Xcode 10.2 beta with Swift 5 is out. We should prepare new templates for Swift 5 with the needed changes (should be minimal).

We won't be able to test it yet until CircleCI adds Xcode 10.2 support (probably soon?).

Related to that change, we should consider deprecating the Swift 3 templates (like we did with 2, keep them around for a while, but unable to test them and removing them with the next major version)

### LINKED ISSUE #631: FontConvertible 'Expression implicitly coerced from 'UIFont?' to Any
Since updating my project to Swift 5 and Xcode 10.2 I am finding these warnings everywhere.

> Expression implicitly coerced from `UIFont?` to `Any`
> Provide a default value to avoid this warning
> Force-unwrap the value to avoid this warning

Seems excessive for using a font in my assets.  
Here is one piece of code showing the problem.

let attributedString = NSMutableAttributedString(string: "Login", attributes: [
            .font: UIFont(font: FontFamily.OurFont.regular, size: 28), <- warning here
            .paragraphStyle: paragraphStyle
            ])

Using SwiftGen 6.1.0

djbe:
Fixes #595.
Fixes #631.

Copied the Swift 4 templates (and made some small adjustements). ~This PR is still a WIP, as SwiftGen (and dependencies) won't compile yet in Xcode 10.2.~ Everything compiles now.

Todo:
- [x] Apply the changes in #573 to Swift 5 templates
- [x] Update CircleCI to 10.2 (thread: https://discuss.circleci.com/t/when-is-xcode-10-2-support-coming-up/28160/5)
- [x] Update Kanna for 10.2 (with https://github.com/tid-kijyun/Kanna/pull/207, ~unreleased~)
- [x] Update SwiftLint to fix ACL warnings (with https://github.com/realm/SwiftLint/pull/2609, ~unreleased~)


SwiftGen-Eve:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

---

</p>

Tests/Fixtures/Generated/Colors/literals-swift4/defaults-publicAccess.swift
```diff
@@ -17,12 +17,12 @@
 // swiftlint:disable identifier_name line_length type_body_length
 public extension ColorName {
   /// 0x339666ff (r: 51, g: 150, b: 102, a: 255)
-  public static let articleBody = #colorLiteral(red: 0.2, green: 0.5882353, blue: 0.4, alpha: 1.0)
+  static let articleBody = #colorLiteral(red: 0.2, green: 0.5882353, blue: 0.4, alpha: 1.0)
```
AliSoftware, line 20: how come this `public` suddenly disappeared here? (and same for others, `public` & `internal` seems to be removed everywhere in that PR)

djbe: See the separate commit for this. The swift compiler in 10.2 warns about unnecessary internal/public like these.

AliSoftware: But won't this break things depending if they are compiled in a framework and used outside in a real project (vs when they are compiled in the same module during our compilation tests)?

djbe: Just tried it in a playground and it works.

If you put files in the extra "Sources" folder, they're compiled as a separate module, so you need to make it public. To be extra sure, once I make the extension `internal`, it stops working.

djbe: Ah, no, there's a case where this doesn't work:
- In a `public` extension, the `static`s work fine.
- In an `enum` inside a `public` extension, the `static` aren't visible, and each needs to be marked as public.

AliSoftware: Is this a behavior change specific to Swift 5, where ACLs declared at type level are now inherited that way on statics? Or was it always the case and those extra ACLs on statics were useless but what's new is they now warn about them?
(Also is this really a Swift 5 compiler warning not a SwiftLint warning?)

djbe: This was for the Swift 4 templates, so Swift 5 compiler in Swift 4 compatibility mode. I saw the warning while running `rake output:compile`, which doesn't run any lint tasks.

AliSoftware: I'm just hoping then that the behavior with "Swift 5 compiler in Swift 4 mode" is the same as the behavior of the Swift 4 compiler. By design they are supposed to, but just want to be sure we don't accidentally remove visibility of our constants in Frameworks for people still using it Xcode/compiler

djbe: Note that I tested the public/internal behaviour in playgrounds using Xcode 10.1, so Swift 4 compiler.

djbe: Quick note, these are the warnings I'm getting with S5 in S4 mode:

> Tests/Fixtures/Generated/Colors/literals-swift4/multiple.swift:19:3: warning: 'internal' modifier is redundant for enum declared in an internal extension
>   internal enum Colors {

> Tests/Fixtures/Generated/Colors/literals-swift4/defaults-publicAccess.swift:20:3: warning: 'public' modifier is redundant for static property declared in a public extension
>   public static let articleBody = #colorLiteral(red: 0.2, green: 0.5882353, blue: 0.4, alpha: 1.0)

> Tests/Fixtures/Generated/Colors/literals-swift4/defaults.swift:20:3: warning: 'internal' modifier is redundant for static property declared in an internal extension
>   internal static let articleBody = #colorLiteral(red: 0.2, green: 0.5882353, blue: 0.4, alpha: 1.0)

> Tests/Fixtures/Generated/Colors/literals-swift4/defaults-customName.swift:19:3: warning: 'internal' modifier is redundant for static property declared in an internal > extension
  internal static let articleBody = #colorLiteral(red: 0.2, green: 0.5882353, blue: 0.4, alpha: 1.0)

djbe: I've pushed some changes to ensure the template doesn't generate ACL specifiers inside the extension, except for deeper levels (where it's needed again). This should fix both the Xcode 10.2 warnings as well as keeping `public` working.

AliSoftware:
TODO: ~apply the changes in #573 to Swift 5 templates too~ done!

djbe:
Rebased on #617, including the availability changes.

mhausherr:
Swift 5 add warnings when using font(size: CGFloat) in a [NSAttributedString.Key: Any] array.
`Provide a default value to avoid this warning`
`Force-unwrap the value to avoid this warning`
`Explicitly cast to 'Any' with 'as Any' to silence this warning`

can be solve by
```
   {{accessModifier}} func font(size: CGFloat) -> Font {
     return Font(font: self, size: size)!
   }
```
in templates/fonts/swift5.stencil line 50

djbe:
I think a `guard` might be better? It'd match what we do in other templates, and provides a more descriptive error message?

```swift
  func font(size: CGFloat) -> Font {
    guard let font = Font(font: self, size: size) else {
      fatalError("some error message")
    }
    return font
  }
```

AliSoftware:
Indeed I prefer the `fatalError(explicitMessage)` as well. We should also apply the same change to other templates probably (I think for many parsers we return IUO so far instead of non-optionals being unwrapped in the func)

Tests/Fixtures/Generated/Fonts/swift5/defaults-customName.swift
```diff
@@ -0,0 +1,101 @@
...
+
+// MARK: - Implementation Details
+
+internal struct FontConvertible {
+  internal let name: String
+  internal let family: String
+  internal let path: String
+
+  internal func font(size: CGFloat) -> Font! {
+    return Font(font: self, size: size)
```
davideme: Can the unwrapping be done inside the function instead of a return parameter, as it's now considered the same as Font? in Swift5?

AlexisQapa:
Does this remove unused access modifier ? Using 6.1.0 and literals-swift4 add redondant access modifiers which Xcode 10.2 warns about.

Update: here is the warning : `'internal' modifier is redundant for static property declared in an internal extension`, it's extensions only. Should I create a separate issue ?

djbe:
Yeah, it's the first commit in this PR, to fix some warnings in Swift 4.

SwiftGen.playground/contents.xcplayground
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<playground version='6.0' target-platform='ios' display-mode='rendered'>
+<playground version='6.0' target-platform='ios' display-mode='rendered' last-migration='1020'>
```
AliSoftware, line 2: Should we also upgrade the sample codes in the Playground to Swift 5?
(🤔 💭  they might not need any change in the user-facing pages after all... but maybe in the content hidden in `Sources/` folders needs some love? Anyway, worth checking if you didn't already)

AliSoftware (approved):

SwiftGen.playground/contents.xcplayground
djbe: Yeah already did that (some minor changes) 👍

Issue #603, opened 2019-02-11
Inconvenient access to Assets of type color & image


Jeehut:
Currently, when I use the default `swift4` template for `xcassets` I need to access my colors like so:

```Swift
view.backgroundColor = Asset.Colors.Theme.accent.color
```

When accessing images it looks like so:

```Swift
imageView.image = Asset.Images.SomeList.add.image
```

In my opinion this is unnecessarily long. It should rather be like these calls:

```Swift
view.backgroundColor = Colors.Theme.accent
imageView.image = Images.SomeList.add
```

Note that I'm aware that we can use our own custom templates (actually we're doing this, see [here](https://github.com/JamitLabs/NewProjectTemplate-iOS/blob/9ac583955ad6dc321a385357bc6c4a63d038f62f/SwiftGen-Colors.stencil) and [there](https://github.com/JamitLabs/NewProjectTemplate-iOS/blob/9ac583955ad6dc321a385357bc6c4a63d038f62f/SwiftGen-Images.stencil)). But I think the default template could be improved in this perspective. 

I mean, the Strings call also doesn't look like `Asset.Strings.Onboarding.Header.title.string`. It's `L10n.Onboarding.Header.title` which is to the point. The context `L10n` makes clear the last component is a `String` and I'm not able to get the `key` name either so why do I need it for images and colors, but not for Strings? Also why isn't there at least a typealias for the start like:

```Swift
typealias Colors = Asset.Colors
typealias Images = Asset.Images
```

I don't seem to understand these points. I'd be glad to hear the rationale.

djbe:
The very major difference is that assets can be quite memory heavy. In your example templates, **all** images are stored in memory, which is absolutely not the way to go. If someone has a few MBs of image or data assets, they'd all be **at all times** in memory.

Something you can do in your codebase is to add extensions to for example `UIImageView`, to have a function like:
```swift
extension UIImageView {
  func set(asset: ImageAsset) {
    image = asset.image
  }
}
```

djbe:
Note, there is a typealias at the start:
```swift
typealias AssetColorTypeAlias = UIColor
typealias AssetImageTypeAlias = UIImage
```

The reason they have a somewhat long name, is to avoid conflicts with the other templates' output (or other modules). You can always customise these using the template parameters, see the documentation for more information (https://github.com/SwiftGen/SwiftGen/blob/master/Documentation/templates/xcassets/swift4.md#customization).

AliSoftware:
[...]

PR #605, opened 2019-02-13
XCAssets: Add new simplified template `simple-swift4`

Jeehut:
This new template removes the need for the last call part like `.color` or `.image`. Also it adds a typealias for each catalog if multiple are used to prevent calls like `Asset.Colors` or `Asset.Images`. See also https://github.com/SwiftGen/SwiftGen/issues/603 for more information on the rationale.


Jeehut:
Unfortunately the tests fail for me locally, probably also will on the CI. Don't know what I missed, I've updated the fixtures manually (generating didn't work even when the schemes were selected). What did I miss?

PR #606, opened 2019-02-14
Add a help message if there's an error #trivial

djbe:
I don't know how far we want to go with this. CocoaPods seems to display a quite nice error message with suggestions (when there's an error), we may want to take some inspiration from there.

### DIFF
--- a/Sources/SwiftGen/Commander/ConfigCommands.swift
+++ b/Sources/SwiftGen/Commander/ConfigCommands.swift
@@ -9,6 +9,26 @@ import PathKit
 import StencilSwiftKit
 import SwiftGenKit
 
+extension Config {
+  fileprivate func run(cmd: String, verbose: Bool) throws {
+    guard let parserCmd = ParserCLI.command(named: cmd) else {
+      throw Config.Error.missingEntry(key: cmd)
+    }
+
+    for var entry in commands[cmd] ?? [] {
+      entry.makingRelativeTo(inputDir: inputDir, outputDir: outputDir)
+      if verbose {
+        for item in entry.commandLine(forCommand: cmd) {
+          logMessage(.info, " $ \(item)")
+        }
+      }
+
+      try entry.checkPaths()
+      try entry.run(parserCommand: parserCmd)
+    }
+  }
+}
+
 extension ConfigEntryOutput {
   func checkPaths() throws {
     guard self.output.parent().exists else {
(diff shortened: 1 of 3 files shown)


djbe:
Depending on how far we go with this, I'll add a proper changelog entry 😄

AliSoftware:
I think we have to find a middle ground and not go too far though: tools printing very long console output in error, even if the intentions are good like give so much advice as possible, generally end up being unread (either because the user sees it as a crashlog or verbose output, or because it's too long to read (/gif ain't nobody got time for that))

So I feel like just pointing to `config lint` and the repo like you did is probably enough and just the right amount 👌

Sources/SwiftGen/Commander/ConfigCommands.swift
```diff
@@ -110,6 +110,13 @@ enum ConfigCLI {
       }
     } catch let error as Config.Error {
       logMessage(.error, error)
+      logMessage(.info, """
```
AliSoftware: Not sure why `.error` was changed to `.info` here… tbh I don't remember if this only impacts the way it's presented (red vs yellow) or if it also impacts the exit code (w should still exit with non-zero in those cases)?

djbe: Ah I thought since it was an info message, it should be `.info`, but good point, I'll change it to `.error`.

Sources/SwiftGen/Commander/ConfigCommands.swift
```diff
@@ -9,6 +9,26 @@ import PathKit
 import StencilSwiftKit
 import SwiftGenKit
 
+extension Config {
+  fileprivate func run(cmd: String, verbose: Bool) throws {
+    guard let parserCmd = ParserCLI.command(named: cmd) else {
+      throw Config.Error.missingEntry(key: cmd)
+    }
```
AliSoftware, line 16: Good point extracting that `guard` from the `for var entry` loop 👍

SwiftGen-Eve:

---

</p>

Tests/SwiftGenTests/ConfigLintTests.swift
```diff
@@ -83,15 +84,17 @@ class ConfigLintTests: XCTestCase {
...
           \(cmd).outputs.templatePath: /templates/\(cmd)/templates.stencil is an absolute path. \
-          Prefer relative paths for portability when sharing your project.
+          Prefer relative paths for portability when sharing your project (unless you are using \
+          environment variables).
           """),
         (.warning, """
           \(cmd).outputs.output: /output/\(cmd)/out.swift is an absolute path. Prefer relative \
-          paths for portability when sharing your project.
+          paths for portability when sharing your project (unless you are using environment \
+          variables).
```
AliSoftware: Wondering if it would be worth putting that expected error message (well the common part starting from "is an absolute part…") in an intermediate constant?! 🤔

AliSoftware (approved):

Tests/SwiftGenTests/ConfigLintTests.swift
djbe: bcf1eb8 what you had in mind?

AliSoftware: Wow didn't expect to go that far, was just thinking of a plain string `let`… but yeah seems great 👍

AliSoftware (approved):

PR #610, opened 2019-02-20
CoreData: expand RawRepresentable to non-enum types

### LINKED ISSUE #609: CoreData: expand RawRepresentable to non-enum types.
While messing around with phantom types, I noticed that `RawRepresentable` doesn't have to be an enum. It can be an `OptionSet`, but also our own type such as a `struct`. So for example, we could have (yay phantom types):

```swift
struct Identifier<T: NSManagedObject>: RawRepresentable {
  let rawValue: Int64

  init?(rawValue: Int64) {
    self.rawValue = rawValue
  }
}
```

And use it in some types:
```swift
final class Test: NSManagedObject {
  internal var id: Identifier<Test> {
    get {
      let key = "id"
      willAccessValue(forKey: key)

      guard let value = primitiveValue(forKey: key) as? Identifier<Test>.RawValue,
        let result = Identifier<Test>(rawValue: value) else {
        fatalError("Could not convert '\(value)' to type 'Identifier<Test>'")
      }
      return result
    }
    set {
      let key = "id"
      willChangeValue(forKey: key)

      setPrimitiveValue(newValue.rawValue, forKey: key)
    }
  }
}
```

So, this code quite closely matches what we already have for `enum` attributes (see #593) with the `enumType` user info key. What I propose is that we rename that user info key to something like `rawRepresentableType` (which can be set to `Identifier<Test>` for the example above).

An extra thing that we should consider adding is some user info key to trigger code generation to unwrap the enum optional as above, so that the resulting attribute is a `Identifier<Test>` instead of a `Identifier<Test>?`. Suggestions for a name are welcome 😄 For example: `unwrapOptional` set to `"true"`?

djbe:
Fixes #609.

This renames the (unreleased) `enumType` to the more generic name `rawType`. It also adds an additional user info key `unwrapOptional`, for avoiding optional `RawRepresentable` attributes (using a `guard` & `fatalError`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,10 +24,12 @@
   [David Jennes](https://github.com/djbe)
   [#591](https://github.com/SwiftGen/SwiftGen/issue/591)
   [#592](https://github.com/SwiftGen/SwiftGen/pull/592)
-* Core Data: the built-in templates now support `RawRepresentable` attributes (such as `enum`, `OptionSet`, …). They'll check the "User Info" of an attribute for a `RawType` key, which should be set to the type name you want to use for that attribute. For more information, check the [template's documentation](Documentation/templates/coredata/swift4.md#userinfo-keys).  
+* Core Data: the built-in templates now support `RawRepresentable` attributes (such as `enum`, `OptionSet`, …). They'll check the "User Info" of an attribute for a `RawType` key, which should be set to the type name you want to use for that attribute. To avoid optional attributes, you can also add the `unwrapOptional` user info key. For more information, check the [template's documentation](Documentation/templates/coredata/swift4.md#userinfo-keys).  
   [David Jennes](https://github.com/djbe)
   [#566](https://github.com/SwiftGen/SwiftGen/issue/566)
+  [#609](https://github.com/SwiftGen/SwiftGen/issue/609)
   [#593](https://github.com/SwiftGen/SwiftGen/pull/593)
+  [#610](https://github.com/SwiftGen/SwiftGen/pull/610)
 * Strings: the built-in templates now accept a parameter for customizing the localization function.  
   [Steven Magdy](https://github.com/StevenMagdy)
   [426](https://github.com/SwiftGen/SwiftGen/issues/426)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@
   [David Jennes](https://github.com/djbe)
   [#591](https://github.com/SwiftGen/SwiftGen/issue/591)
   [#592](https://github.com/SwiftGen/SwiftGen/pull/592)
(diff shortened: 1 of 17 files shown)


SwiftGen-Eve:

---

</p>

jrtibbetts:
Pretty cool. 👍

AliSoftware (commented):
I though if commented on that already (but that must have been on another CoreData-related PR I guess): I feel like `RawValue` would fit better as a userInfo key name rather than `rawType` (I think of that userInfo key similar to declaring a typealias of sorts). Having the userInfo key start with a lowercase while we expect its value to be a type feels strange, and `RawValue` matches the name of the associatedtype of `protocol RawRepresentable` so it's somehow a term of art or at least an official term already…

djbe:
Yeah you mentioned it in the issue (#609). The reason I didn't go with that is that `RawValue` is the raw value of the `RawRepresentable`, so for example `String` or `Int`. That's not what we want from this key, we want the name of the type that is `RawRepresentable`.

AliSoftware:
Ah right. `RawRepresentableType` or just `RawType` then? 😅

djbe:
Yeah `RawRepresentableType` was so long compared to `enumType`, that I just went with `rawType` 😆 

I can change it to uppercase `RawType`, just wanted some consistency with the other keys.

AliSoftware:
😉 Tbh I'm pretty sure that I'm not gonna be a user of that template myself so might not be the best one to comment.

But what I can say is that in gyro we chose a name for a userInfo that we later regretted (`IdentityAttribute` that we should have named `PrimaryKey` in retrospect) and could never change it without breaking existing models… so it's worth choosing one we're sure we'll like 😛

djbe:
Anyone mind doing a last review? (no matter if you're going to use this or not, just review the code)

AliSoftware (commented):
Random comment, not saying this is the right or wrong take, just a random thought about a possible alternative:

`unwrapOptional` only makes sense in attribute that also have `RawType`. So I'm wondering if using some convention instead like "if `RawType` ends with `!` then the optional will be unwrapped" would make more sense, linking them together that way?

djbe:
Dunno, nothing stops a user from marking a `Date` property as having a `RawType`. Who knows, maybe that even works in Swift? (or at some point in the future). Users can create "weird/wrong" configurations like that.

I'd rather not have the attribute itself be a `var stuff: MyRawType!`, as with the changes in swift 4, it automatically becomes a `MyRawType?` when passing it along in your code.

So we'd somehow have to be able to extract that last character (`hasSuffix` filter) and also remove it from the value in the template itself (we can't do that yet?).

AliSoftware:
Yeah fair point probably not worth the trouble

djbe:
Nothing stops us from revising this when we have the needed functionality in Stencil/SSK, and a breaking change with a major release.

AliSoftware (approved):

PR #612, opened 2019-02-23
Update some outdated code samples #trivial

### LINKED ISSUE #604: String flat-swift4.stencil template does not match with documentation
After reading the documentation about `flat-swift4.stencil` found here: https://github.com/SwiftGen/SwiftGen/blob/master/Documentation/templates/strings/flat-swift4.md

I've realised that the generated file is not matching what the documentation said.
In this section https://github.com/SwiftGen/SwiftGen/blob/master/Documentation/templates/strings/flat-swift4.md#generated-code the generated code is an enum with a case per localizable string key.

But after generating the strings file, the localizable strings are `static let` and `static func` like you could find here https://github.com/SwiftGen/SwiftGen/blob/master/Tests/Fixtures/Generated/Strings/flat-swift4-context-localizable.swift

djbe:
Fixes #604.


SwiftGen-Eve:

---

</p>

Documentation/templates/colors/literals-swift4.md
```diff
@@ -29,7 +29,7 @@ Note: if you use `enumName: UIColor` (or `NSColor` on macOS) then the color cons
 **Extract:**
 
 ```swift
-extension ColorName {
+internal extension ColorName {
   /// 0x339666ff (r: 51, g: 150, b: 102, a: 255)
   static let articleBody = #colorLiteral(red: 0.2, green: 0.588235, blue: 0.4, alpha: 1.0)
```
AliSoftware, line 34: 1) Why isn't this one `internal` like with other templates?

Documentation/templates/colors/literals-swift3.md
```diff
@@ -30,11 +30,11 @@ Note: if you use `enumName: UIColor` (or `NSColor` on macOS) then the color cons
 **Extract:**
 
 ```swift
-extension ColorName {
+internal extension ColorName {
   /// 0x339666ff (r: 51, g: 150, b: 102, a: 255)
-  static let articleBody = #colorLiteral(red: 0.2, green: 0.588235, blue: 0.4, alpha: 1.0)
+  internal static let articleBody = #colorLiteral(red: 0.2, green: 0.588235, blue: 0.4, alpha: 1.0)
```
AliSoftware, line 35: 2) Or is it this one being the bad one, like since it's an `internal extension` that `internal` is redundant here?

I don't remember which are the exact rules but since I remember you fixed it in a recent PR when testing with the Swift 5 compiler warning about redundant access modifiers… better at least make all templates, and this all examples in the doc, consistent with each other

Documentation/templates/colors/literals-swift4.md
djbe: Because I already incorporated the swift4 fixes from the swift5 PR

Documentation/templates/colors/literals-swift3.md
djbe: Uhm, this is swift 3, which will never have that warning about redundant ACL modifiers. I'm not even sure what's valid in swift 3, and we haven't changed the templates for it in the swift 5 PR. I think we better keep this one as is (with the extra `internal`).

Documentation/templates/colors/literals-swift4.md
AliSoftware: Mmmh ok but then only this one is different from the other, like `colors/swift3` (the non-literal one) doesn't have the same structure and balance of internal present it not (probably because the literals template declare them in an extension instead of directly in the type declaration?)

Then why do we have this inconsistency in swift4 vs literals-swift4 too (extension vs not)? Not very consistent of us 😅

djbe: Uhm the literals one is so that you can have `UIColor.myRedStuff` if you set the type name to UIColor.

The non-literal version uses struct, with its own properties. It can never be defined as an extension on an existing type such as `UIColor`

AliSoftware: Ah right. Might be worth make the non-literal one accept to be used as an UIColor extension at one point just got the sake of consistency, but well…

djbe: Like I said, we can't do that, as it's a struct with a `rgbaValue` and `color` stored properties.

It'd be weird if we somehow retooled the template to allow it, as you'd then use it as:
```swift
UIColor.articleBody.color
```

AliSoftware: Ah good point!

AliSoftware (approved):

PR #613, opened 2019-02-23
XCAssets: rename our testing catalogs to avoid confusion

djbe:
This addresses a minor point in #603, namely that some users might think that resources are grouped per type (images, data, colors), and then still have to access the actual value using `.image`, `.color`, ...

For now, I've renamed the catalogs to `Food` (for our test images), `Files` (for json/readme data) and `Styles` (for colors).

PR #614, opened 2019-02-24
XCAssets: add support for AR Resource Groups

djbe:
Adds support for [AR Resource Group](https://developer.apple.com/library/archive/documentation/Xcode/Reference/xcode_ref-Asset_Catalog_Format/AR_Resource_Group.html#//apple_ref/doc/uid/TP40015170-CH61-SW1)s, which contain reference images and objects (hence the 2 accessors).

I'm going to work my way down the list of asset types we can support. This just happens to be the first one alphabetically (app icons don't count, we can't do anything with them).

Note: ~based on #613, so merge that first.~
Note: I made sure the test resources are from the public domain, or made them myself (reference object).


SwiftGen-Eve:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

---

</p>

Documentation/templates/xcassets/swift4.md
```diff
@@ -56,6 +65,14 @@ enum Asset {
...
 
 ```swift
+// you can load an AR resource group's items using:
+let bottles = Asset.Targets.bottles.referenceObjects
+let paintings = Asset.Targets.paintings.referenceImages
+
+// Or as an alternative, you use the convenience constructor like this:
+let sameBottles = ARReferenceImage.referenceImages(in: Asset.Targets.bottles)
+let samePosters = ARReferenceObject.referenceObjects(in: Asset.Targets.paintings)
+
```
AliSoftware: I know you listed then in alphabetical order but ARResourceGroups are probably way less known by the average iOS developer (I never encountered then myself, never playing with AR yet) so it would be cool if the documentation didn't highlight them first before the more widely used colors and images

One option to do that would be to separate the examples for colors and images (the most commonly used ones) in one code snippet, then in a second paragraph titled something like "and the principles are the same for other type of resources you can store in an asset catalog:" put another snippet with all the rest?

Another solution could be too make a separate snippet for each resource type and adding a section title like `### Images` above each.
Or maybe even putting then in a collapsible section each?

Tests/Fixtures/Generated/XCAssets/swift3/all-allValues.swift
```diff
@@ -2,11 +2,16 @@
...
+  import AppKit
   internal typealias AssetColorTypeAlias = NSColor
   internal typealias AssetImageTypeAlias = NSImage
-#elseif os(iOS) || os(tvOS) || os(watchOS)
-  import UIKit.UIImage
+#elseif os(iOS)
+  import ARKit
+  import UIKit
+  internal typealias AssetColorTypeAlias = UIColor
+  internal typealias AssetImageTypeAlias = UIImage
```
AliSoftware: Not for this PR but just a though in passing: maybe instead of creating such ugly long names for our typealiases just so we're sure they don't collide with others names choosen by our users:

1) Maybe we could provide a param, defaulting to "iOS", to tell which platform the constants are targeting, and then directly use the right type in 95% of situations. And if people need to use the generated constants in multiple platforms at once only then do we generate a  conditional typealias? Ok this one might not be worth the effort though but at least if we do it we'd have the benefit of having proper well known types in autocomplete… instead of those typealiases
2) Or maybe a simpler solution, namespace those typealiases:

```
enum {{typealiasNamespace|default:"SwiftGenNativeType"}} {
  typealias Color = UIColor
  typealias Image = UIImage
}
```

This could be a breaking change as it would replace multiple `param` (one for each typealias name) with a single one (the name of the namespace for them) but could be nicer to read

3) A third option would be to directly put those `typealias` directly under the enum we create to hold all the constants we generate, or under the type we use to wrap the return values already (eg `ColorName` for colors, etc)

Having those typealiases not at the top level but nested in the namespaces we already create would still prevent them to collide with existing types in the user's codebase but make them more readable than with their long, non-showBuildSettings, root names 😉

Tests/Fixtures/Generated/XCAssets/swift3/all-customName.swift
```diff
@@ -2,11 +2,16 @@
...
 
 #if os(OSX)
-  import AppKit.NSImage
+  import AppKit
   internal typealias XCTColor = NSColor
   internal typealias XCTImage = NSImage
-#elseif os(iOS) || os(tvOS) || os(watchOS)
-  import UIKit.UIImage
+#elseif os(iOS)
+  import ARKit
```
AliSoftware, line 7: Didn't check the template to see if they was already the case, but we should only `import ARKit` if we have one ARResource at least. If we don't have any, importing ARKit means auto-linking to it (that's how clang module imports work!)… for nothing

templates/xcassets/swift3.stencil
```diff
@@ -88,6 +99,40 @@
...
+}
+
+@available(iOS 12.0, *)
+{{accessModifier}} extension ARReferenceObject {
+  static func referenceObjects(in asset: {{arResourceGroupType}}) -> Set<ARReferenceObject> {
+    let bundle = Bundle(for: BundleToken.self)
+    return referenceObjects(inGroupNamed: asset.name, bundle: bundle) ?? Set()
+  }
+}
+#endif
```
AliSoftware, line 146: Should only be present if at least one resource of this type exists in the asset catalog, otherwise it quite pollutes the implementation details if no generated constant use it, and it imports (and links to) ARKit for nothing

AliSoftware (changes requested):

Tests/Fixtures/Generated/XCAssets/swift3/all-allValues.swift
AliSoftware: An additional value we can get by choosing option 3 is that if we add some appropriate conditions in our templates to only generate our `struct ColorName`-like types wrapping resources for lazy evaluation… only if there's at least one resource of this type but not zero, the typealias we'd declare inside those would disappear alongside it and not be uselessly generated either.

Tests/Fixtures/Generated/XCAssets/swift3/all-customName.swift
djbe: That... might be a bit easier said than done. I'll look into it.

templates/xcassets/swift3.stencil
djbe: Ehm, if we do it for this, do we do it consistently for the other types? And what if we at some point want to separate the implementation details? Then we need all the implementations.

AliSoftware: Of course if we start that (which was less necessary before since the other types didn't require to import a less common Framework and we didn't have that many different resource types to begin with, but now that we're growing and spring more resource types…) we might want to apply that concept (only generating the implementation details if there's at least one resource of said type) to all resource types (images, colors, …) ; ARResourceGroups are just the most significant ones for which we'd want that to avoid importing a big framework for nothing, but better make it consistent for all resource types while we're at it 😅

Tests/Fixtures/Generated/XCAssets/swift3/all-allValues.swift
djbe: 1. Bad idea IMO. Especially for cross-platform projects, but also for example in playgrounds, where we can easily switch platforms at will.
2. Better, I'll see if we can make it backwards compatible.
3. We don't namespace assets per type though (see #613, where your confusion might come from?)

djbe: Oh, misread your #3. That's a possibility.

Tests/Fixtures/Generated/XCAssets/swift3/all-customName.swift
AliSoftware: Might be worth making the parser do the counting (like exposing the number of resources for each type in the context) so that it'd be easier to then test `{% if resourceCount.arresourcegroup > 0 %}` in the template — instead of having the template loop on the list of all resources and checking the type of each one etc all in the template

templates/xcassets/swift3.stencil
AliSoftware: I'm not sure I see how separating the implementation details in their own file like we discussed in another issue would be a problem: the templates generating the implementation details could use the context containing all the resource to know what to generate. And a param to force generating each type if people need customization (like if they analyze multiple folders as different inputs in their config file). We'll always find a solution by then when we'll start working on that issue I think, I'm not too worried about that

Sources/SwiftGenKit/Parsers/AssetsCatalog/CatalogEntry.swift
```diff
@@ -7,12 +7,25 @@
...
 extension AssetsCatalog {
   enum Entry {
-    case color(name: String, value: String)
-    case data(name: String, value: String)
-    case group(name: String, isNamespaced: Bool, items: [Entry])
-    case image(name: String, value: String)
+    struct EntryWithValue: AssetsCatalogEntry {
+      let name: String
+      let value: String
+      let type: String
```
AliSoftware: Why is this `String` and not `Item`?

Sources/SwiftGenKit/Parsers/AssetsCatalog/CatalogEntry.swift
```diff
@@ -58,39 +73,55 @@ extension AssetsCatalog.Entry {
...
 
+extension AssetsCatalog.Entry.EntryWithValue {
+  fileprivate init(path: Path, item: Constants.Item, withPrefix prefix: String) {
+    name = path.lastComponentWithoutExtension
+    value = "\(prefix)\(name)"
+    type = AssetsCatalog.Entry.EntryWithValue.type(for: item)
+  }
+
+  private static func type(for item: Constants.Item) -> String {
+    switch item {
```
AliSoftware: Is that for context-facing name? Not sure why we'd want a different name than the extension, except for readability? Using the extension (aka `Constants.Item.rawValue`) would be maybe less readable in the context but it's not like it would be user-facing in the generated code then, right? And at least using the extension would have consistency and avoid template authors having to know about this mapping?

Sources/SwiftGenKit/Stencil/AssetsCatalogContext.swift
```diff
@@ -18,45 +18,54 @@ extension AssetsCatalog.Parser {
...
+  private static func countTypes(catalogs: [AssetsCatalog.Catalog]) -> [String: Int] {
+    var result: [String: Int] = [:]
+    countTypes(entries: catalogs.flatMap { $0.entries }, into: &result)
+    return result
+  }
+
+  private static func countTypes(entries: [AssetsCatalogEntry], into result: inout [String: Int]) {
+    for entry in entries {
+      if let entry = entry as? AssetsCatalog.Entry.EntryWithValue {
+        result[entry.type, default: 0] += 1
```
AliSoftware: ❤️

SwiftGen.playground/Pages/XCAssets-Demo.xcplaygroundpage/Sources/Implementation Details.swift
```diff
@@ -1,24 +1,66 @@
...
+  public typealias Color = NSColor
+  #elseif os(iOS) || os(tvOS) || os(watchOS)
+  public typealias Color = UIColor
+  #endif
+
   @available(iOS 11.0, tvOS 11.0, watchOS 4.0, OSX 10.13, *)
-  public var color: AssetColorTypeAlias {
-    return AssetColorTypeAlias(asset: self)
+  public var color: Color {
+    return Color(asset: self)
```
AliSoftware, line 63: I'm pretty sure it's ok, but just to be sure, did you check in practice that, if somewhere else in the user' source one declares a `struct Color` at top level for example, this won't produce an ambiguity error? I think it's ok because it solves by the closest match iirc, but if it's not we should use the fully qualified type name here and in the return type signature

Tests/Fixtures/Generated/XCAssets/swift3/all-customName.swift
```diff
@@ -2,15 +2,20 @@
...
-#elseif os(iOS) || os(tvOS) || os(watchOS)
-  import UIKit.UIImage
-  internal typealias XCTColor = UIColor
-  internal typealias XCTImage = UIImage
+  import AppKit
+#elseif os(iOS)
+  import ARKit
+  import UIKit
+#elseif os(tvOS) || os(watchOS)
+  import UIKit
```
AliSoftware, line 10: iirc in a previous commit before force-push, you added ARKit under `if is(tvOS/watchOS)` too? So I'm curious, did our unit tests, especially the rake trying to compile generated code, properly caught that `import ARKit` didn't work on tvOS/watchOS, before you fixed that? If not (= if you did just realize that by yourself not thanks to tests) we should make sure our `rake output:compile` catches that in the future.

templates/xcassets/swift3.stencil
```diff
@@ -3,23 +3,33 @@
...
+  import ARKit
+{% endif %}
+  import UIKit
+#elseif os(tvOS) || os(watchOS)
+  import UIKit
 #endif
 
+// Deprecated typealiases
+{% if resourceCount.color > 0 %}
+@available(*, deprecated, message: "This typealias will be removed in SwiftGen 7.0")
```
AliSoftware: Shouldn't we use `@available(*, renamed, …)` instead of `deprecated`? This makes Xcode provide a nice fix-it

djbe: Uhm, it becomes something like `ColorAsset.Color`. Not sure if we want to have people actually use those type(alias)es? They're an internal artefact of having cross-platform code.

But yeah, I can switch to `renamed`.

Tests/Fixtures/Generated/XCAssets/swift3/all-customName.swift
djbe: I'm sure it never was under watchOS/tvOS from the start. I checked apple docs before I initially started to modify the template. Maybe it was some artefact of a bad merge?

Do keep in mind that the swift3 templates are no longer compiled, so CI can't catch it in this case. The swift 4 output is compiled on all 4 platforms.

AliSoftware: Yeah maybe it was also me misreading the diff on my last skimming then 😅

AliSoftware: (Note that my comment wasn't specially on swift3 only, I didn't even realize that I made that comment on Swift 3 template specifically, it just happened to be the first instance where I wondered about that during my review ^^)

templates/xcassets/swift3.stencil
AliSoftware: Well people weren't supposed to directly use that root typealias in the first place even before that namespacing so it always was an internal artefact of cross platform anyway. It should be pretty rare to find users using those typealiases and who would require them to rename. But for those very rare who do…

SwiftGen.playground/Pages/XCAssets-Demo.xcplaygroundpage/Sources/Implementation Details.swift
djbe: Jep, no issues.

Sources/SwiftGenKit/Parsers/AssetsCatalog/CatalogEntry.swift
djbe: Right, but that's a breaking change. Something for a PR in 7.0.

AliSoftware: Ah, missed that

Sources/SwiftGenKit/Parsers/AssetsCatalog/CatalogEntry.swift
djbe: Do you mean `Constants.Item`? See comment below about breaking change.

djbe: We can always use that type, and have the context function transform it into a "stencil" type string. And drop that transform in some future breaking version.

AliSoftware: Yeah that's what I meant. Still feels strange for this to be a String while we only support a subset of possible values.
Couldn't we declare this as `type: Constants.Item`, then call `.name` on it when we expose it in the context, where `var name: String` would be a computed variable on `Constants.Item` doing the switch, moving the code currently in the `static func type(for:)` in that computed var?

djbe: Yeah that's similar to what I was thinking too.

Sources/SwiftGenKit/Parsers/AssetsCatalog/CatalogEntry.swift
AliSoftware: Not sure where that `type` string was used before in the context though, when I saw it here I assumed you added it to expose such type in the new `resourceCount` entry in the context, didn't realize it was already used elsewhere in the context too

djbe: In the template? It's used to know which asset type to generate.

AliSoftware: Ah got it 👍

Sources/SwiftGenKit/Stencil/AssetsCatalogContext.swift
```diff
@@ -18,45 +18,69 @@ extension AssetsCatalog.Parser {
...
+    return [
+      "type": item.type,
+      "name": name,
+      "value": value
+    ]
+  }
+}
+
+extension Constants.Item {
+  var type: String {
```
AliSoftware: Now wondering if `type` is the right name for this property… maybe `name`? (or `contextName` or `description` or whatever?)

SwiftGen.playground/Pages/XCAssets-Demo.xcplaygroundpage/Contents.swift
```diff
@@ -66,11 +65,20 @@ internal enum Asset {
       internal static let tint = ColorAsset(name: "Vengo/Tint")
     }
   }
+  enum Targets {
+    internal static let bottles = ARResourceGroupAsset(name: "Bottles")
+    internal static let paintings = ARResourceGroupAsset(name: "Paintings")
+	internal static let posters = ARResourceGroupAsset(name: "Posters")
```
AliSoftware: Not sure if GitHub misleading me again or not but indentation don't seem to be the same on line 71 than on line 69 and 70… 🧐

djbe: Seems Xcode ignores the project indentation setting in playground pages/sources 🤦‍♂️ 
Did a project-wide search & replace for tabs.

Sources/SwiftGenKit/Stencil/AssetsCatalogContext.swift
djbe: Renamed it to `typeName`.

AliSoftware (approved):

PR #555, opened 2018-10-22
Fix Swift version for swiftenv #trivial

usami-k:
Fix build error on swiftenv environment.

```
$ swiftenv versions
* 4.0 (set by /Users/usami/repo/github.com/SwiftGen/SwiftGen/.swift-version)
  4.1
  4.1.2
  4.2
$ swift package --version
Swift Package Manager - Swift 4.0.0-dev
$ swift build
error: package at '/Users/usami/repo/github.com/SwiftGen/SwiftGen' requires a minimum Swift tools version of 4.2.0 (currently 4.0.0)
```

PR #619, opened 2019-03-01
Cocoapods 1.9

djbe:
2 things of note:
- CDN sources, which allows us to get rid of the (still slow) step of downloading the full repo. We can now also cache the specs folder as it isn't as massive.
- `generate_multiple_pod_projects` and `incremental_installation`, which should reduce the amount of changes during `pod install/update`.


SwiftGen-Eve:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">You should make sure that <code>Podfile.lock</code> and <code>Package.resolved</code> are changed in sync</td>
    </tr>
  </tbody>
</table>

---

</p>

djbe:
Would you look at that, 4 seconds (without cache) 😆 🚗💨

AliSoftware (commented):
We should also remove the deprecated `.swift-version` from the repo and use the `swift_versions` DSL attribute in the podspec instead

djbe:
Uhm we can add the `swift_versions` bit, but some users seem to need that file to be there, see #555.

.circleci/config.yml
```diff
@@ -143,29 +143,31 @@ jobs:
...
+          keys:
+            - cocoapods-{{ checksum "Podfile.lock" }}
+            - cocoapods-
       - run:
           name: Lint podspec
           command: bundle exec rake pod:lint
+      -  save_cache:
+          key: cocoapods-{{ checksum "Podfile.lock" }}
+          paths:
+            - ~/.cocoapods
```
AliSoftware, line 156: With CDN, is it really still worth it to cache the `~/.cocoapods` folder? From experience on other repos it doesn't really seem to be much necessary anymore.

We already commit the `Pods/` folder (and if we didn't, we could cache it) so it should be quite rare to have the need for the spec repo to be present and downloaded locally anyway, as it is now supposed to only be necessary during `pod update` (or if `pod install` needs to actually install pods that were missing in `Pods/` and were not at a version that was matching was was locked in `Podfile.lock`, which in our case should never happen on CI)

Gemfile
```diff
@@ -4,19 +4,19 @@ source 'https://rubygems.org'
 
 # The bare minimum for building, e.g. in Homebrew
 group :build do
-	gem 'rake', '~> 12.3'
+	gem 'rake', '~> 13.0'
```
AliSoftware, line 7: 🎉 This should get rid of the security risk issue too 👍

AliSoftware (commented):
Other than the fact that we should stop probably caching the CDN (and thus winning some more CI time by avoiding to restore it when not needed), LGTM

.circleci/config.yml
AliSoftware: iow, if you remove those steps from CI, and as long as we still commit the `Pods/` folder and that `Podfile.lock` and `Pods/Manifest.lock` are properly in sync, I think `pod install` is now a no-op that shouldn't even require the CDN to be cloned at all.
Maybe I'm wrong so don't hesitate to make some tests to validate that if unsure.

AliSoftware (approved):
I thought the risk of using cache for specs repo was that cache restoration (download+unzip) and save (zip + upload) would risk taking more time than cloning the CDN. But after looking into it, I see that cache restoration only takes 3s. So we might as well keep it

PR #620, opened 2019-03-03
Strings: generate `Any` parameter type for `%@` placeholders (instead of `String`)

djbe:
Not sure if the changelog entry should be under "breaking changes"  or not.  It's a bug fix, but it may somehow break people's code? No idea TBH.

Note: I've purposefully *not* modified the stencil context, it would be a major breaking change to do so. We already have a task for 7.0: https://github.com/SwiftGen/SwiftGen/projects/3#card-13618963


SwiftGen-Eve:

---

</p>

Documentation/templates/strings/flat-swift4.md
```diff
@@ -28,15 +28,19 @@ You can customize some elements of this template by overriding the following par
...
   /// You have %d apples
-  case applesCount(Int)
+  internal static func applesCount(_ p1: Int) -> String {
+    return L10n.tr("Localizable", "apples.count", p1)
+  }
   /// Those %d bananas belong to %@.
-  case bananasOwner(Int, String)
+  internal static func bananasOwner(_ p1: Int, _ p2: Any) -> String {
+    return L10n.tr("Localizable", "bananas.owner", p1, String(describing: p2))
+  }
```
AliSoftware, line 43: We should find a way to ensure those get updated regularly, too easy to forget… (Danger rule probably easier, script auto-updating the doc when generating outputs would probably be too complicated if we want to omit stuff for brevity in those docs)

AliSoftware (approved):

Documentation/templates/strings/flat-swift4.md
djbe: Danger script sounds like a good idea, anytime the template is changed, and not the corresponding docs --> warning.

acecilia:
Hey 👋 
I just found out about this PR. Could you guys explain a bit the rationale behind it? 
* I thought `%@` is supposed to be a string in iOS 🤔 
* Is there any way to have a type safe string with swiftGen after this? Maybe `%s`?

AliSoftware:
Hello @acecilia !


> * I thought `%@` is supposed to be a string in iOS 🤔 

Nope, `%@` placeholder has always meant to be an `NSObject` even since the Objective-C times (and now `Any` in Swift).

Internally at runtime iOS (and the implementation of `String(format:)` specifically) calls `String(describing: theObjext)` on the object you passed for this parameter, to translate that object into a `String` to use where the `%@` placeholder was.

- If that object is a String already, `String(describing: str)` happens to return the string `str` as is
- If that's an object which responds to `CustomStringConvertible` then iOS will call its `description` property.
- If it's none of them it will just use `Mirror` to try to build some string description of the object.

This has always been how Objective-C and Swift (since Swift 1 iinm) have worked with `%@`.

acecilia:
Thanks for the context 🙏   Since now `%@` is converted to `Any`, what is the format that results in a type safe `String`?

AliSoftware:
Well there isn't really one because iOS doesn't provide any placeholder for a `String` type in their `String(format:)` RFC/spec 🤷‍♂️
So since there is no `%…` placeholder for parameters of type `Swift.String` provided by iOS itself, there's not much we can really do in our end — as we're bound to respect the RFC/Spec and the behavior that `String(format:)`'s own implementation and spec imposes on us as our goal is to match what will happen at runtime as per `Swift(format:)` own implementation.

If you really want SwiftGen to generate code that only accepts parameters of type `String` for `%@` (even if technically those placeholders accept any object not just Strings), it's easy enough though: just duplicate the built-in template and change the place where you see `Any` to be `String` in your template copy, and that's it, you can use your modified template as a custom template. _For more info about how to do that and the commands that can help you achieve that with SwiftGen, see the dedicated documentation in `Documentation/` folder in the repo_

PR #638, opened 2019-05-15
Fonts: add support for extracting code points

robbiet480:
This adds support to the SwiftGen fonts context for extracting code points of a font to make them available to the template. This is useful for generating templates for icon fonts. I am doing this work to modernize [Iconic](https://github.com/dzenbot/Iconic) which until this point used a old, modified, Swift 2.3 only version of SwiftGen which meant that recent users have been unable to update their icon fonts. I have adapted the Iconic SwiftGen template but have not submitted it as part of this PR because I was unsure if it was out of scope. If you don't think it will be I will be happy to submit it. I quickly put it up as a [Gist](https://gist.github.com/robbiet480/087c39461b5b8f146cad8e926ccc78b1) if you want to take a look.

Related PR (required to run the linked template, not required for this PR to be merged): https://github.com/SwiftGen/StencilSwiftKit/pull/121

* [X] I've started my branch from the `develop` branch (gitflow)
  * [X] And I've selected `develop` as the "base" branch for this Pull Request I'm about to create
* [X] I've added an entry in the `CHANGELOG.md` file to explain my changes and credit myself  
  * [X] Add the entry in the appropriate section (Bug Fix / New Feature / Breaking Change / Internal Change)
  * [X] Add a period and 2 spaces at the end of your short entry description
  * [X] Add links to your GitHub profile to credit yourself and to the related PR and issue after your description

---


robbiet480:
~~Seems Circle is broken on the branch. I did have to remove the opt in rule `lower_acl_than_parent` from the SwiftLint config to get things building locally. All the ACL issues were in files that I didn't touch though.~~

AliSoftware:
Thanks for the PR (and sorry for the late reply)

Could you update the contexts files with this new code so we can see how the updated YAML contexts look like, and also so that the unit tests are updated accordingly? It's just a matter of using Cmd-U on the "SwiftGenKit - Generate Contexts" scheme in Xcode

I have two main concerns with this change:

- [ ] If I understand the code correctly, it will generate a HUGE list of names for standard fonts. For fonts that are limited to only a couple of glyphs, like "icon fonts" that might be OK. But for a normal font that contains tons of glyphs for almost all possible codepoints in unicode, that is gonna generate a HUGE list. We likely don't want that for most fonts
- [ ] I tried to re-generate the SwiftGenKit contexts (by running the scheme as pointed above) to run the unit tests and it crashes. This is because for regular fonts like the ones we have in our unit tests, like `.SFNSDisplay-Regular` for example, there are multiple glyphs with the same name (in my tests it crashes with `Fatal error: Duplicate values for key: 'space'`)

So in the end I'm not sure if this makes sense to generate those lists. I understand the use case, but that won't work for regular fonts though. Maybe we could make use of parser options so that the user can specify when they want those list of glyph names generated for which fonts, and default to not generate them? That way users can opt-in for this behaviour only for their icon-fonts and avoid the list of glyphs for regular fonts?

djbe:
Ideally we'd have a mechanism for lazily calculating certain context properties (+ cache, so `lazy var`?), such as the "code points" in this PR. But that doesn't really work with how our contexts are generated right now (plain dictionaries), and how we serialize them for testing purposes.

SwiftGen-Eve:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

---

</p>

robbiet480:
@AliSoftware I've gone ahead and added a option for extracting code points. Not sure what I need to do in the tests to get it actually testing though.

robbiet480:
Never mind, all figured out! Think this is good to go?

robbiet480:
Now that I'm actually integrating this, I've realized that this can't yet be merged. For one thing, the option values need to be String so it works with CLI. But now I'm facing this issue and unsure how to get past it. Just pushed the font that has this code point at 7132bca11afd8784766b8694817bf8213f97a384. Suggestions welcome!

```swift
let point: UInt32 = 0xF0000
let scal = Unicode.Scalar(point)!
let char = UniChar(scal.value) // Fatal error: Not enough bits to represent the passed value
```

djbe:
@robbiet480 Not sure what you're trying to do there, but that code point `0xF0000` does not fit inside an `Int16` (which `UniChar` is a typealias of).

robbiet480:
Yeah its a codepoint that came out of the test icon font I'm using, [MaterialDesignIcons](https://materialdesignicons.com/). Guess they have so many icons the code points go out of UInt16 space?

djbe:
You can write a unicode character in a Swift string like this:
```swift
let dollarSign = "\u{24}"        // $,  Unicode scalar U+0024
let blackHeart = "\u{2665}"      // ♥,  Unicode scalar U+2665
let sparklingHeart = "\u{1F496}" // 💖, Unicode scalar U+1F496
```

AliSoftware:
Yeah what you probably want is to convert your `Unicode.Scalar` into a `Character`, not a `UniChar`

https://developer.apple.com/documentation/swift/character/2906502-init

in Unicode, code points are not limited to 2^16 so it's not logical to constrain them to `UInt16`. That's why we use `Character` in Swift when manipulating strings, and not `UniChar` which are more a relic of the past from C before full-unicode compliance

robbiet480:
@AliSoftware That's hopefully the pointer I needed, thanks!!!

robbiet480:
Okay, think it's now all good.

CHANGELOG.md
```diff
@@ -53,6 +53,9 @@
   [#609](https://github.com/SwiftGen/SwiftGen/issue/609)
   [#593](https://github.com/SwiftGen/SwiftGen/pull/593)
   [#610](https://github.com/SwiftGen/SwiftGen/pull/610)
+* Fonts: extract all codepoints of a font and make available to templates. Useful for icon fonts.
```
AliSoftware, line 56: Worth adjusting that CHANGELOG entry to reflect that this is opt-in

Would also rather mention that it extract "glyph names" rather than talking about codepoints, as glyph names are what are more relevant to users here

Sources/SwiftGenKit/Parsers/Fonts/CTFont.swift
```diff
@@ -5,27 +5,87 @@
...
     return descRefs.compactMap { desc -> Fonts.Font? in
       let font = CTFontCreateWithFontDescriptorAndOptions(desc, 0.0, nil, [.preventAutoActivation])
       let postScriptName = CTFontCopyPostScriptName(font) as String
       guard let familyName = CTFontCopyAttribute(font, kCTFontFamilyNameAttribute) as? String,
         let style = CTFontCopyAttribute(font, kCTFontStyleNameAttribute) as? String else { return nil }
 
       let relPath = parent.flatMap { file.relative(to: $0) } ?? file
+
+      let characterSet = CTFontCopyCharacterSet(font) as CharacterSet
+      let cgFont = CTFontCopyGraphicsFont(font, nil)
```
AliSoftware, line 27: Why not put these 2 inside the `if extractCodePoints`?
That might not take _that_ long to run (though maybe that does depending on the font file size?) but better avoid needless calls when the flag is off 🙃

AliSoftware: Also, not completely sure because maybe the bridging of that API to Swift takes care of that, but worth double-checking: since those two calls are `…Copy…` don't they need a balancing `CTRelease` (or `CFRelease`?) call or similar to avoid leaks?

Sources/SwiftGenKit/Parsers/Fonts/CTFont.swift
```diff
@@ -5,27 +5,87 @@
...
+
+      var icons: [String: String] = [:]
+
+      if extractCodePoints {
+        for unicode in characterSet.codePoints {
+          guard let unicodeScalar = UnicodeScalar(unicode) else { continue }
+
+          let utf16codepoints = Array(Character(unicodeScalar).utf16)
```
AliSoftware, line 35: Wondering if there isn't a more efficient way to go from codepoint to utf16 than CodePoint>UnicodeScalar>Character>.utf16 🤔
Will have to look at the various String/Character/Unicode APIs to double check, as this is often tricky stuff with edge cases so better be sure.

Sources/SwiftGenKit/Parsers/Fonts/CTFont.swift
```diff
@@ -5,27 +5,87 @@
...
+
+          var glyphs = [CGGlyph](repeating: 0, count: utf16codepoints.count)
+
+          // Gets the Glyph
+          guard CTFontGetGlyphsForCharacters(font, utf16codepoints, &glyphs, utf16codepoints.count) == true else {
+            print("Failed to get glyph for character", unicode, utf16codepoints, unicodeScalar)
```
AliSoftware, line 41: We have a `logMessage` function to generate warnings and errors during parsing, you should use that instead of `print` here

Sources/SwiftGenKit/Parsers/Fonts/CTFont.swift
```diff
@@ -5,27 +5,87 @@
...
+            continue
+          }
+
+          if glyphs.isEmpty {
+            print("Glyph set was empty!")
```
AliSoftware, line 46: Ditto re:logMessage

Sources/SwiftGenKit/Parsers/Fonts/CTFont.swift
```diff
@@ -5,27 +5,87 @@
...
+
+          var glyphs = [CGGlyph](repeating: 0, count: utf16codepoints.count)
+
+          // Gets the Glyph
+          guard CTFontGetGlyphsForCharacters(font, utf16codepoints, &glyphs, utf16codepoints.count) == true else {
+            print("Failed to get glyph for character", unicode, utf16codepoints, unicodeScalar)
+            continue
+          }
+
+          if glyphs.isEmpty {
```
AliSoftware, line 45: Nit: might make more sense to write this as a `guard !glyphs.isEmpty` semantically speaking 🤷‍♂️

Sources/SwiftGenKit/Parsers/Fonts/FontsParser.swift
```diff
@@ -9,6 +9,14 @@ import Foundation
 import PathKit
 
 public enum Fonts {
+  public enum Option {
+    static let codePoints = ParserOption(
+      key: "codePoints",
+      defaultValue: "false",
+      help: "If enabled, will extract code points from the font. Only useful for icon fonts."
```
AliSoftware, line 16:
```suggestion
      help: "If enabled, will extract the glyph names from the font. Only useful for icon fonts."
```

Sources/SwiftGenKit/Parsers/Fonts/FontsParser.swift
```diff
@@ -9,6 +9,14 @@ import Foundation
 import PathKit
 
 public enum Fonts {
+  public enum Option {
+    static let codePoints = ParserOption(
```
AliSoftware, line 13: Maybe name the option `glyphNames` instead of `codePoints`?

Sources/SwiftGenKit/Parsers/Fonts/FontsParser.swift
```diff
@@ -32,7 +41,13 @@ public enum Fonts {
         return
       }
 
-      let fonts = CTFont.parse(file: path, relativeTo: parent)
+      let getCodePoints: Bool = self.options[Option.codePoints] == "true"
```
AliSoftware, line 44: We need to check how we do string-to-boolean for other config flags. I don't think we have many Boolean Parser Options that we can use as reference but we do have some template parameters that are Boolean (e.g. `forceNameSpaces` for the Assets templates); if there is similar cases, better ensure we do the conversion the same way (`== "true"` vs `!= "false"`, case-insensitivity, etc)

Sources/SwiftGenKit/Stencil/FontsContext.swift
```diff
@@ -18,15 +18,17 @@ extension Fonts.Parser {
...
         let fonts = family
-          .map { (font: Fonts.Font) -> [String: String] in
+          .map { (font: Fonts.Font) -> [String: Any] in
             [
               "name": font.postScriptName,
               "path": font.filePath,
-              "style": font.style
+              "style": font.style,
+              // swiftlint:disable:next force_unwrapping
+              "icons": font.icons.map { ["name": $0.key, "unicode": $0.value] }.sorted { $0["name"]! < $1["name"]! }
```
AliSoftware, line 27: Even if the force-unwrap is technically never supposed to be nil in that specific case, you could avoid the SwiftLint violation using
```
($0["name"] ?? "") < ($1["name"] ?? "")
```

AliSoftware, line 27: You'll need to document that additional key in the documentation for contexts (in `Documentation/StencilSwiftKit Contexts/fonts.md`)

AliSoftware (commented):
Quick review (from phone so might have missed things?) of those latest changes, looking good!

Left a couple of suggestions, and one or two interrogations about things we might want to just conform to be sure, but otherwise LGTM 👍

AliSoftware:
Also, you mention in the PR description that you have a template that includes those glyph names but kept it as a gist and didn't include it in the PR, but I think it might be worth it to make some use of it.

I think we should either:
 - create a separate template eg `iconfont-swift5.stencil` that exposes those icons and glyph names in the generated code
 - or modify the existing templates so that they accept an `includeGlyphNames` parameter, or even better, maybe we don't need the explicit parameter and can just do something like `{% if icons.empty %}// List of glyphs not generated. If this is an icon font and you want the list of glyphs there, use "option: glyphNames" as an option for the parser in your config file{% else %}let glyphNames = …code to generate the list{% endif %}`
 - or keep the template as a gist but mention the URL to the gist somewhere in the documentation. But at that point I think we would be better off including the template in the built in templates

So I'd vote for 1 or 2, using solution 2 only if it's a small and localized change to the existing templates that won't make them too convoluted or to hard to maintain. Otherwise or if the code for icon fonts makes more sense to be very specific and different from normal fonts, go with 1 and have a dedicated template for them, which could actually make more sense in practice and would be easier to do given you already have a gist for it.

(Note that we could keep that addition of your template for icon fonts as a separate PR if you prefer to keep things separate and easier to review)

Sources/SwiftGenKit/Parsers/Fonts/FontsParser.swift
AliSoftware: Also, this parser option needs to be documented in `Documentation/Parsers/fonts.md` in the "Supported Options" section (you might need to merge develop into your branch first, as we renamed that documentation folder in a recent PR that was open and merged after you opened yours)

Sources/SwiftGenKit/Parsers/Fonts/FontsParser.swift
AliSoftware: Might actually be a nice addition to add to `ParserOptionValues` by overloading the subscript si that if the return type is inverted to `Bool`, that `ParserOptionValues` is the one doing the string-to-Bool conversion. That would allow us to ensure that all future Boolean parser options are treated the same way.

Sources/SwiftGenKit/Parsers/Fonts/CTFont.swift
AliSoftware: Actually there is, and it's more correct: `Array(unicodeScalar.utf16)`

AliSoftware:
So… I tested the code with `SF-Text-Pro-Regular.otf` (SFSymbols) and I've managed to make it list the glyph names but only the ones for the common textual characters. It won't output any of the SFSymbol names like, say, `camera.fill` or similar… worth investigating why and if there's a solution for that.

I've actually managed to get the unicode code point of one of the SFSymbols (drag & drop one of the icon from the SF Symbols macOS app into a new file in a text editor, save in utf16, use `xxd` to show the utf16 hex values) but when I try to call  `CTFontGetGlyphsForCharacters` with that surrogate pair of utf16 code units, it fails (the method returns `false`). So no idea how we could get a `CGGlyph` for those characters. And `cgFont.numberOfGlyphs` returns `7645` which is the number of `case` that the current code generates, which means that this `cgFont.numberOfGlyphs` doesn't take the SFSymbol glyphs present in the SFSymbol font into account anyway… 🤔

PR #644, opened 2019-06-14
Add bundle detection in dynamic frameworks

PatrickNLT:
When using SwiftGen inside a dynamic framework (eg. integrated using CocoaPods), `Bundle(for: BundleToken.self)` will return the framework itself, not the associated bundle. Thus, no resource can be found.
This merge request fixes this issue by going deeper, using the first available bundle in that framework. Since the `bundle` variable is already cached, there is no noticeable performance hit.


SwiftGen-Eve:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

---

</p>

djbe:
That's something that I've encountered a few times as well, but your solution is dangerous. You're assuming there will only be 1 bundle inside the framework. I could just as well split up some resources into different bundles.

How I solved it in a few projects is having a custom template that accepts a bundle name as a parameter: https://github.com/appwise-labs/AppwiseCore/blob/master/Sources/strings.stencil#L63-L70
Note that this is a custom template that makes the assumption that the user'll always provide a bundle parameter, which should be fixed.

PatrickNLT:
I agree that this solution has its limits, but I think it's still better than nothing and the best we can do without dealing with additional parameters. I'm not sure it's _dangerous_, though.
It could also serve as a basis for a parametrized template.

AliSoftware:
Can one of you explain to me the use case for this? I'm not sure I understand the kind of configuration / framework structure you're using there. How come your bundle be inside your framework, given that a `.framework` is already itself a bundle and that nesting frameworks into other frameworks is not allowed (Xcode makes it possible but that will be rejected on the AppStore).

Or are you talking about embedding just your framework's resources inside a sub-bundle, so inside `MyApp.app/Frameworks/MyFmk.fmk/` you have the `MyFmk` dylib but instead of also having the resources at its root, you have them inside a `MyApp.app/Frameworks/MyFmk.fmk/Resources.bundle/Localizable.strings` for example?

if that's the latter, I'd really prefer using the feature of template parameters to allow users of SwiftGen to provide which sub-bundle we should default to if not the framework's, as as David said just relying on the first bundle found inside the framework seems quite arbitrary and could be flaky.

djbe:
@AliSoftware Exactly. For example, if you create a pod and use the (recommended) `resource_bundles` option (https://guides.cocoapods.org/syntax/podspec.html#resource_bundles), then you get `.bundle`s inside your framework.

The reason for that is if the user selected static linking (instead of frameworks), there are no frameworks but everything is put in the root (bundle) of the app, so with resource bundles the pod's resources don't collide with the app's resources.

Like I mentioned before, it's dangerous to assume there's going to be only 1 bundle, and that the resources are going to be in there. The templates embedded in SwiftGen should be "best in class" and be as safe as possible. So that's why a template parameter with a bundle name is the better option.

- If the template parameter is empty --> no sub bundle, generate existing code
- If the template parameter is filled in by the user --> generate code to load from specified bundle

djbe:
The mechanism discussed above (by @AliSoftware and myself) has been implemented in #737 and scheduled for release in SwiftGen 6.3.0. It allows users to set a `bundle` template parameter to force SwiftGen to load resources from a specific bundle. We've even written a full guide on how to use this feature (and the more advanced `lookupFunction` parameter).

I appreciate the effort put into this PR, but I'm going to close it as it has been superseded by #737.

Issue #662, opened 2019-10-30
Plist and inline-swift template not working with Arrays


fjtrujy:
Hello,
I'm trying to use the template `plists/inline-swift4.stencil` template to generate the content of a plist.

So far the content of the plist is:

```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>Names</key>
	<array>
		<string>Jhon</string>
		<string>Peter</string>
		<string>Nick</string>
	</array>
	<key>Surnames</key>
	<array>
		<string>Smith</string>
		<string>Jhonson</string>
		<string>Williams</string>
	</array>
</dict>
</plist>
```

And visually it looks:

![Screenshot 2019-10-30 at 12 53 11](https://user-images.githubusercontent.com/10010409/67856520-021f4700-fb15-11e9-8ff0-5e502c606726.png)

Then in the `.swiftgen.yml` I just have somethings as:

```yml
.........
plist:
    inputs: Precompile/OriginPlists/
    outputs:
        templateName: inline-swift4
        output: Generated/Plist.swift
.....
```

When I execute it, it generates wrong data:
```swift
internal static let names: [String] = [Jhon, Peter, Nick]
internal static let surnames: [String] = [Smith, Jhonson, Williams]
```

If you see it misses the `quotes` in every single element in the array, and in the plist, they are declared as `String`

Digging into the template I found where the issue "could be", I have created a solution, I tried locally and is working fine, but I don't know if this is breaking something else. Besides I see that the same logic is applied in several templates, so not sure if in these templates is broken as well.

The solution that I applied in the `plists/inline-swift4.stencil` was:

```diff
@@ -55,9 +55,9 @@ import Foundation
     Date(timeIntervalSinceReferenceDate: {{ value.timeIntervalSinceReferenceDate }})
   {% elif metadata.type == "Optional" %}
     nil
-  {% elif metadata.type == "Array" and metadata.element.items %}
-    [{% for itemMetadata in metadata.element.items %}
-      {% call valueBlock value[forloop.counter0] itemMetadata %}
+  {% elif metadata.type == "Array" and value %}
+    [{% for currentValue in value %}
+      {% call valueBlock currentValue metadata.element %}
       {% if not forloop.last %}, {% endif %}
     {% endfor %}]
[...]

Issue #551, opened 2018-10-12
JSON parser strips newlines


yonaskolb:
SwiftGen version 6.0.2

We have a json file that has newlines in some of it's values
```json
{
    "value": "This is a \n multiline string,
}
```
When output via the `inline-swift4` template it seems that newlines are stripped at some level so that the template doesn't output them.
We actually have a custom template based of the default so if we can fix it somehow in there that's great, otherwise it might require a parser fix

djbe:
The `removeNewlines:"leading"` filter here is the culprit:
https://github.com/SwiftGen/SwiftGen/blob/master/templates/json/inline-swift4.stencil#L52

I think the best solution may be passing the value of a `String` through a `replace` filter, replacing newlines with `\n`. @yonaskolb How'd you fix it in your custom template?

djbe:
Note: this also applies to the inline plist templates.

djbe:
Just tried it with `{{ value|replace:"\n","\\n" }}`, which doesn't work as the replace filter receives `"\\n"` as search string, and `"\\\\n"` as replacement string. So it should become `{{ value|replace:"\n","\n" }}`, but again, that doesn't work because the search string is still escaped. I tried using an actual newline, but that gets converted to a ` ` (space).

The issue is that both the `propertyBlock` and `valueBlock` macros are wrapped in `removeNewlines` filters, and recurse for arrays/dictionaries, so removing those filters isn't a good idea. Also the `swift3` templates can't use multiline strings 🤷‍♂️

AliSoftware:
Side question: should we use multiline strings in the `inline-swift4` template if we detect (via a Stencil tag) a `\n` in the string? (Which we currently can't because of SwiftGen/StencilSwiftKit#109 I guess), for ease of readability? Or would that render the template itself too complex to understand for little benefit?

djbe:
I tried the multiline strings path, but it's really hard because of the recursion.

djbe:
The template becomes this (wip):
```
{% macro valueBlock value metadata %}{% if metadata.type == "String" %}"""
    {{ value }}
    """
  {% elif metadata.type == "Optional" %}nil
  {% elif metadata.type == "Array" and metadata.element.items %}[
    {% for itemMetadata in metadata.element.items %}
      {% call valueBlock value[forloop.counter0] itemMetadata %}
      {% if not forloop.last %}, {% endif %}
    {% endfor %}]
[...]

PR #687, opened 2020-03-29
Fix issue with Array in the Inline templates

fjtrujy:
## Description

This PR fixes issues with `plists/inline-swift4.stencil` or `plists/inline-swift5.stencil` templates to generate the content of a list that contains arrays.

So far the content of the plist is:

```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>Names</key>
	<array>
		<string>Jhon</string>
		<string>Peter</string>
		<string>Nick</string>
	</array>
	<key>Surnames</key>
	<array>
		<string>Smith</string>
		<string>Jhonson</string>
		<string>Williams</string>
	</array>
</dict>
</plist>
```

And visually it looks:

Screenshot 2019-10-30 at 12 53 11

Then in the `.swiftgen.yml` I just have somethings as:

```yml
.........
plist:
    inputs: Precompile/OriginPlists/
    outputs:
        templateName: inline-swift4
        output: Generated/Plist.swift
.....
```

When I execute it, it generates wrong data:
```swift
internal static let names: [String] = [Jhon, Peter, Nick]
internal static let surnames: [String] = [Smith, Jhonson, Williams]
```

If you see it misses the `quotes` in every single element in the array, and in the plist, they are declared as `String`

Digging into the template I found where the issue "could be", I have created a solution, I tried locally and is working fine, but I don't know if this is breaking something else. Besides I see that the same logic is applied in several templates, so not sure if in these templates is broken as well.

The solution that I applied in the `plists/inline-swift4.stencil` was:

```diff
@@ -55,9 +55,9 @@ import Foundation
     Date(timeIntervalSinceReferenceDate: {{ value.timeIntervalSinceReferenceDate }})
   {% elif metadata.type == "Optional" %}
     nil
-  {% elif metadata.type == "Array" and metadata.element.items %}
-    [{% for itemMetadata in metadata.element.items %}
-      {% call valueBlock value[forloop.counter0] itemMetadata %}
+  {% elif metadata.type == "Array" and value %}
+    [{% for currentValue in value %}
+      {% call valueBlock currentValue metadata.element %}
       {% if not forloop.last %}, {% endif %}
     {% endfor %}]
   {% elif metadata.type == "Dictionary" %}
```

The new output is:

```swift
internal static let names: [String] = ["Jhon", "Peter", "Nick"]
internal static let surnames: [String] = ["Smith", "Jhonson", "Williams"]
```

🎉 🎉 🎉 🎉 🎉 🎉 🎉 

Here you have the reference to the specific github issue that I created time ago. https://github.com/SwiftGen/SwiftGen/issues/662

Thanks!


SwiftGen-Eve:

---

</p>

AliSoftware:
Thanks for the PR!

It seems you didn't update the tests accordingly (or didn't add a test to cover this case that we missed). Could you add one please?
Also don't forget to credit yourself in the CHANGELOG ;)

CHANGELOG.md
```diff
@@ -54,6 +54,9 @@
 
 ### Bug Fixes
 
+* Plist: Fix issue with Array in the Inline templates.
```
  <tbody>
    <tr>
      <td>:warning:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/SwiftGen/SwiftGen/blob/4481992932bf16d3ef02b2ada1135c1a4a5e7889/CHANGELOG.md#L57"></span>Line describing your entry should end with a period and 2 spaces.</td>
    </tr>
  </tbody>
</table>

fjtrujy:
Hello,
@AliSoftware I will try to create a proper test to fix this issue.
Thanks!

djbe:
- I actually started updating this PR, and noticed the same change could be applied to json/yaml inline templates.
- Then after generating code (and adding a test case), I also noticed that this fix actually breaks code generation for mixed arrays (`[Any]`).
- ~Furthermore, I tried testing the old template with the given test case, and it just works...~

~Added my findings to #662, will need more testing to see if we actually need a PR.~
Edit: I wasn't pressing the right test shortcut 🤦

djbe:
Clarification for these latest changes:
- Old template assumed that there was always going to be `metadata.element.items[...]`, which isn't true for homogeneous arrays such as `[String]`
- The initial fix always passed along `metadata.element`, which isn't correct for mixed arrays, where we need the right metadata for each item.
- Correct solution is to try to use metadata for specific item, and fallback to general element metadata.

fjtrujy:
Thanks @djbe for progressing with the work.
I was not able to finish it.
Thanks again

Tests/Fixtures/Resources/YAML/good/mapping.yaml
```diff
@@ -4,3 +4,7 @@ foo:
   bar: banana
   baz: orange
 car: null
+names:
+  - John
+  - Peter
+  - Nick
```
AliSoftware, line 10: Not sure if this should be addressed by that PR or a separate one but I just realised that we might need to add test cases for strings containing (escaped) quotes like `- "Craig \"Hair Force One\" Federighi"` to see how our templates handle those… 

e.g. maybe we should at least use `##"raw string literals"##` in our stencil templates to limit the risk with those edge cases?

AliSoftware (commented):
The fix for the particular problem looks good to me.

But I think we might have another issue (and likely not just with strings in arrays but with strings in general) when they contain escaped quotes…
(Feel free to decide if it's better to either fix this in this current PR or to split that in a separate one)

Tests/Fixtures/Resources/YAML/good/mapping.yaml
djbe: Check #551, where we have issues with multiline strings. Whatever fix there should also escape quotes.

But yeah, that's for a separate PR.

AliSoftware: Ah cool, as long as that's tracked. Then let's merge this one and tackle the quote escapes and multilines issues separately via #551 👍

AliSoftware (approved):

PR #697, opened 2020-05-26
Introduce `swiftgen template doc`

AliSoftware:
* [x] I've started my branch from the `develop` branch (gitflow)
  * [x] And I've selected `develop` as the "base" branch for this Pull Request I'm about to create
* [x] I've added an entry in the `CHANGELOG.md` file to explain my changes and credit myself  
      _(or added `#trivial` to the PR title if I think it doesn't warrant a mention in the CHANGELOG)_
  * [x] Add the entry in the appropriate section (Bug Fix / New Feature / Breaking Change / Internal Change)
  * [x] Add a period and 2 spaces at the end of your short entry description
  * [x] Add links to your GitHub profile to credit yourself and to the related PR and issue after your description

---

This PR:

 - [aca2804a20afa43cf68f797d41b05e80cb646500] Renames occurrences of "subcommand" in the docs with "parser" or "parser command" as appropriate for consistency
- [cc56a41f61287b903a52afd250b85d85d6b9f4a1] Renames `swiftgen templates` subcommand to `swiftgen template` (singular) and deprecate the plural form of the subcommand, which will be removed in next major release
- [82ed9ddf4fed739b8eb1c3eeb223f34ea5fd74d8] Introduces the `swiftgen template doc [parser] [template]` to quickly open the documentation for templates on GitHub

PR #702, opened 2020-05-27
JSON: use JSONSerialization instead of Yams

### LINKED ISSUE #698: Use JSONSerialization instead of Yams for JSON
Right now, JSON parsing is done via `Yams.load_all` since JSON is a subset of Yaml. 

That's clever, but comes with a perfomance cost. For my project, I'm using SwiftGen to parse Lottie JSON animation files into enums and that's taking almost ~3s per (incremental build).

My initial thought was to add caching support (#695). But I think we can avoid that (for now?) and do something way simpler: using `JSONSerialization` instead. In my small test, this made `swiftgen` to take ~750ms instead.

This is what I did - I can clean this and submit a PR, just need some guidance:

```swift
// MARK: JSON File Parser

public enum JSON {
  public enum ParserError: Error, CustomStringConvertible {
    case invalidFile(path: Path, reason: String)

    public var description: String {
      switch self {
      case .invalidFile(let path, let reason):
        return "Unable to parse file at \(path). \(reason)"
      }
    }
  }

  public class Parser: SwiftGenKit.Parser {
    var files: [File] = []

    public required init(options: [String: Any] = [:], warningHandler: Parser.MessageHandler? = nil) {
    }

    public class var defaultFilter: String {
      return ".*\\.(?i:json)$"
    }

    public func parse(path: Path, relativeTo parent: Path) throws {
      files.append(try File(path: path, relativeTo: parent))
    }
  }

  // is it ok to duplicate this struct? I did it to avoid breaking any public APIs, but I guess I could extract it to another type and have a typealias here. Thoughts?
  struct File {
    let path: Path
    let name: String
    let documents: [Any]

    init(path: Path, relativeTo parent: Path? = nil) throws {
      guard let data: Data = try? path.read() else {
        throw ParserError.invalidFile(path: path, reason: "Unable to read file")
      }

      self.path = parent.flatMap { path.relative(to: $0) } ?? path
      self.name = path.lastComponentWithoutExtension

      do {
        let item = try JSONSerialization.jsonObject(with: data)
        self.documents = [item] // why does the Yaml version take an array? is this the right thing for JSON?
      } catch let error {
          throw ParserError.invalidFile(path: path, reason: error.localizedDescription)
      }
    }
  }
}


// this is pretty much the same as `Yaml.Parser`. What is the right abstraction here? Are we ok with duplicating it?
extension JSON.Parser {
  public func stencilContext() -> [String: Any] {
    let files = self.files
      .sorted { lhs, rhs in lhs.name < rhs.name }
      .map(map(file:))

    return [
      "files": files
    ]
  }

  private func map(file: JSON.File) -> [String: Any] {
    return [
      "name": file.name,
      "path": file.path.string,
      "documents": file.documents.map(map(document:))
    ]
  }

  private func map(document: Any) -> [String: Any] {
    return [
      "data": document,
      "metadata": Metadata.generate(for: document)
    ]
  }
}

```

marcelofabri:
Fixes #698


SwiftGen-Eve:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

---

</p>

Sources/SwiftGenKit/Parsers/Yaml/YamlParser.swift
```diff
@@ -41,13 +41,3 @@ public enum Yaml {
     }
   }
 }
-
-// MARK: JSON File Parser
-
-public enum JSON {
-  public final class Parser: Yaml.Parser {
-    override public class var defaultFilter: String {
```
djbe, line 49: As we're removing this override, we should change the `defaultFilter` definition above (in YAML parser) to:
```swift
public static let defaultFilter = "[^/]\\.(?i:ya?ml)$"
```

Sources/SwiftGenKit/Parsers/JSON/JSONParser.swift
```diff
@@ -0,0 +1,42 @@
...
+    public var warningHandler: Parser.MessageHandler?
+
+    public required init(options: [String: Any] = [:], warningHandler: Parser.MessageHandler? = nil) throws {
+      self.options = try ParserOptionValues(options: options, available: Parser.allOptions)
+      self.warningHandler = warningHandler
+    }
+
+    public class var defaultFilter: String {
+      "[^/]\\.(?i:json)$"
+    }
```
djbe: As mentioned below, this can become a `let` now that we don't need to override:
```suggestion
    public static let defaultFilter = "[^/]\\.(?i:json)$"
```

CHANGELOG.md
```diff
@@ -110,6 +110,9 @@
 * Updated Pods and Gems dependencies.  
   [David Jennes](https://github.com/djbe)
   [#684](https://github.com/SwiftGen/SwiftGen/pull/684)
+* Use `JSONSerialization` for parsing JSONs, making `json` actions faster.
```
AliSoftware: Actually we're currently having a debate about how we should call those "actions".

In the current documentation, we interchangeably call them `subcommands` (mainly that's legacy from when we didn't have a config file and those were proper command line subcommands) and `parsers`
I've started to uniformise the term we use to always be `parser` from now (see https://github.com/SwiftGen/SwiftGen/commit/aca2804a20afa43cf68f797d41b05e80cb646500 in #697) but if you have a suggestion for a better name we're all ears.

I personally think `parser` makes more sense than `action` (even from a user's perspective) and I fear that `action` might be too generic (like, when documenting commands like `swiftgen templates list --only <action>` or `swiftgen template cat <action> <templateName>` it feels strange in that context), wdyt?

```suggestion
* Use `JSONSerialization` instead of `Yams` for parsing JSONs, making the `json` parser faster.  
```

Sources/SwiftGenKit/Parsers/JSON/JSONParser.swift
```diff
@@ -0,0 +1,42 @@
...
+    var files: [File] = []
+    public var warningHandler: Parser.MessageHandler?
+
+    public required init(options: [String: Any] = [:], warningHandler: Parser.MessageHandler? = nil) throws {
+      self.options = try ParserOptionValues(options: options, available: Parser.allOptions)
+      self.warningHandler = warningHandler
+    }
+
+    public class var defaultFilter: String {
+      "[^/]\\.(?i:json)$"
```
AliSoftware: Completely unrelated to that PR but note to self: we definitively need to have a helper function to build those default filters based on a file extension ^^

SwiftGen.xcodeproj/project.pbxproj
```diff
@@ -689,6 +704,7 @@
 				C38E571320FF8F8F009C9C23 /* CoreData */,
 				DD9F687D20955DEF004EB720 /* Fonts */,
 				DDCCACD5209550F60017FB43 /* InterfaceBuilder */,
+				D4A93E25247E305200BE8651 /* JSON */,
```
AliSoftware, line 724: 👌 I appreciate that you inserted that new folder in the middle to keep alphabetical order 👍 🙂

Sources/SwiftGenKit/Stencil/JSONContext.swift
```diff
@@ -0,0 +1,39 @@
...
+
+import Foundation
+
+//
+// See the documentation file for a full description of this context's structure:
+// Documentation/SwiftGenKit Contexts/Yaml.md
+//
+
+extension JSON.Parser {
+  public func stencilContext() -> [String: Any] {
```
AliSoftware, line 15: _Note to self, just realised that while reviewing here: why don't we call those files `JSONParser+Context.swift` instead of `JSONContext.swift`? Maybe some cleanup to do in a separate PRs one day…_

Sources/SwiftGenKit/Parsers/JSON/JSONFile.swift
```diff
@@ -0,0 +1,32 @@
...
+//
+
+import Foundation
+import PathKit
+
+public extension JSON {
+  struct File {
+    let path: Path
+    let name: String
+    let documents: [Any]
```
AliSoftware: Maybe we could already change that to `let content: Any` (and replace `self.documents = [item]` by `self.content = item` in `init`) at the parser level, and do the "wrapping in a `documents` array" (to keep API compatibility) at the `stencilContext()` level (i.e. replace `"documents": file.documents.map(map(document:))` in the exposed dict with `"documents": map(content: file.content)`)?

AliSoftware (changes requested):
Thanks for the PR! Looks good 👍 
I just think we need to move some unit tests around to split YAML tests in 2 before merging though.

 - Currently, our Unit tests are only on Yaml, and we just have one `.json` fixture used by the Yaml tests. We should instead move that `json` fixture in its dedicated `Tests/Fixtures/Resources/JSON` folder, and maybe add one or two other JSONs (like maybe one whose root is a dict, another with an array instead…)
 - Then add a `Tests/SwiftGenKitTests/JsonTests.swift` test file and move the `func testJSON()` test case (from `YamlTests.swift`) in there (as well as create 1 or 2 more test cases for 1 or 2 other JSON fixtures?)
    - You'll have to add an entry to the `Fixtures.Directory` (see `TestsHelper.swift`), and create the corresponding subdirectory in `Fixtures/StencilContexts` for the JSON contexts to be generated there for the tests
    - Then be sure to change the `sub:` parameter in the implementation of `testJSON()` to use that new `Fixture.Directory.json`
    - Once you've done those changes, use the `SwiftGenKit - Generate Contexts` Xcode scheme to automatically re-generate the unit test fixture files
    - Tip: I suggest you to delete the content of the `Fixtures/StencilContexts/Yaml` before running that scheme, so that any file that is not needed (especially `configuration.yml` context generated from `configuration.json` input file will likely have to be deleted from the Yaml StencilContext subdir) will not stay there and only the necessary files will be re-created
 - Finally, split the `YAML-Demo` page in the demo Playground in 2 separate pages, extracting the JSON example in its own page

Sources/SwiftGenKit/Parsers/JSON/JSONFile.swift
marcelofabri: that's a good idea!

Sources/SwiftGenKit/Stencil/JSONContext.swift
```diff
@@ -0,0 +1,39 @@
...
+// SwiftGen
+// Copyright © 2019 SwiftGen
+// MIT Licence
+//
+
+import Foundation
+
+//
+// See the documentation file for a full description of this context's structure:
+// Documentation/SwiftGenKit Contexts/Yaml.md
```
AliSoftware:
```suggestion
// Documentation/SwiftGenKit Contexts/Json.md
```

djbe (changes requested):
Just couple of small things, rest seems good (assuming it's mostly a copy&paste from the yaml parser).

djbe:
Tagging this 6.2.1 for now, unless it's done before we create the 6.2.0 PR.

Sources/SwiftGenKit/Parsers/Yaml/YamlParser.swift
AliSoftware: And make the `Yaml.Parser` now `final` too 👍

marcelofabri: done!

marcelofabri:
> Finally, split the YAML-Demo page in the demo Playground in 2 separate pages, extracting the JSON example in its own page

I didn't quite understand the playground. Is that embedded code generated by SwiftGen automatically?

djbe:
Currently we copy code from one of the generated files (for the output tests), and sometimes process them a bit as needed (to keep the pages as minimal as possible)

marcelofabri:
> Currently we copy code from one of the generated files (for the output tests), and sometimes process them a bit as needed (to keep the pages as minimal as possible)

hmm, do we need to make any changes here then? since the generated code will be the same, I don't see why we need to update the playgrounds

djbe:
If the test changes that @AliSoftware mentioned are implemented, then that'll mean that the resources and generated code **will** change (and we'll have a separate JSON page instead of 1 for both JSON & YAML). We'll get to that when we've covered the test changes.

marcelofabri:
> If the test changes that @AliSoftware mentioned are implemented, then that'll mean that the resources and generated code **will** change (and we'll have a separate JSON page instead of 1 for both JSON & YAML). We'll get to that when we've covered the test changes.

i've updated the tests, but the thing is that the template tests still are using ymls. I don't think this is too much of a deal right now, so I wonder if we can do this clean up later (as we update the templates to use `document`, for example)

Sources/SwiftGenKit/Stencil/JSONContext.swift
```diff
@@ -0,0 +1,41 @@
...
+      .map(map(file:))
+
+    return [
+      "files": files
+    ]
+  }
+
+  private func map(file: JSON.File) -> [String: Any] {
+    // `documents` is an array for compability reasons with YAML and we'll be updated to be a single document in a
+    // future release
```
AliSoftware: s/we'll/will/

Though that comment will likely need to change if you implement @djbe 's suggestion of having both `"document"` (new key introduced even before 7.0 so people can start migrate) and `"documents"` (compatibility)

```suggestion
    // `documents` is an array for compability reasons with YAML
    // and will be updated to be a single document in a future release
```

AliSoftware:
Ok, for the playground that's my bad, I missed that we already had a `JSON-Demo` page it seems. So no new page to create there and probably nothing much to change.

I think I confused this with the case of some playground pages in which we _do_ have multiple examples in a single playground page (for example the `Plist-Demo` page… hell even the `JSON-Demo` page… where we have 2 examples in these pages, one to show the code for `inline-swift4` templates and the other for `runtime-swift4`. For some reason that memory of this made me think for a minute that we had a Yaml-Demo page with 2 examples too, one for Yaml one for JSON… but that's not the case (and wouldn't make sense since indeed the outputs are the same for both anyway), and we already have the Json and Yaml demos separate in playground


The only thing you could include in that PR regarding playground changes is to update the examples – for example the `Yaml-Demo` page contains `internal enum Configuration` in its examples while after your change the generated code in tests (and the file in `Fixtures/Generated/JSON/…`) doesn't contain that anymore (since that file has move to JSON fixtures and is not used in Yaml tests. Technically, the playground page is not incorrect, but it could be confusing to someone looking both at the test generated fixtures and the playground 🤷  not a bit deal though

Tests/Fixtures/Generated/JSON/inline-swift4/all-customName.swift
```diff
@@ -10,12 +10,6 @@ import Foundation
 
 // swiftlint:disable identifier_name line_length number_separator type_body_length
 internal enum CustomJSON {
-  internal enum Configuration {
-    internal static let apiVersion: String = "2"
-    internal static let country: Any? = nil
-    internal static let environment: String = "staging"
-    internal static let options: [String: Any] = ["screen-order": ["1", "2", "3"]]
-  }
   internal enum Documents {
```
AliSoftware: I see that this Generated output is containing values from the YAML contexts: `enum Documents`, `enum GroceryList`, `enum Mapping`… those are the result of parsing `documents.yml`, `grocery-list.yml`, `mappings.yml` input files that are located in `Fixtures/StencilContexts/Yaml`

Instead for consistency's sake we should use the `JSON` input files and especially the `StencilContexts/Json/*` they generate as input for the `JSON` tests.

This discrepency there is likely due to `Tests/TemplatesTests/JsonTests.swift` which needs to be updated to change every `resourceDirectory:` parameter used in `test(…)` calls to be `.json` and not `.yaml` anymore

AliSoftware (commented):
Latest changes seems to address everything, except that comment below about the tiny change left in `TemplatesTests/`; once this is addressed I think we'll be ready to merge! 🎉

Tests/Fixtures/Generated/JSON/inline-swift5/compilation-configuration.yml
```diff
@@ -1,4 +0,0 @@
---- !ruby/object:CompilationConfiguration
-common:
-  swift_versions:
-    - 5
```
AliSoftware: This file should not have been deleted: it's not a fixture per se, but instead used by our `Rakefile` task `rake output:compile` – which is in charge of compiling the generated outputs to ensure that they are producing valid Swift that is able to compile

Tests/Fixtures/Generated/JSON/inline-swift4/compilation-configuration.yml
```diff
@@ -1,5 +0,0 @@
---- !ruby/object:CompilationConfiguration
-common:
-  swift_versions:
-    - 4
-    - 4.2
```
AliSoftware: ditto

Tests/Fixtures/Generated/JSON/runtime-swift4/compilation-configuration.yml
```diff
@@ -1,5 +0,0 @@
---- !ruby/object:CompilationConfiguration
-common:
-  swift_versions:
-    - 4
-    - 4.2
```
AliSoftware: same

Tests/Fixtures/Generated/JSON/runtime-swift5/compilation-configuration.yml
```diff
@@ -1,4 +0,0 @@
---- !ruby/object:CompilationConfiguration
-common:
-  swift_versions:
-    - 5
```
AliSoftware: likewise

Tests/Fixtures/Generated/JSON/inline-swift5/compilation-configuration.yml
marcelofabri: oops

AliSoftware (approved):

marcelofabri:
Looks like the tests are failing because the template is not stable anymore, the order of the keys is changing between runs:

```diff
 internal enum JSONFiles {
-  internal static let items: [[String: Any]] = [[AnyHashable("id"): "1", AnyHashable("name"): "Anna"], [AnyHashable("name"): "Bob", AnyHashable("id"): "2"]]
+  internal static let items: [[String: Any]] = [[AnyHashable("name"): "Anna", AnyHashable("id"): "1"], [AnyHashable("name"): "Bob", AnyHashable("id"): "2"]]
 }
```

This is not surprising given that `Dictionary` doesn't have stable keys. I wonder why this wasn't a problem before (and if there's an easy way to fix it).

AliSoftware:
Ah, this has indeed been a problem in other places in the past.

This is why we order most of the content alphabetically in most of the places. Like the order of the entries in `.strings` is sorted before being output (see [StringsContext.swift](https://github.com/SwiftGen/SwiftGen/blob/master/Sources/SwiftGenKit/Stencil/StringsContext.swift#L69))

We might need to do the same for the JSON content then. Indeed not sure why this was not a problem before.

Maybe it was sheer luck that the order on macOS (our dev machines) and Linux (CI) were the same in the past implementation. Maybe it was due to some implementation details of `Yams` when parsing the file.

In any case the fact is that `JSONSerialization` is not deterministic in the keys orders (as expected for a dictionary) so we'll likely need to add some sorting indeed when building the context. The bad news is that we'll need to do that sorting on embedded dicts recursively too, to ensure consistency at all levels of the JSON structure being parsed… (we might want to add just another fixture that contains some deeper JSON structure to test that part btw)

AliSoftware:
_PS: in case you were thinking of using the special build setting that we can use in Xcode to make Dictionary hashes be stable across runs for testing purposes, this is not really a solution since we also need to generate stable outputs in production, not just during tests, so that the generated code is also stable when the user runs `swiftgen` on every build via a script build phase_

djbe:
Note on that "sorting of keys" issue:
This was actually adressed in Stencil itself (https://github.com/stencilproject/Stencil/pull/242), it'll sort dictionaries before iterating if it detects a `[String: Any]`.

What has changed that defeats that check? And is there something that can be adressed in Stencil? I'm preparing a release there, so we can include a fix if needed.

marcelofabri:
> Note on that "sorting of keys" issue:
> This was actually adressed in Stencil itself ([stencilproject/Stencil#242](https://github.com/stencilproject/Stencil/pull/242)), it'll sort dictionaries before iterating if it detects a `[String: Any]`.
> 
> What has changed that defeats that check? And is there something that can be adressed in Stencil? I'm preparing a release there, so we can include a fix if needed.

so I think the issue is not the order in the template, but the order inside a template variable

djbe:
The template outputs variables using a for loop:
https://github.com/SwiftGen/SwiftGen/blob/develop/templates/json/inline-swift5.stencil#L63

marcelofabri:
> The template outputs variables using a for loop:
> https://github.com/SwiftGen/SwiftGen/blob/develop/templates/json/inline-swift5.stencil#L63

thanks for the link! i thought for some reason that this was handled at the context level.

I have no clue why this is happening then :(

marcelofabri:
Maybe the issue is that those keys are treated as `AnyHashable` for some reason so it's not getting into that codepath in Stencil (as it casts to `[String: Any]`)?

https://github.com/stencilproject/Stencil/pull/242/files#diff-697be680572c247eef6a463a952c9e8bL95

djbe:
I wonder why it worked before, then? Clearly the old template output was also `AnyHashable`... Gonna need to debug this 😞 (it's a tricky thing to debug, I remember this from the original issue)

marcelofabri:
> I wonder why it worked before, then? Clearly the old template output was also `AnyHashable`... Gonna need to debug this 😞 (it's a tricky thing to debug, I remember this from the original issue)

It's not that clear to me - I don't think we had any tests covering a `Dictionary` inside an `Array` or another `Dictionary`. Thanks for the help investigating it!

AliSoftware:
Yeah I was thinking either that `AnyHashable`, or — because of `NSJSONSerialisation`'s ObjC legacy, maybe it's returning an `NSDictionary` or a `[NSString: NSObject]` which for some reason is maybe not fully bridged to `[String: Any]` at least maybe not recursively, making the cast in Stencil fail?

marcelofabri:
I took another look today and looks like this has always been a problem. The test that trigger this goes through this line in the template: https://github.com/SwiftGen/SwiftGen/blob/master/templates/json/inline-swift4.stencil#L28, which relies on `.description` to provide the representation.

~I'll try to repro this in the `develop` branch now. If it happens there, I propose we just file a new issue and update the unit tests to not hit this bug 😬~

I was able to repro it with the YML parser, so this is really a template bug. I've filled https://github.com/SwiftGen/SwiftGen/issues/712 and IMO we should decouple that with this PR since it's an unrelated template bug.

Tests/SwiftGenKitTests/JSONTests.swift
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  func testArray() throws {
+    let parser = try JSON.Parser()
+    try parser.searchAndParse(path: Fixtures.path(for: "array.json", sub: .json))
+
+    let result = parser.stencilContext()
+    XCTDiffContexts(result, expected: "array", sub: .json)
+  }
+
```
djbe, line 34: We're missing the `testDirectoryInput` here (same as in YAML tests), which produces a `all.yaml` context.

Tests/TemplatesTests/JsonTests.swift
```diff
@@ -9,12 +9,12 @@ import XCTest
 
 final class JsonTests: XCTestCase {
   private enum Contexts {
-    static let all = ["empty", "all"]
+    static let all = ["empty", "array", "configuration"]
```
djbe: This should be changed back once we have the `all` context again (see comment above).

djbe (changes requested):

djbe (changes requested):
This also needs to update the docs at `Documentation/SwiftGenKit Contexts/JSON.md`. Note that this currently is a symlink to the `Yaml.md` file

CHANGELOG.md
```diff
@@ -164,6 +164,9 @@ _None_
 * Updated Pods and Gems dependencies.  
   [David Jennes](https://github.com/djbe)
   [#684](https://github.com/SwiftGen/SwiftGen/pull/684)
+* Use `JSONSerialization` instead of `Yams` for parsing JSONs, making the `json` parser faster.
```
djbe: This entry needs to move to the new dev release notes.

Tests/TemplatesTests/JsonTests.swift
```diff
@@ -9,7 +9,7 @@ import XCTest
 
 final class JsonTests: XCTestCase {
   private enum Contexts {
-    static let all = ["empty", "all"]
+    static let all = ["empty", "array", "configuration", "all"]
```
djbe: To avoid churn of output files in future changes, we can just test `empty` and `all`. Note: don't forget to delete the corresponding generated output files for `array` and `configuration`.

SwiftGen.playground/Pages/JSON-Demo.xcplaygroundpage/Contents.swift
```diff
@@ -31,12 +31,6 @@ import Foundation
 
 // swiftlint:disable identifier_name line_length number_separator type_body_length
 internal enum JSONFilesInline {
-  internal enum Configuration {
```
djbe, line 34: I'm not sure this is correct? Also we need to add the right resources to playgrounds, @marcelofabri if you want I can do that part.

djbe (changes requested):

djbe (changes requested):
GitHub is being difficult, so can't point to the right line, but:
In the json context documentation, you can remove the `documents` part (plural), we don't usually document deprecated properties in the Docs.

Sources/SwiftGenKit/Stencil/JSONContext.swift
djbe: Adressed in #729

Sources/SwiftGenKit/Parsers/JSON/JSONParser.swift
djbe: Adressed in #729

SwiftGen.playground/Pages/JSON-Demo.xcplaygroundpage/Contents.swift
marcelofabri: it would be great if you could take over the playground piece 🙏

djbe (approved):
Allright this looks pretty complete to me. Maybe @AliSoftware wants to do a last pass, otherwise we can merge this. The (functional) change itself is pretty small after all.

AliSoftware:
idk what's up with GitHub this evening. Can't post a PR review nor comment inline either.

#### `Documentation/SwiftGenKit Contexts/Json.md`, line 5
typo: `s/mutiple/multiple/` + wording `s/each file/each file's content/`

```diff
-The JSON parser accepts mutiple files or directories (which it'll recursively search). Each file will be loaded into the context, but the parser will also generate metadata about the structure of the file.
+The JSON parser accepts multiple files or directories (which it'll recursively search). Each file's content will be loaded into the context; but the parser will also generate metadata about the structure of the file content.
```

This diff actually needs to be applied to `Documentation/SwiftGenKit Contexts/Yaml.md` as well

#### `Documentation/SwiftGenKit Contexts/Json.md`, line 38

I know @djbe said in an earlier feedback that we don't document deprecated keys in contexts, but then it looks strange to have that `documents:` key in the example while it not being documented in the previous section…

imho we should either document that `documents:` key and mention it's deprecated, or omit it from the example (even if that makes the example not completely true to what is actually generated because of retro-compatibility until SwiftGen 7) so that it's less confusing.

#### `Tests/TemplatesTests/JsonTests.swift`

nit: It seems like `Contexts.all` only refers to `["empty", "all"]` and doesn't reference `configuration` and `array` contexts. That might be fine (maybe we don't want to test all the cases everywhere and pollute the tests too much) but then we might consider renaming that variable to something other than `all`? `commonTestSet` or something?

AliSoftware:
btw wondering if we should merge this in `develop` as soon as we officially approve it, or if we should wait to have a fix for #712 before we merge that transition to `JSONSerialization`. I know the issue with #712 was already present before that PR though, but wondering if that already-existing issue will start to become way more common and thus more problematic if we merge that migration to `JSONSerialization` 🤔

djbe:
There are issues with the inline templates in general:
- unstable output #712
- strings issues (multiline, quotes, ...) #551

Seeing as those are already present in the current & older releases, I don't think this PR should be held back because of them.

AliSoftware:
Ok, updated [my previous comment](https://github.com/SwiftGen/SwiftGen/pull/702#issuecomment-648520981) with my latest feedback. Mainly just a typo and a question about including `documents` key or not in docs example

djbe:
I'm merging everything and processing the feedback (+ fixing the typos in other places).

Note: @AliSoftware we already use the `all` context in other parsers, if we rename it here, we should rename it everywhere. And TBH we can delegate that to another PR.

AliSoftware:
> I'm merging everything and processing the feedback (+ fixing the typos in other places).

👍  cool. I see that you fixed `s/mutiple/multiple/` in those other places, but did not apply the same `s/each file/each file's content/` fix everywhere tho 😅 
 
> Note: @AliSoftware we already use the `all` context in other parsers, if we rename it here, we should rename it everywhere. And TBH we can delegate that to another PR.

Ah, right. Let's keep `Context.all` then 👍 

just another nit: I've also noticed that we seem to use inconsistent case in fine names for `Documentation/SwiftGenKit Parsers` vs `Documentation/Parsers` (`Json.md` vs `json.md` for example). _(sorry, I'm quite nitpicky lately, that's just my perfectionist side speaking… hope you don't take it the wrong way 😬)_

AliSoftware (approved):

PR #708, opened 2020-05-28
Run parsers in parallel

### LINKED ISSUE #699: Run commands in parallel
If I understood correctly, this is where SwiftGen starts the process of running commands from configs:

https://github.com/SwiftGen/SwiftGen/blob/1cf0a8fb410b82f30d5615e0e19e0a7b9b033b22/Sources/SwiftGen/Commander/ConfigCommands.swift#L117-L120

Could this be done in parallel threads as an easy way to speed up SwiftGen with multiple commands? Any gotchas if we decide to implement it?

marcelofabri:
Fixes #699 

Putting it here to gather some early feedback


Sources/SwiftGen/Commander/ConfigCommands.swift
```diff
@@ -162,3 +160,20 @@ enum ConfigCLI {
...
+    let errors = commands.keys.sorted().parallelCompactMap { cmd -> Swift.Error? in
+      do {
+        _ = try run(cmd: cmd, verbose: verbose)
+        return nil
+      } catch {
+        return error
+      }
+    }
+
+    if let error = errors.first {
```
marcelofabri: we could create an error type to combine multiple errors into one, put i thought this was a good enough solution for now (since in the existing implementation the user will only get the first error that they've faced)

SwiftGen-Eve:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

---

</p>

Sources/SwiftGen/Commander/ConfigCommands.swift
AliSoftware: Ah!, I read your code from Fork.app (as when I did a `git fetch` I saw that new branch and was curious) before seeing that draft PR, and thought exactly of this. Great minds ;)

I think it would be easy enough though to add an `indirect case composite([Config.Error])` to the `enum Config.Error`, with its `description`  being:
 `case .composite(let errors): return errors.map({ " - \($0.description)" }).joined(separator: "\n")` for example.

Sources/SwiftGen/Array+Parallel.swift
```diff
@@ -0,0 +1,29 @@
...
+    parallelMap(transform: transform).flatMap { $0 }
+  }
+
+  func parallelCompactMap<T>(transform: (Element) -> T?) -> [T] {
+    parallelMap(transform: transform).compactMap { $0 }
+  }
+
+  func parallelMap<T>(transform: (Element) -> T) -> [T] {
+    var result = ContiguousArray<T?>(repeating: nil, count: count)
+    return result.withUnsafeMutableBufferPointer { buffer in
```
AliSoftware, line 22: nit: I'm curious as to why using an UMBP here as opposed to accessing the `ContiguousArray` directly? Is it for performance reasons? Or for thread-safety reasons and avoid risk of concurrent access to the `ContinuousArray`?
I'm guessing the latter, but would be nice to add a code comment

Sources/SwiftGen/Commander/ConfigCommands.swift
```diff
@@ -162,3 +160,20 @@ enum ConfigCLI {
     NSWorkspace.shared.open(docURL)
   }
 }
+
+private extension Config {
+  func runCommands(verbose: Bool) throws {
+    let errors = commands.keys.sorted().parallelCompactMap { cmd -> Swift.Error? in
+      do {
+        _ = try run(cmd: cmd, verbose: verbose)
```
AliSoftware: Why the need for `_ =`? Doesn't `run(cmd:verbose:)` return `Void`?

Sources/SwiftGen/Logs.swift
```diff
@@ -39,17 +39,37 @@ enum ANSIColor: UInt8, CustomStringConvertible {
...
+  let queue = DispatchQueue(
+    label: "swiftgen.outputQueue",
+    qos: .userInteractive,
+    target: .global(qos: .userInteractive)
+  )
+
+  #if !os(Linux)
+  atexit_b {
+    queue.sync(flags: .barrier) {}
+  }
```
AliSoftware, line 68: nit: I understand what this does (ensuring every pending log message gets dequeued before we really exit) but given that `atexit_b` is not a common function known to every devs, a code comment could be welcome

Sources/SwiftGen/Logs.swift
```diff
@@ -39,17 +39,37 @@ enum ANSIColor: UInt8, CustomStringConvertible {
...
-  case .error:
-    fputs(ANSIColor.red.format("swiftgen: error: \(string)\n"), stderr)
+  outputQueue.async {
+    switch level {
+    case .info:
+      fputs(ANSIColor.green.format("\(string)\n"), stdout)
+    case .warning:
+      fputs(ANSIColor.yellow.format("swiftgen: warning: \(string)\n"), stderr)
+    case .error:
+      fputs(ANSIColor.red.format("swiftgen: error: \(string)\n"), stderr)
```
AliSoftware, line 52: I know it's not directly related to the immediate scope of this PR, but could be the occasion to quick-fix https://github.com/SwiftGen/SwiftGen/issues/547 here 🤔

AliSoftware (commented):
Looks promising 👍

Sources/SwiftGen/Commander/ConfigCommands.swift
djbe: Yeah, that shouldn't be needed afaik, the signature is:
```swift
func run(cmd: String, verbose: Bool) throws
```

marcelofabri: i was using `try?` initially, that's why it was there 😅

Sources/SwiftGen/Logs.swift
marcelofabri: i'd rather leave that for another PR since I don't have any context about that issue

djbe: Agreed, no need to mix that in here.

Sources/SwiftGen/Array+Parallel.swift
marcelofabri: yeah, I think it's the latter, but i'm not super sure tbh 😬

Sources/SwiftGen/Array+Parallel.swift
```diff
@@ -0,0 +1,29 @@
...
+      DispatchQueue.concurrentPerform(iterations: buffer.count) { idx in
+        buffer[idx] = transform(self[idx])
+      }
+      return buffer.compactMap { $0 }
```
djbe, line 26: This is different to the SwiftLint code it's based on, which uses:
```swift
return buffer.map { $0! }
```

The result could be different? (not 100% sure)

marcelofabri: It shouldn’t (i changed it to avoid the force unwrap which triggers a SwiftLint violation)

djbe:
So I was bored and parallelized entries as well 😅 Just based it on the existing parallelization code.

**Edit**: refactored it a bit to only parallelize 1 array (after flattening commands & entries into 1), as to avoid dispatch queue bonanza.

Sources/SwiftGen/Commander/ConfigCommands.swift
```diff
@@ -153,7 +131,58 @@ enum ConfigCLI {
...
+    }
+
+    try entry.checkPaths()
+    try entry.run(parserCommand: cmd)
+  }
+
+  /// Flatten all commands and their corresponding entries into 1 list
+  func collectCommandsAndEntries() throws -> [(ParserCLI, ConfigEntry)] {
+    try commands.keys.sorted()
+      .compactMap { cmd in
```
AliSoftware: Why not just `map`?
If I'm not mistaken, that closure takes a string and returns a (non-optional) Parser, right?
(And if a parser can't be found it `throws`, exiting from the whole method directly not just the single iteration on that particular cmd anyway)

Sources/SwiftGen/Logs.swift
```diff
@@ -39,17 +39,38 @@ enum ANSIColor: UInt8, CustomStringConvertible {
...
+    }
   }
 }
 
+private let outputQueue: DispatchQueue = {
```
AliSoftware: nit: I'd call it `logQueue` instead (and adapt the label accordingly too), as "output" in the context of SwiftGen can be misleading and make me thing about the code we generate, not what we print to stdout

Sources/SwiftGen/Commander/ConfigCommands.swift
```diff
@@ -153,7 +131,58 @@ enum ConfigCLI {
...
+        guard let parserCmd = ParserCLI.command(named: cmd) else {
+          throw Config.Error.missingEntry(key: cmd)
```
AliSoftware: I know that you didn't change that part and that's the error we already used before you moved the code, but I'm wondering if that's the right one to throw? If a user is referring to a parser name that doesn't actually exist, it's not really that we're missing an entry for that key but rather that the key is invalid, right?

CHANGELOG.md
```diff
@@ -18,7 +18,9 @@ _None_
 
 ### Internal Changes
 
-_None_
+* Commands are now executed in parallel, making SwiftGen faster when multiple parsers are used.
```
AliSoftware: s/Commands/Parsers/? (For consistency in our vocabulary throughout the documentation)

Sources/SwiftGen/Commander/ConfigCommands.swift
djbe: Probably a left-over from getting type errors for the 2nd closure.

Sources/SwiftGen/Config/Config.swift
```diff
@@ -200,17 +200,23 @@ extension Config {
...
       case .missingEntry(let key):
         return "Missing entry for key \(key)."
-      case .wrongType(let key, let expected, let got):
-        return "Wrong type for key \(key ?? "root"): expected \(expected), got \(got)."
+      case .multipleErrors(let errors):
+        return errors.map { $0.localizedDescription }.joined(separator: "\n")
       case .pathNotFound(let path):
         return "File \(path) not found."
+      case .unknownAction(let name):
+        return "Action `\(name)` does not exist."
```
AliSoftware: s/Action/Parser/?

Sources/SwiftGen/Logs.swift
```diff
@@ -39,17 +39,38 @@ enum ANSIColor: UInt8, CustomStringConvertible {
...
+      fputs(ANSIColor.yellow.format("swiftgen: warning: \(string)\n"), stderr)
+    case .error:
+      fputs(ANSIColor.red.format("swiftgen: error: \(string)\n"), stderr)
+    }
   }
 }
 
+private let logQueue: DispatchQueue = {
+  let queue = DispatchQueue(
+    label: "swiftgen.outputQueue",
```
AliSoftware:
```suggestion
    label: "swiftgen.log.queue",
```

AliSoftware (commented):
nits

Sources/SwiftGen/Config/Config.swift
djbe: Actually did that on purpose, as `config lint` also outputs "action" instead of parser.

AliSoftware: Right. Shouldn't we try to uniformise tho to avoid losing the users by using different terms? Not sure what's best 🤷🏻

AliSoftware (commented):
Shouldn't we add tests for this?

Testing that they are run in parallel and that there's no race conditions would be tricky if ever possible so probably not worth the effort, but testing that multiple errors properly end up generating a `multipleErrors` and properly prints _all_ the errors encountered in the log output, I think that would still be easily feasible and beneficial

Sources/SwiftGen/Config/Config+Run.swift
```diff
@@ -10,12 +10,12 @@ import StencilSwiftKit
 import SwiftGenKit
 
 extension Config {
-  func runActions(verbose: Bool) throws {
+  func runActions(verbose: Bool, logger: @escaping (LogLevel, String) -> Void = logMessage) throws {
```
AliSoftware: I'm not sure what makes it required to be `@escaping`? I mean yeah it's passed around to the next helper function, but none of the helper/private functions we pass that closure to ends up _capturing_ / _escaping_ the closure, they use it right away, right?

Or is it because of `parallelCompactMap` which implicitly captures it? (if so, I think we could work around it either by using a local variable outside the `parallelCompactMap` referencing the `logger` parameter, or by using `withoutActuallyEscaping` 🤔

Sources/SwiftGen/Config/Config+Run.swift
```diff
@@ -29,18 +29,23 @@ extension Config {
...
-  private func run(cmd: ParserCLI, entry: ConfigEntry, verbose: Bool) throws {
+  private func run(
+    cmd: ParserCLI,
+    entry: ConfigEntry,
+    verbose: Bool,
+    logger: @escaping (LogLevel, String) -> Void
+  ) throws {
     var entry = entry
 
     entry.makingRelativeTo(inputDir: inputDir, outputDir: outputDir)
```
AliSoftware: _(note for separate PR: we should rename that method to `makeRelativeTo` or something, that current naming feels odd)_

Tests/SwiftGenTests/ConfigRunTests.swift
```diff
@@ -10,33 +10,88 @@ import Yams
...
-
-    try configFile.parent().chdir {
-      try config.runActions(verbose: true)
+    let logQueue = DispatchQueue(label: "swiftgen.log.queue")
+    let logger = { (level: LogLevel, msg: String) -> Void in
+      logQueue.async {
+        if let idx = missingLogs.firstIndex(where: { $0 == level && $1 == msg }) {
+          missingLogs.remove(at: idx)
+        } else if unwantedLevels.contains(level) {
+          XCTFail("Unexpected log: \(msg)")
```
AliSoftware:
```suggestion
          XCTFail("Unexpected log: \(msg)", file: file, line: line)
```

Tests/SwiftGenTests/ConfigRunTests.swift
```diff
@@ -10,33 +10,88 @@ import Yams
...
+        }
+      }
+    }
+    let errorHandler = { (error: Error) -> Void in
+      if let idx = missingLogs.firstIndex(where: { $0 == .error && $1 == String(describing: error) }) {
+        missingLogs.remove(at: idx)
+      } else {
+        XCTFail("Unexpected error: \(error)", file: file, line: line)
+      }
     }
```
AliSoftware: nit: wondering if we could extract the `logger` and `errorHandler` closures into private methods to make that `_testRun` method shorter and more readable? Though they capture `missingLogs` (r/w) and `unwantedLevels` (readonly) variables so might not be as easy – except if we pass the `missingLogs` as `inout` 🤔 

Even if not extracting them out of the `_testRun` function, making them `func` instead of closures might make it more readable when human-parsing the code structure? 🤷 


Also, it could be a good occasion to rename that `_testRun` private func to something like `runTest()`?

AliSoftware (commented):
Looks good overall, just a question about if we could get rid of `@escaping` and a nit or two, otherwise 👌

Tests/SwiftGenTests/ConfigRunTests.swift
djbe: I'll look into it, just based myself on the existing config lint test code 🤷

Sources/SwiftGen/Config/Config+Run.swift
djbe: Yeah it's because of `parallelCompactMap`. Didn't think about those options, will see if I can get those to work.

Sources/SwiftGen/Config/Config.swift
```diff
@@ -57,11 +57,13 @@ extension Config {
     guard let config = yaml as? [String: Any] else {
       throw Config.Error.wrongType(key: nil, expected: "Dictionary", got: type(of: yaml))
     }
-    self.inputDir = (config[Keys.inputDir] as? String).map { Path($0) }
-    self.outputDir = (config[Keys.outputDir] as? String).map { Path($0) }
+    self.inputDir = (config[Keys.inputDir.rawValue] as? String).map { Path($0) }
+    self.outputDir = (config[Keys.outputDir.rawValue] as? String).map { Path($0) }
+
     var cmds: [String: [ConfigEntry]] = [:]
```
AliSoftware: Just realised: it might not be useful to parse the Config commands as a `[String: [ConfigEntry]]` anymore if it's only to later transform that representation into a `[(String, ConfigEntry)]` or `[(Parser, ConfigEntry)]` list in `collectCommandsAndEntries()`

Do we use the dictionary (one parser -> list of entries for that parser) structuration in any way other than here? If not, we might as well generate the flatten list right away and get rid of `collectCommandsAndEntries` converting it later 😉

Tests/SwiftGenTests/ConfigRunTests.swift
```diff
@@ -23,16 +23,18 @@ class ConfigRunTests: XCTestCase {
     var missingLogs = expectedLogs
     let configFile = Path(path)
     let logQueue = DispatchQueue(label: "swiftgen.log.queue")
-    let logger = { (level: LogLevel, msg: String) -> Void in
+
+    func log(level: LogLevel, msg: String) {
```
AliSoftware: 👍

djbe:
Those last 2 commits don't **really** belong in this PR, but it's such a small change to internal stuff that I can't be bothered to open a separate PR.

Sources/SwiftGen/Config/Config.swift
```diff
@@ -121,26 +130,25 @@ extension Config {
       logger(.error, "output_dir: Output directory \(Message.doesntExist(outputDir))")
     }
 
-    for (cmd, entries) in commands {
-      if let replacement = Config.deprecatedCommands[cmd] {
-        logger(.warning, Message.deprecatedAction(cmd, for: replacement))
+    let groupedCommands: [ParserCLI: [ConfigEntry]] = Dictionary(grouping: commands) { $0.command }
+      .mapValues { $0.map { $0.entry } }
```
AliSoftware, line 134: Is the re-grouping only used so that we log "N entries for command X"? (Which we should s/command/parser/ btw)? If so, I'm not sure it's worth it tbh; we can just change the logs to print one line per config entry even if there are multiple entries using the same parser and get rid of the "N entries" count in log message if that's the only reason for Dictionary(grouping:…) to be there.

AliSoftware: Ah wait that's just for when running `config lint` (I thought this was for `config run --verbose`, missed the context in Mobile GitHub and focused snippets review) this is fine then.

Sources/SwiftGen/Config/Config+Run.swift
```diff
@@ -0,0 +1,98 @@
...
+
+    if errors.count == 1 {
+      throw errors[0]
+    } else if errors.count > 1 {
+      throw Error.multipleErrors(errors)
+    }
+  }
+
+  private func run(
+    cmd: ParserCLI,
```
AliSoftware: nit: s/cmd:/parserCommand:/ ? 🤔

AliSoftware (approved):
💙

Sources/SwiftGen/Config/Config+Run.swift
djbe: Dunno 😅 The terminology keeps being ~difficult~ tricky:
- SGK has parsers, that's clear
- From CLI we invoke commands
- `ParserCLI` are an abstraction of a command to invoke a parser
- `ParserCLI.command()` converts that abstraction to an actual command
- Probably some other stuff I'm missing.

AliSoftware: Yeah I agree. Naming and terminology is hard.

I want us to be consistent and use the same term everywhere for the same thing to avoid confusion in code and docs, but ParserCLI is indeed not really a parser and more like a command telling what parser to run but also with which in/out parameters, so `command` terminology is probably right here…

djbe:
@AliSoftware do we feel confident enough about shipping this in 6.2.1, or delay it to 6.3.0?

AliSoftware:
Technically (semver speaking) since it's a new feature we should make it bump to 6.3.0

We could consider skipping 6.2.1 and directly do a 6.3.0 with both the bug fixes we already have ready _and_ this PR though

djbe:
Technically, it's not a user visible feature though, users have no control over it. I purposefully moved the changelog entry to "internal changes".

I'd like to just push out 6.2.1 with bug fixes (and potentially this internal feature), and have a clearer 6.3 release with new features such as strings plurals.

AliSoftware:
Fair enough, let's push it then 🙌

marcelofabri:
Thanks for bringing this to the finish line, @djbe 🚀

Issue #700, opened 2020-05-26
Optionally skip metadata generation for JSON/Yaml parsers


marcelofabri:
When profiling parsing JSONs (Lottie animations), I've noticed that most of the time is spent generating the `metadata` field, although my template doesn't use it at all.

By commenting [this line](https://github.com/SwiftGen/SwiftGen/blob/develop/Sources/SwiftGenKit/Stencil/YamlContext.swift#L36), I was able to improve the duration from ~750ms (using `JSONSerialization`, see #698) to ~100ms.

Since figuring out whether we need to parse it or not based on the template would be tricky, I propose that we add a key to `ConfigEntryOutput` named `includeMetadata` that will be `true` by default.

The configuration file would look like this:

```yml
json:
  inputs:
    - Assets/Lottie/
  outputs:
    - templatePath: templates/lottie.stencil
      output: Generated/Lottie.swift
      includeMetadata: false # or should it be `skipMetadata` instead?
```

marcelofabri:
I guess we should use `ParserOptionValues` instead,

AliSoftware:
Yep, that's typically exactly a use case for which `ParserOptionValues` was introduced. Would be a great addition as a parser option indeed!

You're full of great ideas man! ❤️

djbe:
There may be a possibility of calculating the metadata lazily, ping ponging some ideas on Slack 😆

AliSoftware:
Yeah, the "lazy" idea would be to leverage the fact that Stencil is not limited to take NSDictionaries/NSArrays as context but can also work with objects. (I think it uses KVC on NSObjects found inside the context and `Mirror` on non-`@objc`, plain Swift types?)

So the idea would be to make the entry for each _document_ not be a `[String: Any]` anymore with keys `data` and `metadata`, but to use an actual object for that, with a `lazy var`, basically turning this:
```swift
  private func map(document: Any) -> [String: Any] {
    [
      "data": document,
      "metadata": Metadata.generate(for: document)
    ]
  }
```
into something like this:
```swift
struct Document {
  let data: Any
  lazy var metadata: [String: Any] = { Metadata.generate(for: self.data) }()
}
```
The hope being that the metadata property would only be constructed if accessed from the template. Which means that we would not even need a parser option (`ParserOptionsValue`) after all!

We've never used that feature of exposing classes to Stencil in the past so this might require experimenting a bit with it to test that this idea above would actually work, but it's definitively worth a try.

[...]

PR #709, opened 2020-05-29
Lazily evaluate `metadata` in JSON/Plist/YAML parsers

marcelofabri:
Fixes https://github.com/SwiftGen/SwiftGen/issues/700


SwiftGen-Eve:

---

</p>

djbe:
@marcelofabri I've applied your changes to the other 2 data parsers, and also tried making the test helpers a bit more dynamic.

Sources/SwiftGenKit/Stencil/ContextHelpers.swift
```diff
@@ -0,0 +1,21 @@
...
+// MIT Licence
+//
+
+import Foundation
+
+// MARK: - Lazy evaluation
+
+// We need to use a `NSObject` subclass + `@objc` to make Stencil use KVC to access this property,
+// and thus allow that `lazy var` not to be resolved by Stencil too early or before it's accessed in the template
+final class StencilContextLazyDocument: NSObject {
```
djbe, line 13: At first I thought having a protocol here, something like `DictionarySerializable` or something like that, with a `toDictionary()` method.

But then the types such as LazyDocument here would have to implement that protocol in SGK, which shouldn't happen. Test details should **not** be in normal source code.

Tests/SwiftGenKitTests/TestsHelper.swift
```diff
@@ -155,6 +155,40 @@ func convertToString(_ value: Any) -> String? {
...
+
+func convertNSObjectToDictionary(_ object: NSObject) -> [String: Any] {
+  var result: [(String, Any)] = []
+  var mirror: Mirror? = Mirror(reflecting: object)
+
+  repeat {
+    // ensure we can property handle lazy variables
+    // we use `object.value(forKey:)` to force load lazy variables
+    result += mirror?.children
+      .compactMap { $0.label?.deletingPrefix("$__lazy_storage_$_") }
```
djbe, line 179: So apparently at runtime when mirroring, labels for lazy vars get changed from `metadata` to `$__lazy_storage_$_metadata`...

If Swift ever changes that naming, we'd have to adapt. Luckily, our tests will catch it, and with ABI stability, we'll be safe for quite a while.

Tests/SwiftGenKitTests/TestsHelper.swift
```diff
@@ -155,6 +155,40 @@ func convertToString(_ value: Any) -> String? {
...
+      .map { ($0, object.value(forKey: $0) as Any) } ?? []
```
djbe, line 180: We can't use a `$0.value` (a Mirror child's value), as that evaluates to `nil` and doesn't trigger lazy evaluation.

Cool trick though if you want to check if `lazy` has been evaluated yet.

Sources/SwiftGenKit/Stencil/JSONParser+Context.swift
```diff
@@ -23,19 +23,12 @@ extension JSON.Parser {
...
-    ]
-  }
-
-  private func map(document: Any) -> [String: Any] {
-    [
-      "data": document,
-      "metadata": Metadata.generate(for: document)
+      "document": StencilContextLazyDocument(data: file.document),
+      // Deprecated: remains for legacy/compatibility reasons; will be removed in 7.0
+      "documents": [StencilContextLazyDocument(data: file.document)]
```
AliSoftware: Not sure if it's really a problem or not, but here we're creating 2 instances of the `StencilContextLazyDocument` class, once for `document` and one for `documents`. Wouldn't it be better to use the same instance in both instead?

Sources/SwiftGenKit/Stencil/YamlParser+Context.swift
```diff
@@ -26,14 +26,7 @@ extension Yaml.Parser {
...
       "path": file.path.string,
-      "documents": file.documents.map(map(document:))
-    ]
-  }
-
-  private func map(document: Any) -> [String: Any] {
-    [
-      "data": document,
-      "metadata": Metadata.generate(for: document)
+      "documents": file.documents.map(StencilContextLazyDocument.init)
```
AliSoftware, line 29: Just had a thought: do you think it would be useful to have a `"document"` key that would alias to `documents.first`? Because sure YAMLs can technically have multiple documents in it as per the YAML specification… but let's be honest, most YAML files only contain one document and many people don't know about the ability for YAML to contain multiple docs, so… 🤔

Tests/SwiftGenKitTests/TestsHelper.swift
```diff
@@ -155,6 +155,40 @@ func convertToString(_ value: Any) -> String? {
   }
 }
 
+func convertToDictionary(_ value: Any) -> [String: Any]? {
```
AliSoftware: nit: `private`?

Tests/SwiftGenKitTests/TestsHelper.swift
```diff
@@ -155,6 +155,40 @@ func convertToString(_ value: Any) -> String? {
...
+  case let value as [String: Any]:
+    return value
+  case let value as NSObject:
+    return convertNSObjectToDictionary(value)
+  default:
+    return nil
+  }
+}
+
+func convertNSObjectToDictionary(_ object: NSObject) -> [String: Any] {
```
AliSoftware: nit: `private`?

Tests/SwiftGenKitTests/TestsHelper.swift
AliSoftware: This should be part of the comment imho

```
// At runtime when mirroring, labels for lazy vars get the "$__lazy_storage_$_" prefix
// so we need to delete the prefix to handle those lazy var cases.
// We use `object.value(forKey:) to force load lazy variables because `someMirrorChild.value`
// evaluates to `nil` for lazy variables and doesn't trigger the lazy var evaluation.
```

Tests/SwiftGenKitTests/TestsHelper.swift
AliSoftware: Shouldn't we try to do recursion here? Like if the object's value is itself a dictionary-like thing, recursively call `convertNSObjectToDictionary` on that too?

I mean, currently the only `NSObject` that we have in our contexts is gonna be this top-level `StencilContextLazyDocument` object so that shouldn't really cause any trouble now, but just wondering for the future.

(Also, maybe that kind of conversion better belongs into `Stencil` itself? 🤔 )

AliSoftware (commented):
Should we add some kind of additional unit tests to validate this lazy evaluation behaviour or do you think the existing ones are enough? (I think they probably are, but just wanted to double-check)

Sources/SwiftGenKit/Stencil/YamlParser+Context.swift
djbe: This is for the template writers, they have to handle input correctly. We're deprecating/removing the multiple documents from plist/json, because it isn't correct there. I'd say adding `document` to yaml would be just as incorrect.

I'd rather have correct contexts that force users to deal with the real/actual structure of things, instead of shortcuts that, in this case, barely impact context/template complexity.

djbe:
@AliSoftware Can't seem to answer your comment in-line about "mirror recursion".

There's no need to recurse ourselves, because this is only used by the tests to compare dictionaries. When it goes "deeper" to compare keys & values, it'll automatically trigger the `convertNSObjectToDictionary` as needed.

In general I'd like to get rid of that whole custom comparison mechanism in favour of something like https://github.com/krzysztofzablocki/Difference, or a different test mechanism where we don't compare contexts to YAML files. But those are topics beyond this PR.

AliSoftware (approved):

PR #710, opened 2020-05-29
Colors: Fix long expression type check warning

### LINKED ISSUE #704: Compile time warning in UIColor.init(rgbaValue:)
This generated function can take a non-trivial amount of time to compile:

https://github.com/SwiftGen/SwiftGen/blob/c6477a6950d313c94c964c6c8ec960e3aeccb8f6/templates/colors/swift4.stencil#L49-L58

This was flagged by `-Xfrontend -warn-long-expression-type-checking=250 -Xfrontend -warn-long-function-bodies=250`. The function was taking around 300ms to compile.

I propose an equivalent implementation that is faster to compile:

```swift
// swiftlint:disable operator_usage_whitespace
internal extension {{colorAlias}} {
  convenience init(rgbaValue: UInt32) {
    let components = rgbaValue.normalizedRGBAComponents
    self.init(red: components[0], green: components[1], blue: components[2], alpha: components[3])	
  }
}
// swiftlint:enable operator_usage_whitespace

fileprivate extension UInt32 {
  var rgbaComponents: [CGFloat] {
    let shifts: [UInt32] = [
      self >> 24, // red
      self >> 16, // green
      self >> 8,  // blue
      self        // alpha
    ]
    return shifts.map {
      CGFloat($0 & 0xff)
    }
  }

  var normalizedRGBAComponents: [CGFloat] {
    rgbaComponents.map { $0 / 255.0 }
  }
}
```

Using a custom template with this implementation resolved the issue for us, but it would be great to get this or an equivalent fix made so we can use the built-in template.

ryanmasondavies:
Fixes #704.


SwiftGen-Eve:

---

</p>

CHANGELOG.md
```diff
@@ -101,6 +101,9 @@
 * Core Data: the generated code was missing `,` (comma) for fetch requests with multiple arguments.  
   [David Jennes](https://github.com/djbe)
   [#692](https://github.com/SwiftGen/SwiftGen/pull/692)
+* Colors: Fix compile time warning when long expression type checking is enabled.  
+  [Ryan Mason-Davies](https://github.com/iotize)
+  [#704](https://github.com/SwiftGen/SwiftGen/issues/704) [#710](https://github.com/SwiftGen/SwiftGen/pull/710)
```
AliSoftware:
```suggestion
  [#704](https://github.com/SwiftGen/SwiftGen/issues/704)
  [#710](https://github.com/SwiftGen/SwiftGen/pull/710)
```

Tests/Fixtures/Generated/Colors/swift4/defaults-customName.swift
```diff
@@ -35,18 +35,32 @@ internal struct XCTColors {
 
 // MARK: - Implementation Details
 
-// swiftlint:disable operator_usage_whitespace colon
+// swiftlint:disable operator_usage_whitespace
```
AliSoftware: I wonder why is `operator_usage_whitespace` still useful? I don't see in the remaining code any violation of that rule, is it? 🤔

Tests/Fixtures/Generated/Colors/swift4/defaults-customName.swift
```diff
@@ -35,18 +35,32 @@ internal struct XCTColors {
...
-    let green: CGFloat = CGFloat((rgbaValue >> UInt32(16)) & UInt32(0xff)) / CGFloat(255.0)
-    let blue:  CGFloat = CGFloat((rgbaValue >> UInt32( 8)) & UInt32(0xff)) / CGFloat(255.0)
-    let alpha: CGFloat = CGFloat((rgbaValue              ) & UInt32(0xff)) / CGFloat(255.0)
+    let components = rgbaValue.normalizedRGBAComponents
+    self.init(red: components[0], green: components[1], blue: components[2], alpha: components[3])
+  }
+}
+// swiftlint:enable operator_usage_whitespace
+
+fileprivate extension UInt32 {
```
AliSoftware: Maybe it's just me and I'm being pedantic, but I don't really like having `extension` on a type as common as `UInt32` to add a feature that is _specific_ to colors and not general to every `UInt32`.
I know that this `extension` here is `fileprivate` so it shouldn't affect the rest of the code, by just by principle, wouldn't it be more logical to have that as a free function instead?

`fileprivate rgbaCompoments(from: UInt32) -> [CGFloat]`

Tests/Fixtures/Generated/Colors/swift4/defaults-customName.swift
```diff
@@ -35,18 +35,32 @@ internal struct XCTColors {
...
+      self >> 8,  // blue
+      self        // alpha
+    ]
+    return shifts.map {
+      CGFloat($0 & 0xff)
+    }
+  }
 
-    self.init(red: red, green: green, blue: blue, alpha: alpha)
+  var normalizedRGBAComponents: [CGFloat] {
```
AliSoftware: ditto
```suggestion
  func normalizedRGBAComponents(from value: UInt32) -> [CGFloat] {
```

AliSoftware (commented):
Looking good!

Just suggested a couple of nits to clean up the code

Tests/Fixtures/Generated/Colors/swift4/defaults-customName.swift
ryanmasondavies: Must just be a relic of the past 🙂 I'll remove it.

ryanmasondavies:
@AliSoftware I agree regarding the extensions on `UInt32`. I've added an `RGBAComponents` type for dealing with the extraction. Could you take another look?

Tests/Fixtures/Generated/Colors/swift4/defaults-customName.swift
```diff
@@ -35,18 +35,35 @@ internal struct XCTColors {
...
-    let red:   CGFloat = CGFloat((rgbaValue >> UInt32(24)) & UInt32(0xff)) / CGFloat(255.0)
-    let green: CGFloat = CGFloat((rgbaValue >> UInt32(16)) & UInt32(0xff)) / CGFloat(255.0)
-    let blue:  CGFloat = CGFloat((rgbaValue >> UInt32( 8)) & UInt32(0xff)) / CGFloat(255.0)
-    let alpha: CGFloat = CGFloat((rgbaValue              ) & UInt32(0xff)) / CGFloat(255.0)
+    let components = RGBAComponents(rgbaValue: rgbaValue).normalized
+    self.init(red: components[0], green: components[1], blue: components[2], alpha: components[3])
+  }
+}
+
+fileprivate struct RGBAComponents {
```
AliSoftware: Seems this generates a SwiftLint violation 🤔 

```
 warning: Private over fileprivate Violation: Prefer `private` over `fileprivate` declarations
```

Which is a common SwiftLint warning/rule… in that case I'm not sure tbh: I _think_ we _do_ indeed want `fileprivate`, but the SwiftLint rule says that we usually always want `private` over `fileprivate` for top level definition? (I have to refresh a bit my knowledge of the difference between the two in that special case to be sure…)

AliSoftware (commented):
I like the change 👍 

Just a thing, that code generates a SwiftLint violation, we need to decide if `private` is fine in that case instead of `fileprivate`, or if not and we really do need `fileprivate`, add a `// swiftlint:disable:next` comment for it.

Tests/Fixtures/Generated/Colors/swift4/defaults-customName.swift
djbe: AFAIK, a `private` extension is visible for other types in the same file, so that'd work here. `fileprivate` annotations would only be needed here for specific functions/properties.

AliSoftware: @djbe That's not an `extension` here though, that's the declaration of a new `struct` 😉 
So I think making it `private struct` would still expose that `struct` to any other files in the same module, right?

djbe: Ah yeah, it's a `struct`. Shouldn't matter, `private` type definitions don't go beyond the file (actually scope, file in this case). `private` properties/functions on the other hand don't go beyond the Type (in this case, those access modifiers apply to the scope).

AliSoftware:
@iotize do you think you could change that `fileprivate` into a `private` so we can approve and merge this? Thx!

ryanmasondavies:
> @iotize do you think you could change that `fileprivate` into a `private` so we can approve and merge this? Thx!

Done.

djbe (approved):
I've checked and we don't need to update the template docs (sample code), readme, nor the playground page (it uses the literal template).

LGTM (after a rebase)

Issue #91, opened 2016-02-17
NSString.localizedStringWithFormat to take stringsdict pluralization into account


AliSoftware:
Use `NSString.localizedStringWithFormat(format, args)` (instead of just `String(format:arguments:)` in the strings templates, to take `.stringsdict` pluralization into account.

See @marmelroy's [talk at dotSwift'16 here](https://www.youtube.com/watch?v=R9_jFbQsFWA&t=500).

AliSoftware:
Actually according to the documentation, I'm not so sure that `String(format: arguments:)` doesn't do that already, so maybe there is no need to change anything after all?

marmelroy:
Cool question...

So when you look at the implementation over at https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/String.swift:

```
    public init(format: String, arguments: [CVarArgType]) {
        self = String(format: format, locale: nil, arguments: arguments)
    }
```

whereas localizedStringWithFormat is implemented as:
``` 
    public static func localizedStringWithFormat(
        format: String, _ arguments: CVarArgType...
        ) -> String {
        return String(format: format, locale: NSLocale.currentLocale(),
                              arguments: arguments)
    }
```

According to Apple locale=nil is equivalent to locale=NSLocale.systemLocale() and: 

```Use the system locale when you don’t want any localizations. Use the current locale to format text that you display to users.```

**Comment by marmelroy** [comment]
So localizedStringWithFormat is the correct one.

**Comment by AliSoftware** [comment]
Ah! Good thinking into looking at the Swift OSS source code! Thanks for digging into it!

**Comment by AliSoftware** [comment]
Still feels non-Swifty to me to use a static func when an `init` is present…

I think I'll use `return String(format: format, locale: NSLocale.currentLocale(), arguments: arguments)` directly (so I actually use an initializer) instead of calling the static func `return String.localizedStringWithFormat(format, arguments)`… see any drawback in doing that?

**Comment by marmelroy** [comment]
I see your point... there is something quite ObjC-y about it. :smile: 

Anyway, there's zero difference between the two. Go with what looks nicer.

**Comment by djbe** [comment]
I've used pluralisation in a few projects, and used this for now:
```swift
String(format: NSLocalizedString("(%ld Ratings)", comment: "Ratings count"), data.ratingsCount)
```

It just works™. I honestly don't think SwiftGen needs to add separate cases, because the __keys__ stay the same. 

For reference, the corresponding stringsdict entry:
[...]

PR #716, opened 2020-06-14
Strings: better `localizeFunction`

djbe:
Looking back at #573, I've realised that we should have gone with a function that mirrors `Bundle.localizedString(forKey:...)` as it was initially implemented, and not the current `NSLocalizedString`.

Reason for this is:
- It avoids issues with `genstrings`, Xcode and other tooling to manage strings, because right now they see an `NSLocalizedString` invocation with an invalid (dynamic) key.
- Removes a SwiftLint warning, always nice.

Only disadvantage is that the method must exist on `Bundle`, in for example an extension. Imo, that's almost the same as a global function in terms of namespace pollution (maybe slightly better).

Note 1: not a breaking change, `localizeFunction` hasn't been released yet.
Note 2: I also noticed that the obj-c template forgot to use the current locale, so fixed that as well. Didn't bother adding `localizeFunction` as I don't know if it's worth the effort, and that should definitely go in a separate PR if we really want it.


SwiftGen-Eve:

---

</p>

Documentation/templates/strings/flat-swift4.md
```diff
@@ -21,7 +21,7 @@ You can customize some elements of this template by overriding the following par
 | `enumName` | `L10n` | Allows you to change the name of the generated `enum` containing all string tables. |
 | `noComments` | N/A | Setting this parameter will disable the comments describing the translation of a key. |
 | `publicAccess` | N/A | If set, the generated constants will be marked as `public`. Otherwise, they'll be declared `internal`. |
-| `localizeFunction` | `NSLocalizedString` | Allows you to set your own custom localization function. Your custom function must have the same signature as the one provided by `Foundation`, i.e. `yourFunctionName(_:tableName:bundle:comment:)` |
+| `localizeFunction` | `localizedString` | Allows you to set your own custom localization function. Your custom function must have the same signature as the instance method provided by `Foundation` on `Bundle`, i.e. `bundle.yourFunctionName(forKey:value:table:)` |
```
AliSoftware: "… and must be declared as an `extension` on `Bundle`"?

Though given my other comment on the fact that maybe this shouldn't be a call to the `BundleToken.bundle` instance, that doc might need to be adapted accordingly

Tests/Fixtures/Generated/Strings/flat-swift4/localizable-customName.swift
```diff
@@ -52,8 +52,7 @@ internal enum XCTLoc {
 
 extension XCTLoc {
   private static func tr(_ table: String, _ key: String, _ args: CVarArg...) -> String {
-    // swiftlint:disable:next nslocalizedstring_key
```
AliSoftware, line 55: 🎉

Tests/Fixtures/Generated/Strings/objc-m/localizable.m
```diff
@@ -14,11 +14,12 @@ @implementation BundleToken
...
 
     NSBundle *bundle = [NSBundle bundleForClass:BundleToken.class];
-    NSString *format = [bundle localizedStringForKey:key value:@"" table:tableName];
+    NSString *format = [bundle localizedStringForKey:key value:nil table:tableName];
+    NSLocale *locale = [NSLocale currentLocale];
 
     va_list args;
     va_start(args, key);
-    NSString *result = [[NSString alloc] initWithFormat:format arguments:args];
+    NSString *result = [[NSString alloc] initWithFormat:format locale:locale arguments:args];
```
AliSoftware, line 22: Good catch!

Had a doubt for a minute so had to [dig back that old discussion in old issue](https://github.com/SwiftGen/SwiftGen/issues/91) to ensure the difference between `locale: [NSLocale currentLocale]` vs `locale: nil` vs no locale parameter. And indeed the right solution is to explicitly pass current :+1:

templates/strings/flat-swift4.stencil
```diff
@@ -70,8 +70,7 @@ import Foundation
 
 extension {{enumName}} {
   private static func tr(_ table: String, _ key: String, _ args: CVarArg...) -> String {
-    // swiftlint:disable:next nslocalizedstring_key
-    let format = {{param.localizeFunction|default:"NSLocalizedString"}}(key, tableName: table, bundle: BundleToken.bundle, comment: "")
+    let format = BundleToken.bundle.{{param.localizeFunction|default:"localizedString"}}(forKey: key, value: nil, table: table)
```
AliSoftware:
```suggestion
    let format = {{param.localizeFunction|default:"BundleToken.bundle.localizedString"}}(forKey: key, value: nil, table: table)
```

When you want a custom localisation function it's usually because you want to look for localisation in the main bundle first before the fmk bundle, or in a separate bundle, etc… so making the template calling the function on the instance of `BundleToken.bundle` kinda defeats the idea 😉

AliSoftware (commented):
I'm a bit torn by that fact that this is an instance method on `Bundle` now.

Because one of the main goal of using a custom function is typically when you want to first look for a localisation in the `Bundle.main` first, and fallback into the localisation in the framework's bundle if none is found in the main bundle. Or vice-versa. Or if you want to look at your localisation in a dedicated resource bundle instead of directly in the framework bundle.

So in all those cases, you **don't** want to look into `BundleToken.bundle`, while having the method call be `BundleToken.bundle.XCTLocFunc(…)` ties it with the bundle associated with the framework where you'll include the generated SwiftGen code…

Any drawback of making the whole `localizeFunction` default value include `BundleToken.bundle`, letting the user provide the full base name of the custom function if they customise the param and not forcing them to have a method being called on `BundleToken.bundle`? That would still allow then to either provide a global method, or an static method on a type they want if they don't want to pollute the global namespace, or an instance method on a specific singleton… whatever they need

Tests/Fixtures/Generated/Strings/objc-m/localizable.m
djbe: Yep, had to look up the exact same convo on Slack 😆

djbe:
Understandable, but how would this interact with #644 (template parameter for bundle)? There we discussed adding a template parameter to override which bundle is used (for all templates). Whatever solution we devise here should be compatible with that.

templates/strings/flat-swift4.stencil
AliSoftware: Another advantage of going that route, is that people wanting to just replace the default behaviour to instead just look at localisation in the main bundle, can just pass `Bundle.main.localizedString` for the `localizeFunction` without having to implement a dedicated function. Or `ResourceBundle.bundle.localizeFunction` if they want to make the behaviour look at localization in a resources bundle, without having to declare a custom method implementation for that.

AliSoftware:
> Understandable, but how would this interact with #644 (template parameter for bundle)? There we discussed adding a template parameter to override which bundle is used (for all templates). Whatever solution we devise here should be compatible with that.

imho, #644 doesn't provide the best solution anyway. As discussed over there, the current implementation for the (not-yet merged) #644 is that it hardcodes the fact that it looks for the _first_ bundle that is found inside the framework bundle, and use that if one exists, otherwise use the framework's bundle. That's quite an opinionated choice, that won't fit with all the use cases (what if I have more than one `.bundle` in my framework, one for images, one for strings… and that depending on the resource I want I don't want to always look in the first? What if my resource bundle is not inside my dynamic framework but provided as a sibling bundle? …) so we shouldn't merge #644 as is anyway.

I think the path taken by this PR #716 with allowing parametrisation of the function to retrieve the resource (here the localisable string) is more flexible and allows to adapt to way more custom situations.

djbe:
I meant what was mentioned in the comments about a "bundle name" parameter for example. Totally agree that #644 as is isn't what we want.

AliSoftware:
Yeah but even bundle name doesn't cover all cases. Especially custom fallback mechanisms.

Most of the projects I encountered where I needed to customise the way SwiftGen looked up for the resources were something like "search in main bundle, [sometimes: if not found search in alternate table], and if still not found, search in the framework". That allowed use to make the framework provide sensible defaults, but allow white-label apps to override values for some resources/strings as needed.
Providing just a "bundle name" parameter won't allow the user to provide such a custom behavior. While providing a "resource fetch function" ("localizeFunction") parameter would allow the user to either just pass `Bundle.main.localizedString` or `RsrcBundle.bundle.localizedString` for simple cases but also allow to provide `fetchLocalizedString` as the value of the parameter and implement some `internal func fetchLocalizedString(…)` method in the framework that does the whole logic of {use main bundle, fallback to default table main bundle, then fallback to framework bundle} or anything similar that people might need. Something that "bundle name" parameter wouldn't be able to cover

djbe:
So, to summarise:

djbe:
So to summarize:
- Rename `localizeFunction` to `lookupFunction`
- It is a free function (not on `Bundle`), and add the full parameters list (key and table) inside it

In follow-up PRs (not necessarily 6.2.0), we will:
- Add a template parameter (to all templates that support) to override the bundle, which will default to `BundleToken.bundle`. In essence, what #644 should have been.
- Adapt the `lookupFunction` parameter to other templates that support it: fonts, IB scenes, json, and plist.
- For now, we won't adapt `lookupFunction` to the assets templates, as we'd need multiple lookup functions (one for each type).

Documentation/templates/strings/flat-swift4.md
```diff
@@ -21,7 +21,7 @@ You can customize some elements of this template by overriding the following par
 | `enumName` | `L10n` | Allows you to change the name of the generated `enum` containing all string tables. |
 | `noComments` | N/A | Setting this parameter will disable the comments describing the translation of a key. |
 | `publicAccess` | N/A | If set, the generated constants will be marked as `public`. Otherwise, they'll be declared `internal`. |
-| `localizeFunction` | `NSLocalizedString` | Allows you to set your own custom localization function. Your custom function must have the same signature as the one provided by `Foundation`, i.e. `yourFunctionName(_:tableName:bundle:comment:)` |
+| `lookupFunction` | N/A | Allows you to set your own custom localization function. You must provide the complete function signature (including parameters), and your function accepts 2 parameters: the localization key (`String`), and the localization table (`String`). E.g. `yourFunctionName(forKey:table:)` |
```
AliSoftware:
```suggestion
| `lookupFunction` | N/A | Allows you to set your own custom localization function. The function needs to accepts 2 parameters: the localization key (`String`), and the localization table (`String`). If the function has named parameters, you must provide the complete function signature, including those named parameters – e.g. `yourFunctionName(forKey:table:)` |
```

Reworded because if the function takes unnamed parameters, it's ok to use `myFunction` (it's also ok to pass `myFunction(_:_:)` but not needed). wdyt? (if you're ok with that new wording don't forget to apply the same change suggestion to the 3 other documentation files)

AliSoftware (approved):
Just one suggestion about documentation wording, otherwise LGTM

PR #703, opened 2020-05-27
Enable code coverage #trivial

marcelofabri:
* [x] I've started my branch from the `develop` branch (gitflow)
  * [x] And I've selected `develop` as the "base" branch for this Pull Request I'm about to create
* [x] I've added an entry in the `CHANGELOG.md` file to explain my changes and credit myself  
      _(or added `#trivial` to the PR title if I think it doesn't warrant a mention in the CHANGELOG)_
  * [ ] Add the entry in the appropriate section (Bug Fix / New Feature / Breaking Change / Internal Change)
  * [ ] Add a period and 2 spaces at the end of your short entry description
  * [ ] Add links to your GitHub profile to credit yourself and to the related PR and issue after your description

---

This makes it way easier for me to know that the unit tests are testing the code paths I'm touching

PR #722, opened 2020-06-21
Fix: revert code coverage

### LINKED ISSUE #719: SwiftGen 6.2.0 generates default.profraw files
I've just updated to 6.2.0 and now I'm seeing `default.profraw` files being generated whenever I run `swiftgen`. 

`default.profraw` is apparently related to code coverage and generated by LLVM.

Maybe I introduced it in https://github.com/SwiftGen/SwiftGen/commit/1695642e24a3844434020a4acc230efd9f95cb43?

djbe:
Fixes #719. 

Reverts #703.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,9 @@ _None_
 
 ### Bug Fixes
 
-_None_
+* Prevent generating `default.profraw` (code coverage) files.  
+  [David Jennes](https://github.com/djbe)
+  [#722](https://github.com/SwiftGen/SwiftGen/pull/722)
 
 ### Internal Changes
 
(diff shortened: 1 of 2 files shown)


SwiftGen-Eve:

---

</p>

AliSoftware:
I think it should be possible to still activate code coverage for debug builds, but disable it for release builds (we use release config for our GH/CP/HB releases)?

djbe:
@AliSoftware note that this is at the scheme level, not the config level, so I don't think that's possible.

AliSoftware:
> @AliSoftware note that this is at the scheme level, not the config level, so I don't think that's possible.

Right. I was more thinking about using the right xcodebuild flag when compiling via rake to disable code coverage there during release 🤷🏻

djbe:
Ah, from a quick search, that would maybe be `-enableCodeCoverage NO`. Will have to test that.

djbe:
So, that flag doesn't work:
> The flag -enableCodeCoverage is only supported when testing.

I've also tried these to no avail:
```
CLANG_ENABLE_CODE_COVERAGE=NO GCC_INSTRUMENT_PROGRAM_FLOW_ARCS=NO
```

Can't seem to find much (anything) about disabling this from CLI.

AliSoftware:
Ok, let's disable it completely then.

AliSoftware (approved):

PR #726, opened 2020-06-23
CoreData: Change fetchRequest to makeFetchRequest

### LINKED ISSUE #725: CoreData: Ambiguous use of fetchRequest()
There is an autogenerated method on all `NSManagedObject` objects called `fetchRequest` which has the below implementation:

```swift
/* A new fetch request initialized with the Entity represented by this subclass. 
   This property's getter is only legal to call on subclasses of 
   NSManagedObject that represent a single entity in the model.
 */
@available(iOS 10.0, *)
open class func fetchRequest() -> NSFetchRequest<NSFetchRequestResult>
```

Because we generate a method with the same name (but a more concrete type) it leads to ambiguous use errors unless you explicitly define the type like below which feels a little "un-swifty"

```swift
let fetchRequest: NSFetchRequest<Foo> = Foo.fetchRequest()
```

Although it would be a breaking change I would suggest naming the SwiftGen generated methods to something different like `createFetchRequest`

davidrothera:
Fixes #725

Technically I can see this as a breaking change so added as such in the `CHANGELOG`


SwiftGen-Eve:

---

</p>

CHANGELOG.md
```diff
@@ -6,7 +6,8 @@
 
 ### Breaking Changes
 
-_None_
+* CoreData: Changes `fetchRequest` auto-generated method to `createFetchRequest` to avoid ambiguous function usage.  
+  [David Rothera](https://github.com/davidrothera)
```
djbe: Please add a link to the pull request using the same format as other changelog entries.

djbe (changes requested):
You'll need to apply the same change to the swift 4 template.

I'm not 100% certain what the naming conventions are, and am very bad at it myself, so unless there's a better suggestion, let's go with `createFetchRequest()`.

davidrothera:
> You'll need to apply the same change to the swift 4 template.
> 
> I'm not 100% certain what the naming conventions are, and am very bad at it myself, so unless there's a better suggestion, let's go with `createFetchRequest()`.

Yeah I'm not sure on whether this is the correct name for the method either, it was either this or `makeFetchRequest` when I was thinking about it.

djbe:
@davidrothera thx for the quick PR!

Maybe we can avoid this being a breaking change, if we keep the old method name, and add a new method `createFetchRequest()`. We could maybe even add a `@available` annotation marking the old one as deprecated:

```swift
@available(*, deprecated, renamed: "createFetchRequest", message: "To avoid collisions with the less concrete method in `NSManagedObject`, please use `createFetchRequest()` instead.")
@nonobjc internal class func fetchRequest() -> NSFetchRequest<SecondaryEntity> {
  return NSFetchRequest<SecondaryEntity>(entityName: entityName)
}

@nonobjc internal class func createFetchRequest() -> NSFetchRequest< SecondaryEntity > {
  return NSFetchRequest<SecondaryEntity>(entityName: entityName)
}
```

davidrothera:
> @davidrothera thx for the quick PR!
> 
> Maybe we can avoid this being a breaking change, if we keep the old method name, and add a new method `createFetchRequest()`. We could maybe even add a `@available` annotation marking the old one as deprecated:
> 
> ```swift
> @available(*, deprecated, renamed: "createFetchRequest", message: "To avoid collisions with the less concrete method in `NSManagedObject`, please use `createFetchRequest()` instead.")
> @nonobjc internal class func fetchRequest() -> NSFetchRequest<SecondaryEntity> {
>   return NSFetchRequest<SecondaryEntity>(entityName: entityName)
> }
> 
> @nonobjc internal class func createFetchRequest() -> NSFetchRequest< SecondaryEntity > {
>   return NSFetchRequest<SecondaryEntity>(entityName: entityName)
> }
> ```

Thanks for the suggestion, yeah I would always prefer not having to break backwards compatibility. I'll get this changed shortly 👍

davidrothera:
I've added the method back in along with a deprecation notice, I updated the `CHANGELOG` to move it to the "New Features" section but not sure whether it would be best there or under one of the other headings.

Documentation/templates/coredata/swift5.md
```diff
@@ -46,7 +46,7 @@ internal class MainEntity: NSManagedObject {
     return NSEntityDescription.entity(forEntityName: entityName, in: managedObjectContext)
   }
 
-  @nonobjc internal class func fetchRequest() -> NSFetchRequest<MainEntity> {
```
djbe, line 49: Please also make these changes in the swift 4 docs.

djbe (changes requested):
So we also need to update:
- swift 4 docs
- readme usage example code
- playgrounds page code

djbe:
Changelog location is ok for now, we can always move it later if really needed.

davidrothera:
Hey @djbe, have you had chance to look over the latest changes you requested?

djbe:
Allright, seems like those last changes cover everything.

One unfortunate bit is that I read through the API design guidelines:
> Begin names of factory methods with “make”, e.g. x.makeIterator().

So it should become `makeFetchRequest()` 😅 Luckily, we can find/replace now that the initial effort is done.
Feel free to rebase your commits if you find that easier.

davidrothera:
Good find and nice to know, there are probably some places in my own code which I should clean up in that case as well 😝

davidrothera:
Ok should be good to go now @djbe

djbe (approved):
Nice, thx for the update!

I'm approving this now, but not merging it yet, as we still need to release 6.2.1. Depending on how that release goes, we may need a 6.2.2. After all that, this can be merged for 6.3.0.

PR #729, opened 2020-06-23
Small code improvements #trivial

### LINKED ISSUE #547: SwiftGen config warnings are invisible in Xcode

djbe:
Implement a few things that were mentioned in other issues & PRs:
- Rename context files to `ParserName+Context.swift` (https://github.com/SwiftGen/SwiftGen/pull/702#discussion_r431420916)
- Helper function for creating file filter regex for parsers (https://github.com/SwiftGen/SwiftGen/pull/702#discussion_r431416530)
- ~Fixes #547 by logging warnings to `stdout` instead of `stderr`.~ Seems like it's not needed, need more input


SwiftGen-Eve:

---

</p>

Documentation/ConfigFile.md
```diff
@@ -109,7 +109,7 @@ If you're running `swiftgen` as part of a Script Build Phase in Xcode, this espe
 strings:
   inputs: ${PROJECT_DIR}/${TARGET_NAME}/Resources/
   outputs:
-    templateName: swift4
+    templateName: structured-swift5
```
AliSoftware, line 112: good catch!

README.md
```diff
@@ -247,7 +247,7 @@ SwiftGen comes bundled with some templates for each of the parsers (`colors`, `c
 
 💡 You can use the `swiftgen template list` command to list all the available bundled templates for each parser, and use `swiftgen template cat` to show a template's content and duplicate it to create your own variation.
 
-For more information about how to create your own templates, [see the dedicated documentation](Documentation/Creating-your-templates.md).
+For more information about how to create your own templates, [see the dedicated documentation](Documentation/Articles/Creating-custom-templates.md).
```
AliSoftware, line 259: 👍 
not sure how search/replace missed those ones during the previous PR 😓

Sources/SwiftGen/Logs.swift
```diff
@@ -47,7 +47,7 @@ func logMessage(_ level: LogLevel, _ string: CustomStringConvertible) {
     case .info:
       fputs(ANSIColor.green.format("\(string)\n"), stdout)
     case .warning:
-      fputs(ANSIColor.yellow.format("swiftgen: warning: \(string)\n"), stderr)
+      fputs(ANSIColor.yellow.format("swiftgen: warning: \(string)\n"), stdout)
     case .error:
       fputs(ANSIColor.red.format("swiftgen: error: \(string)\n"), stderr)
```
AliSoftware: Should we use `stdout` for errors here if we detect that the context is Xcode (like I'm sure there's some env var like `XCODE_VERSION` defined we can check)? Or do errors get logged properly by Xcode even in `stderr` and it's only for warnings that it didn't work (would be a strange asymmetry from Xcode… but 🤷 )

iirc we decided on that (stdout for warnings, stderr for errors) because that's how SwiftLint also does it, but still it's worth checking…

Sources/SwiftGenKit/Parsers/Filter.swift
```diff
@@ -33,6 +33,17 @@ public struct Filter {
   }
 }
 
+extension Parser {
+  /// Create the regex for a filter for the given list of file extensions. The extensions will be matched at the end
+  /// of a filename, and will be matched case-insensitively.
+  ///
+  /// - Parameter extensions: The list of file extensions. Note that these can contain regex syntax if needed.
+  /// - Returns: The regex as a `String`.
+  static func filterRegex(for extensions: [String]) -> String {
```
AliSoftware: naming nitpick: wondering if the API shouldn't be called `filterRegex(extensions:)` or `filterRegex(forExtensions:)`.

imho seeing just `filterRegex(for: x)` at call side is not as readable and clear what's expected as parameter here

Sources/SwiftGenKit/Parsers/Filter.swift
```diff
@@ -33,6 +33,17 @@ public struct Filter {
...
+    "[^/]\\.(?i:\(extensions.joined(separator: "|")))$"
```
AliSoftware: We should `NSRegularExpression.escape(…)` those extensions. I mean, in practice it's quite unlikely that we'd ever have an extension that uses anything other than alphanumeric characters, but still, for correctness' sake… :wink:

Sources/SwiftGenKit/Parsers/Yaml/YamlParser.swift
```diff
@@ -33,7 +33,7 @@ public enum Yaml {
     }
 
     public class var defaultFilter: String {
-      "[^/]\\.(?i:ya?ml)$"
+      filterRegex(for: ["ya?ml"])
```
AliSoftware: After `filterRegex(forExtensions:)` ends up applying `NSRegularExpression.escape(…)` (see comment above) then for this we would use `["yml", "yaml"]` instead – which would actually make it more readable imho

AliSoftware (commented):
Nice improvements! 👍

Sources/SwiftGenKit/Parsers/CoreData/CoreDataParser.swift
```diff
@@ -41,7 +41,7 @@ public enum CoreData {
       self.warningHandler = warningHandler
     }
 
-    public static let defaultFilter = "[^/]\\.xcdatamodeld?$"
+    public static let defaultFilter = filterRegex(for: ["xcdatamodeld?"])
```
AliSoftware: if we apply `NSRE.escape` don't forget to also change this to `"xcdatamodel", "xcdatamodeld"`

Sources/SwiftGen/Logs.swift
djbe: Not 100% sure about this one. We have a check for "dumb" terminals, but would that match too much? (we currently use it to decide whether to output in color)

AliSoftware: I'm confused now.

Added a shell script build phase to swiftgen to test
```sh
echo "swiftgen: warning: to stdout"
>&2 echo "swiftgen: warning: to stderr"
echo "swiftgen: error: to stdout"
>&2 echo "swiftgen: error: to stderr"
```
image

But it prints all the warnings correctly in all cases…
Or was this an issue with a specific and older version of Xcode which has been fixed now?
image

djbe: Possibly...

AliSoftware: I also just remembered that the initial reason for our logs to be printed in stderr is that swiftgen allows to print the generated content to stdout (`swiftgen xcassets --templateName swift5 Assets.xcassets` will print to stdout since there's no `--output`). See also `OutputDestination`. How should we handle those cases so that when printing generated content to stdout the output is not mixed up with the log messages then?

Maybe one more reason to conditionally print to stderr vs stdout depending on being on Xcode? 🤔

djbe: Well, seeing your tests above, it doesn't matter if we print to `stderr` or not. I think that for now we should revert this change, and ask the OP in #547 if they still have the issue.

Sources/SwiftGenKit/Parsers/Filter.swift
```diff
@@ -33,6 +33,17 @@ public struct Filter {
...
 }
 
+extension Parser {
+  /// Create the regex for a filter for the given list of file extensions. The extensions will be matched at the end
+  /// of a filename, and will be matched case-insensitively.
+  ///
+  /// - Parameter extensions: The list of file extensions.
+  /// - Returns: The regex as a `String`.
+  static func filterRegex(forExtensions extensions: [String]) -> String {
+    "[^/]\\.(?i:\(extensions.map { NSRegularExpression.escapedPattern(for: $0) }.joined(separator: "|")))$"
```
AliSoftware: nit for regex readability:
```suggestion
    let extensionsList = extensions.map { NSRegularExpression.escapedPattern(for: $0) }.joined(separator: "|")
    // - require at list one character that is not a "/" before the "." in order to avoid matching dotfiles
    // - then a literal "."
    // - then one of the accepted list of extensions, in a case-insensitive way (`(?:i…)`), at the end of the string (`$`)
    return "[^/]\\.(?i:\(extensionsList))$"
```

AliSoftware (commented):
Appart from the open question about stdout/stderr and how we should handle this, the rest LGTM

Sources/SwiftGen/Logs.swift
AliSoftware: Yeah, I agree. Let's revert and keep it to stderr for now, ping the OP (and close the issue if no response in a couple of months or so)

Sources/SwiftGenKit/Parsers/Filter.swift
```diff
@@ -33,6 +33,21 @@ public struct Filter {
...
+  /// Create the regex for a filter for the given list of file extensions. The extensions will be matched at the end
+  /// of a filename, and will be matched case-insensitively.
+  ///
+  /// - Parameter extensions: The list of file extensions.
+  /// - Returns: The regex as a `String`.
+  static func filterRegex(forExtensions extensions: [String]) -> String {
+    let extensionsList = extensions.map { NSRegularExpression.escapedPattern(for: $0) }.joined(separator: "|")
+    // - require at least one character that is not a "/" before the "." in order to avoid matching dotfiles
+    // - then a literal "."
+    // - then one of the accepted extensions, in a case-insensitive way (`(?:i…)`), at the end of the string (`$`)
```
AliSoftware, line 46: > in a case-insensitive way (`(?:i…)`), at the end of the string (`$`)

Dammit and it's only now that I've made that suggestion that you already mentioned that last part in the main function's documentation 🤦 

> The extensions will be matched at the end of a filename, and will be matched case-insensitively.

Meh, better repeating ourselves in the doc than not documenting enough I guess, so might as well leave it 🤷‍♂️

AliSoftware (approved):
👌

PR #733, opened 2020-06-25
XCAssets: image loading performance on macOS

### LINKED ISSUE #648: Asset Catalog: on macOS images are not loaded from cache
The swift4 template uses `bundle.image(forResource: )` to load images from any bundle. However this method does not cache the image contrary to `NSImage(named: )`. This can slow down the UI if we load images from disk for each item in a huge list.

Perhaps we should test if the current bundle is the main one, and use the cached image initilializer. 
```swift
    let image: NSImage?
    if bundle == Bundle.main {
      image = NSImage(named: NSImage.Name(name))
    } else {
      image = bundle.image(forResource: NSImage.Name(name))
    }```

### PR DESCRIPTION
**Pull request by djbe**
Fixes #648.

On macOS, images loaded via `bundle.image(forResource:)` are not cached. Images loaded using `NSImage(named:)` are cached, but that method only works in the `main` bundle. 

The solution is simple: use the faster method if we can, namely check if the bundle == `main`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,10 @@ _None_
 * CoreData: Deprecates `fetchRequest()` and adds `makeFetchRequest()` to avoid ambiguous function usage.  
   [David Rothera](https://github.com/davidrothera)
   [#726](https://github.com/SwiftGen/SwiftGen/pull/726)
+* XCassets: image assets now load faster on macOS if they're in the `main` bundle. MacOS only provides a caching mechanism for images in the `main` bundle, for other cases you may need to provide your own caching mechanism as needed.  
+  [David Jennes](https://github.com/djbe)
+  [#648](https://github.com/SwiftGen/SwiftGen/issues/648)
+  [#733](https://github.com/SwiftGen/SwiftGen/pull/733)
 
 ### Bug Fixes
 
@@ -56,7 +60,7 @@ There are no major changes in this release, although JSON & Plist template write
   [@fjtrujy](https://github.com/fjtrujy)
 * Avoid breaking the system swift installation when installing SwiftGen via Homebrew on macOS 10.14.4 or higher.  
   [David Jennes](https://github.com/djbe)
-  [#686](https://github.com/SwiftGen/SwiftGen/issue/686)
+  [#686](https://github.com/SwiftGen/SwiftGen/issues/686)
   [#730](https://github.com/SwiftGen/SwiftGen/pull/730)
 
 ### Internal Changes
--- a/Tests/Fixtures/Generated/XCAssets/swift4/all-allValues.swift
+++ b/Tests/Fixtures/Generated/XCAssets/swift4/all-allValues.swift
@@ -239,7 +239,8 @@ internal struct ImageAsset {
     #if os(iOS) || os(tvOS)
     let image = Image(named: name, in: bundle, compatibleWith: nil)
     #elseif os(macOS)
-    let image = bundle.image(forResource: NSImage.Name(name))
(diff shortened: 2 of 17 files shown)

### CODE REVIEW

**Comment by SwiftGen-Eve** [comment]

---

</p>

**Comment by eaigner** [comment]
On a different note, why not just make the `image` property lazy and the `ImageAsset` a `final class` like it is on `ColorAsset`?

**Comment by djbe** [comment]
See my comment in the original issue:
https://github.com/SwiftGen/SwiftGen/issues/648#issuecomment-649121005
> There's a 2nd concern though: once loaded into memory, images will never deallocate with this method. With colors this isn't much of an issue because at worst it's a few bytes, but with images, that can result in quite a lot of memory, 100MB+ for a normal sized application. Note that this solution would be applied on all platforms, so also on iPhones and Apple Watches, which are much more memory constrained.

**Comment by eaigner** [comment]
Thats true. But I still think the best solution would be just to provide multiple accessors. For instance instead of just having an `image` accessor, there could also be a `lazyImage` accessor. That way the consumer can chose what he wants. I feel its kind of wrong to force a watchOS limitation on a macOS app for instance.

Nobody will complain on macOS that a 16px icon for a table cell doesn't deallocate.

**Comment by djbe** [comment]
Sounds like a danger prone approach. If someone wants to use such an approach, they're welcome to write a custom template. The default templates in SwiftGen should not have such pitfalls.

This PR tries to improve the performance for (most) macOS users without such dangers. 

Further discussion on how to tackle this macOS issue in general should be done in #648 (or open a new issue with suggestions). The discussion here should be about this specific PR.

File: CHANGELOG.md [hunk 1]
```diff
@@ -56,7 +60,7 @@ There are no major changes in this release, although JSON & Plist template write
   [@fjtrujy](https://github.com/fjtrujy)
 * Avoid breaking the system swift installation when installing SwiftGen via Homebrew on macOS 10.14.4 or higher.  
   [David Jennes](https://github.com/djbe)
-  [#686](https://github.com/SwiftGen/SwiftGen/issue/686)
+  [#686](https://github.com/SwiftGen/SwiftGen/issues/686)
```
AliSoftware: Good catch. Added a `Danger` rule to catch potential others in the future in https://github.com/SwiftGen/SwiftGen/pull/745

AliSoftware (approved):
Nice!

PS: Wondered if we should not make the `name` property be an `NSImage.Name` directly instead of a `String` in the case of `#if os(macOS)` – as it seems that it's the type that is used on this platform for referencing the name –… but that would be a breaking change, so probably not worth it

Tests/Fixtures/Generated/XCAssets/swift4/all-allValues.swift
```diff
@@ -239,7 +239,8 @@ internal struct ImageAsset {
     #if os(iOS) || os(tvOS)
     let image = Image(named: name, in: bundle, compatibleWith: nil)
     #elseif os(macOS)
-    let image = bundle.image(forResource: NSImage.Name(name))
+    let name = NSImage.Name(self.name)
+    let image = (bundle == .main) ? NSImage(named: name) : bundle.image(forResource: name)
```
AliSoftware, line 243: For refs: I was wondering if `Bundle` was conforming to `Equatable` in a more specific way that the default `NSObject.==` it is inheriting from (which is just pointer comparison) – which could lead to issues if we were trying to compare `Bundle(path: pathToMain)!` vs `Bundle.main`.

But after some tests it seems that the `Bundle.init` magically caches and shares bundle instances pointing to the same path, so we're good using `==` here anyway, even if it might just use pointer comparison 👍

PR #734, opened 2020-06-25
New parser: file resources

### LINKED ISSUE #109: Add support for bundle folder
Hi,

please add support for items in Resources.bundle e.g.: Resources.bundle/images/myimage@2x.png. 

We use bundle folders for easy collaboration with designers. They behave like referenced folders in Xcode, no change to the .pbxproj file for easy GIT. No need to handle .xcassets JSON structure, no slow drag-&-drop of resources in Xcode, no problems with resources renaming and moving. Designers just export PNGs into the bundle and push to GIT repository. No need to learn designers anything about the Xcode and GIT conflicts resolving.

Bundles have too many pluses to not to use them.

### LINKED ISSUE #146: Support for loose images (outside of XCAssets).

### LINKED ISSUE #284: Add file names generator (nibs, plists, etc)
Hi,

@AliSoftware We've met at AppDevCon, I wanted to adapt #108 to the new structure and making it more generic. As discussed back then, I'm opening an issue to discuss this.
You told me you were working on writing some docs on how to contribute now the project has been split in subprojects.
The first step would be to work on the SwiftGenKit part, right ?

### LINKED ISSUE #496: It's possible to create a struct of all files inside one folder?
Similar to the storyboards, there should be a template, which summarizes files from a folder with subfolders in a struct. Do you have any idea to realize this with the upcoming version 6.0?

Example output:
```
struct File {
	var name: String

	// MARK: - Getter
	var path: String {

		let bundle = Bundle(for: BundleToken.self)
		return bundle.path(forResource: self.name, ofType: nil) ?? ""
	}
	var url: URL {
		return URL(fileURLWithPath: self.path)
	}


	enum Html {
		/// test.html
		static let test = File(name: "test.html")
	}
	enum Json {
		/// test.json
		static let test = File(name: "test.json")
	}
}

private final class BundleToken {}
```

### LINKED ISSUE #637: Add support for files
It would be good to add support for any file types and not just Plist, Strings and Text. 
I am working on a project where we inject JS and CSS files into the webpage.

### LINKED ISSUE #665: [Feature] General file resources from bundle
I'd like to work towards adding support for parsing other file resources that need to be accessed from the bundle, usually with `Bundle.main.url(forResource:withExtension:)` and the like.

Looking through the issues it seems this is a desired functionality, but not much work has been towards achieving it
#109, #146, #284, #496, #637 

My thoughts on the implementation:
- A good name for the subcommand might be `resources`
- Correct me if I'm wrong, but it looks like `inputs` and `filter` are the best way to specify files or file types to include
- It seems I can probably implement the generator pretty quickly duplicating functionality from the yaml and plist generators
- Possible params:
  - enumName (default `Resource`)
  - publicAccess (default `false`)

Thoughts on template:
- Containing enum: `Resource`
- Each resource as a struct `ResourceFile`
- Each resource should have a name including the extension for uniqueness and clarity
- `ResourceFile` can expose two computed properties
  - `url` for `url(forResource:withPath:)
  - `path` for `path(forResource:ofType:)

Need input on:
- Arrays of resources
- Handling subdirectories implicitly/explicitly/disregarding

Thoughts? Ideas? Did I miss anything?

mgray88:
Add a new parser which allows accessing any other specified files that need to be accessed from the bundle, usually with `Bundle.main.url(forResource:withExtension:)` and the like.

---

Fixes #665, fixes #637, fixes #496, fixes #284, fixes #146, fixes #109.


Sources/SwiftGenKit/Parsers/Resources/ResourceParser.swift
```diff
@@ -0,0 +1,40 @@
+//
+// SwiftGenKit
+// Copyright © 2019 SwiftGen
+// MIT Licence
+//
+
+import Foundation
+import PathKit
+
+public enum Resource {
```
AliSoftware: I think that the term `Resource` can be misleading. Is it a parser to process `.rsrc` files (like back in the ResEdit days 😅 ), or some kind of other file format? I got confused at first 😉 

Naming is hard… but I suggest to use something like `DirList` instead (for the name of the enum, parser, the command, etc), what do you think?

Other suggestions/ideas I had in mind: `FileList`, `FileSystem`, …

Sources/SwiftGenKit/Stencil/ResourcesContext.swift
```diff
@@ -0,0 +1,26 @@
...
+      .map(map(file:))
+
+    return [
+      "files": files
+    ]
+  }
+
+  private func map(file: Resource.File) -> [String: Any] {
+    return [
+      "name": file.name,
```
AliSoftware: I think it could be useful to not only have the `"name"` but also the `"path"` (relative path).

That way, if a developer adds a folder as a "Folder reference" to their project – leading to the directory structure to be kept and copied as-in in the final app bundle – they can still access it using that `path` (while if it's included in the folder as a normal file it will end up at the root of the bundle's resource directory and users would use `name` instead)

We actually used a similar logic in the Fonts parser if you want to get some inspiration (see [here](https://github.com/SwiftGen/SwiftGen/blob/stable/Sources/SwiftGenKit/Parsers/Fonts/FontsParser.swift#L24) and [here](https://github.com/SwiftGen/SwiftGen/blob/stable/Sources/SwiftGenKit/Stencil/FontsParser%2BContext.swift#L23-L24) for example)

AliSoftware (commented):
Hello @mgray88 and thanks for the draft PR!

I know it's only a draft for now, but:
 - Could you please add a description as to what this PR does?
 - It seems to be an implementation of #665 – which you've referenced in the title, tho honestly I missed it at first – but mentioning it in the PR description helps having GitHub cross-link the PR and issue. You might also want to cross-reference #284.

 Thx! 🙂

Sources/SwiftGenKit/Stencil/ResourcesContext.swift
djbe: 👍 for paths. Note that the relative path calculation actually happens here: https://github.com/SwiftGen/SwiftGen/blob/stable/Sources/SwiftGenKit/Parsers/Fonts/CTFont.swift#L22

We can easily add mime-type info as well, or some sort of "readable" type information. In case the template wants to do some special case handling for some specific types.

Sources/SwiftGenKit/Parsers/Resources/ResourceParser.swift
mgray88: Naming is definitely hard 😆 
I like `FileList`, or maybe even just `Files`

mgray88:
I'm running into the issue again where swiftlint is ignoring the `//swiftlint:disable all` at the top of my Generated Fixtures files. The linter output shows it linting the other generated files and successfully ignoring them. However it reaches the files I've added, and spits out `file_header` violations

AliSoftware:
I can take a look tomorrow (I'm on my phone rn) to check if it's indeed what I think it is, but it could be expected. Because we have a somewhat special setup for linting our template files

Basically the way our linting script is set up on SwiftGen for the generated files in fixtures is that it intentionally removes the `// swiftlint:disable all` from the generated files before linting them. The goal of that is so that:

1. We, as contributors and creators of templates that are gonna ship with SwiftGen, want to ensure we ship templates that respect some linter rules we set for ourselves. Like we want the templates that we ship with SwiftGen to be consistent in their indentation, use of spaces around colons, etc, and pass the `swiftlint.yml` rules of the SwiftGen repo itself
2. But, for people _using_ those templates in their projects, they might be using a different set of SwiftLint rules, etc, so even if they wouldn't trigger _our_ rules they could still trigger _the user's_ rules. Which is why all our templates include `// swiftlint:disable:all` once shipped and used by end users

I can get how this could be confusing and seem convoluted 😅 (we should definitively document that special setup in our CONTRIBUTING.md btw…). But that's the way we found to both disable swiftlint for end users of our templates while still be able to lint those templates for ourselves when working on SwiftGen — to ensure they are at least consistent with _our_ style rules.

[EDIT] If you're curious, [here is where the magic happens](https://github.com/SwiftGen/SwiftGen/blob/stable/Scripts/SwiftLint.sh#L38-L46)

Sources/SwiftGenKit/Parsers/Resources/ResourceParser.swift
```diff
@@ -0,0 +1,40 @@
...
+
+    public required init(options: [String: Any] = [:], warningHandler: Parser.MessageHandler? = nil) throws {
+      self.options = try ParserOptionValues(options: options, available: Parser.allOptions)
+      self.warningHandler = warningHandler
+    }
+
+    public static var defaultFilter: String = ""
+
+    public func parse(path: Path, relativeTo parent: Path) throws {
+      files.append(try File(path: path, relativeTo: parent))
```
AliSoftware: I think ultimately we want it to be recursive, to list files inside folders deep inside the parsed root path, right?

templates/resources/swift5.stencil
```diff
@@ -0,0 +1,64 @@
...
+// MARK: - Resource Files
+
+// swiftlint:disable identifier_name line_length number_separator type_body_length
+{{accessModifier}} enum {{enumName}} {
+  {% for file in files %}
+  {{accessModifier}} static let {{file.name|swiftIdentifier:"pretty"|lowerFirstWord|escapeReservedKeywords}}{% if file.ext %}{{file.ext|swiftIdentifier:"pretty"|capitalize|escapeReservedKeywords}}{% endif %} = {{resourceType}}(name: "{{file.name}}", ext: {% if file.ext %}"{{file.ext}}"{% else %}nil{% endif %})
+  {% endfor %}
+}
+
+// MARK: - Implementation Detailsh
```
AliSoftware:
```suggestion
// MARK: - Implementation Details
```

templates/resources/swift5.stencil
```diff
@@ -0,0 +1,64 @@
...
+    }
+    return result
+  }
+}
+
+// swiftlint:disable convenience_type
+private final class BundleToken {
+  static let bundle: Bundle = {
+    Bundle(for: BundleToken.self)
+  }()
```
AliSoftware:
```
static let bundle = Bundle(for: BundleToken.self)
```
No need for `{ foo }()` when you can use `foo` directly 😉

(Our templates used to use that uselessly-complicated form too — not sure why — until recently when we changed it to the simpler form; so you probably borrowed that by copy/pasting before we changed it)

Sources/SwiftGenKit/Parsers/Resources/ResourceParser.swift
mgray88: Agreed

templates/resources/swift5.stencil
mgray88: Yep, copy/pasted. Fixed

mgray88:
Gotcha. That all makes sense. I did notice the `Scripts/SwiftLint.sh` but it doesn't appear to discriminate based on folder/file names and just globs everything under `Tests/Fixtures/Generated`

AliSoftware:
Yes it does force-lint (= remove disable:all before linting) everything under `Tests/Fixtures/Generated` but that's expected as that's where the fixtures generated by bundled templates are expected to be and that's what we want to lint.

Did you use the "Generate Outputs" Xcode scheme to auto-generate those fixtures from the templates (so that they can then be used by our unit tests)? Or did you write them by hand?

templates/files/swift5.stencil
```diff
@@ -0,0 +1,68 @@
+// swiftlint:disable all
+// Generated using SwiftGen - https://github.com/SwiftGen/SwiftGen
```
AliSoftware: Is that the proper type of dash that you used here (em dash vs en dash vs hyphen)? Not totally sure as it's hard to check from iPhone, but I think that could be the cause of your `file_header` violations 😅

```suggestion
// Generated using SwiftGen — https://github.com/SwiftGen/SwiftGen
```

mgray88: That would be nuts, but no I believe it's the right dash. I just ran the generator (I had written them by hand) and the header remained the same

AliSoftware: Tbh now seeing how the GitHub suggestion is rendered, highlighting the difference before/after, I'm pretty sure that's the reason tho. The one you used seems to be a hyphen while the one I put in my GitHub suggestion seems to be 1 or 2 pixels wider and be an em dash (which is the more grammatically correct one in this context anyway). Veeery subtle I agree, but still different thus triggering swiftlint

It's not surprising that running the generator scheme didn't change anything as you also used the hyphen instead of em dash in the template itself (`.stencil`) 😛 you need to fix that at the template level, then regenerate to update the fixtures using the template

mgray88: Ah. Haha, yes I see it now. Thanks :)

AliSoftware: Cool! Pfeew, that was a tricky and super subtle one tho 😅

SwiftGen-Eve:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

---

</p>

mgray88:
Can anyone provide insight on why the lint task is failing? All it says is `Danger has failed this build` but all the lint step succeeded with 0 violations. Oh wait, Danger is failing because of the CHANGELOG isn't it?

<strike>Also there is no output for the failing test. The tests build and run fine on my computer so I'm not sure how to fix them on CI</strike> It initially had a failed run, but now it doesn't?

mgray88:
@AliSoftware @djbe Will the cli additions need to be included with this PR or a new one? I'm also working on writing up some of the documentation

mgray88:
Also I'm thinking I should switch it to be structured enum as the default and have an option for a flattened structure. Flattened file list presents the possibility of name clashes, which I'm not sure how/if we should handle that

AliSoftware:
> Can anyone provide insight on why the lint task is failing? All it says is `Danger has failed this build` but all the lint step succeeded with 0 violations. Oh wait, Danger is failing because of the CHANGELOG isn't it?

Yup, because Danger generated an _error_ message (we consider the entry to the CHANGELOG mandatory), that's why the CI ended up with a _failure_. That is intentional, it's to ensure that we won't accidentally merge the Pull Request if you haven't given yourself the due credit in the CHANGELOG first.

> Will the cli additions need to be included with this PR or a new one?

I'd say so, yes. It _shouldn't_ be too complicated once you have the general command running to make the CLI counter part of it calling the command.

>  I'm also working on writing up some of the documentation

❤️ 

> Also I'm thinking I should switch it to be structured enum as the default and have an option for a flattened structure. Flattened file list presents the possibility of name clashes, which I'm not sure how/if we should handle that

I haven't had the change to review the code yet, so not sure I can provide good feedback here, I'll get back to you when I have more time to review your recent changes.

But my gut feeling, without having seen what the current code generates, would be for the ideal solution to be similar to what we do with `strings` parser. That is, provide everything needed in the `Context` to support both, then implement 2 different templates (one which uses the context to generate the `flatten` list and one which uses the same context to generate the `structured` one) that we will both ship as part of SwiftGen to let the user choose which one they prefer.

([EDIT] Actually I just saw that you implemented that via a parser option. That works too I guess.)

mgray88:
> That is, provide everything needed in the `Context` to support both, then implement 2 different templates (one which uses the context to generate the `flatten` list and one which uses the same context to generate the `structured` one) that we will both ship as part of SwiftGen to let the user choose which one they prefer.

I was starting to think I should go that route instead of what I have done which is expose the Parser's `options` to the `stencilContext` generator and return different contexts based on the option set.

mgray88:
Am I right in understanding all that's needed for the CLI is the `.init` lines added to `ParserCLI.swift`?

AliSoftware:
> Am I right in understanding all that's needed for the CLI is the `.init` lines added to `ParserCLI.swift`?

It has been a while since I've touched that part, but from a quick look at it I think that's indeed all there is to it! 👌 

---

You could use the occasion to edit the "swiftgen" Xcode scheme, select the "Run" action on the left, then the "Arguments" tab. There you'll find that we put a pre-defined a list of useful arguments that you might want to pass to swiftgen when debugging it in Xcode.

For example by default the arguments passed by default currently when you do a ⌘R in Xcode are `--help`, since that's the only one with a checkmark next to it:
<details><summary>Expand to see screenshot</summary>image</details>

I encourage you to add your own command here (ideally amongst the other `run …` pre-defined commands and keeping alphabetical order, so `run files …` would be right before the `run font …` line) in the predefined list.

Then you can just uncheck the `--help` line and check your new line to choose to send those arguments instead when doing a ⌘R, which will let you debug your command right in Xcode as if it was executed via `swiftgen run files …` in the terminal 🙂

Sources/SwiftGen/ParserCLI.swift
```diff
@@ -100,6 +100,12 @@ extension ParserCLI {
       description: "generate code for custom yaml configuration files",
       pathDescription: "YAML files (or directories that contain them) to parse."
     ),
+    .init(
+      parserType: Files.Parser.self,
+      name: "files",
+      description: "generate code for referencing specified files",
+      pathDescription: "Files and/or directories to recursively search"
+    ),
```
AliSoftware: Perfect 👍 
Just a nitpick: we usually keep that list of `.init` ordered by the `name` property, so the parser for your new `"files"` command should go before the one for `"fonts"`. It doesn't change anything in the behavior, but it's just to satisfy our OCD 😅

mgray88: Ah. I guess my OCD didn't look close enough. I saw they weren't exactly in order by `parserType` so I placed it at the end :)

I'll change it up

Documentation/SwiftGenKit Contexts/files.md
```diff
@@ -0,0 +1,60 @@
...
+
+The files parser accepts multiple files or directories, which it will recursively search for files that match the given filter (default `.*`).
+
+The list of files will be converted into a structured directory tree.
+
+## Output
+
+The output context has the following structure:
+
+-`files`:`Array` — list of files in the base directory specified
```
AliSoftware: I haven't double-checked at how we did format the documentation for the other parsers, but I think it would be more clear here to say that it's an `Array<File>`, then below on line 19 say that "Each `File` has the following properties". wdyt?

Documentation/SwiftGenKit Contexts/files.md
```diff
@@ -0,0 +1,60 @@
...
+-`dirs`:`Dictionary` — tree structure of subdirectory and its contents
```
AliSoftware: Seems to me that this is not a dictionary but more an array of dictionaries, right?
(With each Dictionary/entry having the keys `name`, `files` and `dirs`)

(Actually you properly documented it as an `Array` for the recursive entry on line 17, which confirms my understanding that this one should be one as well)

Documentation/SwiftGenKit Contexts/files.md
```diff
@@ -0,0 +1,60 @@
...
+ -`name`:`String` — name of the directory
+ -`files`:`Array` — list of files in this directory
+ -`dirs`:`Array` — recursive tree of subdirectories
+
+Each file has the following properties:
+
+-`name`:`String` — name of the file
+-`ext`:`String` — file extension
+-`path`:`String` — the path to the file, relative to the folder being scanned
+-`mimeType`:`String` — mime type of the file
```
AliSoftware: My OCD speaking here, but I'd add a space after each of those `:` 😛

Documentation/SwiftGenKit Contexts/files.md
```diff
@@ -0,0 +1,60 @@
+# Files parser
+
+## Input
+
+The files parser accepts multiple files or directories, which it will recursively search for files that match the given filter (default `.*`).
```
AliSoftware: Given that unlike any other parsers we have, this one is a bit special as it doesn't actually parse the _content_ of the files that it scans, but just their name and mimetype, I'd put a little more intro for this one, as it could be confusing for people who don't have more context to understand what it does ("but wait, what type of files does it parse? binary files? text files??")

For example maybe lead with:

> The files parser is intended to just list the name and mimetype of the files and subdirectories in a given directory. Unlike other parsers, this one doesn't parse the file _content_ per se, but only list those files so that you can point to them (get back their path or URL relative to your bundle) easily in your code.

Open for other wordings, I'm not super inspired and it's late here 😅 , but hopefully you see my point, the goal being to avoid the potential confusion for users that would think that this parser reads the content of the files like all our other parsers do.

Documentation/templates/files/swift5.md
```diff
@@ -0,0 +1,51 @@
+## Template Information
+
+| Name      | Description       |
+| --------- | ----------------- |
+| File name | files/swift5.stencil |
```
AliSoftware: I'm nitpicking again, and naming is hard, but wondering if we should name that template just `swift5` like you did here, or `structured-swift5`, to mimic the naming we've been using for the name of the `strings` templates?

Documentation/templates/files/swift5.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+[Full generated code](../../../Tests/Fixtures/Generated/Files/swift5/defaults.swift)
+
+## Usage example
+
+```swift
+// Access files using the `url` or `path` fields
+let txt = Files.testTxt.url
+let video = Files.subdir.aVideoWithSpacesMp4.path
```
AliSoftware: Looking at the example, I'm wondering if it wouldn't be nice to have a template parameter `useExtension` (boolean) or similar (not super inspired for the name) that would allow to not use the extension in the generated Swift constant name. i.e. that would turn this line into `Files.subdir.aVideoWithSpaces.path`?

_Of course, that would generate a risk of conflicts (if there are 2 files with the same basename but only their extension being different), but it's not much different than the case of the flat template; I think many people won't have conflict and will thus prefer disabling using the extension in the constant's name, while people having conflicting file names will just keep the extension to avoid conflicts_

README.md
```diff
@@ -532,6 +534,47 @@ let mainItems = try myContext.execute(request)
 let relatedItem = myMainItem.manyToMany.first
 ```
 
+## Files
+
+```yaml
+files:
+  inputs: path/to/search
+  filter: .+\.xib$
```
AliSoftware: I'm wondering if this doesn't risk to be a confusing example, using `.xib` as the filter here. Maybe using `.+\.txt$` or `.+\.mp4$` in that example would avoid the risk of confusion with XIBs and the `ib` parser?

AliSoftware (commented):
Only took a quick look at the documentation for now (will review the rest later but it's way past bedtime here so I'll keep that for later 😅 ). Looking great 👍

AliSoftware:
One piece that we often forget to update when doing changes or adding a new parser, is updating the demo Playground. Didn't look at the whole diff yet, but if like us you forgot to update the playground, it would be nice to add a page for it too 🙂

Sources/SwiftGenKit/Parsers/Files/FilesParser.swift
```diff
@@ -0,0 +1,47 @@
...
+
+    public required init(options: [String: Any] = [:], warningHandler: Parser.MessageHandler? = nil) throws {
+      self.options = try ParserOptionValues(options: options, available: Parser.allOptions)
+      self.warningHandler = warningHandler
+    }
+
+    public static var defaultFilter: String = ".*"
+
+    public func parse(path: Path, relativeTo parent: Path) throws {
+      if path.isDirectory {
```
djbe: This check is not necessary: when you get here, you have a path to a file that matches your regex.
The actual recursion happens here:
https://github.com/SwiftGen/SwiftGen/blob/stable/Sources/SwiftGenKit/Parsers/Parser.swift#L57-L70

Sources/SwiftGenKit/Parsers/Files/FilesParser.swift
```diff
@@ -0,0 +1,47 @@
...
+    private let options: ParserOptionValues
+    var files: [File] = []
+    public var warningHandler: Parser.MessageHandler?
+
+    public required init(options: [String: Any] = [:], warningHandler: Parser.MessageHandler? = nil) throws {
+      self.options = try ParserOptionValues(options: options, available: Parser.allOptions)
+      self.warningHandler = warningHandler
+    }
+
+    public static var defaultFilter: String = ".*"
```
djbe:
```suggestion
    public static let defaultFilter: String = ".*"
```

Sources/SwiftGenKit/Parsers/Files/FilesFile.swift
```diff
@@ -0,0 +1,42 @@
...
+    let mimeType: String
+
+    init(path: Path, relativeTo parent: Path? = nil) throws {
+      guard path.exists else {
+        throw ParserError.invalidFile(path: path, reason: "Unable to read file")
+      }
+
+      self.ext = path.extension
+      self.name = path.lastComponentWithoutExtension
+      if let relative = parent.flatMap({ path.relative(to: $0) })?.parent(),
```
djbe, line 24: Can this bit be simplified to this?
```swift
let relativeParent = (parent.flatMap({ path.relative(to: $0) }) ?? path).parent()
self.path = relativeParent.components.drop { $0 == "." }
```

Also not 100% sure if that `.parent()` is needed?

Sources/SwiftGenKit/Parsers/Files/FilesFile.swift
```diff
@@ -0,0 +1,42 @@
...
+        relative != "." {
+        self.path = Array(relative.components.dropFirst())
+      } else {
+        self.path = []
+      }
+
+      if let ext = self.ext,
+        let uti = UTTypeCreatePreferredIdentifierForTag(
+          kUTTagClassFilenameExtension, ext as NSString, nil
+        )?.takeRetainedValue(),
```
djbe, line 34: Note: this must be `takeRetainedValue()`, as we are responsible for anything with `Create` or `Copy` in the function name. (so it's correct, just had to look it up because I wanted to be sure)

Sources/SwiftGenKit/Stencil/Files+Context.swift
```diff
@@ -0,0 +1,60 @@
...
+// MIT Licence
+//
+
+import Foundation
+import PathKit
+
+extension Files.Parser {
+  public func stencilContext() -> [String: Any] {
+    let files = self.files
+      .sorted { lhs, rhs in lhs.name < rhs.name }
```
djbe: May want to sort case insensitively, like:
```suggestion
      .sorted { $0.name.lowercased() < $1.name.lowercased() }
```
Note: based on strings context.

Sources/SwiftGenKit/Stencil/Files+Context.swift
```diff
@@ -0,0 +1,60 @@
...
+    entries: [Files.File],
+    atKeyPath keyPath: [String] = [],
+    usingMapper mapper: @escaping Mapper
+  ) -> [String: Any] {
+    var structuredFiles: [String: Any] = [:]
+    if let name = keyPath.last {
+      structuredFiles["name"] = name
+    }
+
+    let files = entries
```
djbe: We should add the same comments as in the strings context, this function can get confusing otherwise.

Sources/SwiftGenKit/Stencil/Files+Context.swift
```diff
@@ -0,0 +1,60 @@
...
+
+    let childEntries = entries.filter { $0.path.count > keyPath.count }
+    let children = Dictionary(grouping: childEntries) { $0.path[keyPath.count] }
+      .sorted { $0.key < $1.key }
+      .map { name, entries in
+        structure(entries: entries, atKeyPath: keyPath + [name], usingMapper: mapper)
+      }
+
+    if !children.isEmpty {
+      structuredFiles["dirs"] = children
```
djbe: Either use `children` like the strings context, or use `directories` (or `folders`?), no need for shorthand.

Documentation/templates/files/swift5.md
djbe: Would definitely be useful for this case/parser. Collision issues are up to the user.

What the default should be (on/off) is another matter 😅 It's safer with extensions (less collisions), but without gives nicer enums.

Documentation/templates/files/swift5.md
djbe: We should definitely mirror the strings templates here. We had enough trouble with the migration from `swift3` -> `flat-swift3` back in the day.

Tests/SwiftGenKitTests/FilesTests.swift
```diff
@@ -0,0 +1,55 @@
...
+    let parser = try Files.Parser()
+    try parser.searchAndParse(path: Fixtures.directory(sub: .files))
+
+    let result = parser.stencilContext()
+    XCTDiffContexts(result, expected: "defaults", sub: .files)
+  }
+
+  func testMp4s() throws {
+    let parser = try Files.Parser()
+    let path = Fixtures.directory(sub: .files)
```
djbe: This should use the `searchAndParse` function (like above) with the filter `filterRegex(forExtensions: ["mp4"])`.

Tests/Fixtures/Generated/Files/swift5/mp4s.swift
```diff
@@ -0,0 +1,71 @@
...
+    }
+    return result
+  }
+
+  internal var path: String {
+    path(locale: nil)
+  }
+
+  internal func path(locale: Locale?) -> String {
+    let bundle = BundleToken.bundle
```
djbe: I think it's easier to just do `url(locale: locale).path`?

djbe (changes requested):
Just did a general code review, haven't checked anything about the docs yet.

Documentation/SwiftGenKit Contexts/files.md
mgray88: Thanks! Writing copy is a weak point of mine 😆

Documentation/templates/files/swift5.md
mgray88: I was following along with the convention of the other templates here. They all start like that (`fonts/swift5.stencil`, etc.)

Documentation/templates/files/swift5.md
mgray88: That's a pretty good idea. I'll add it

Documentation/templates/files/swift5.md
AliSoftware: The doc format is ok, the question is more about how should we name the stencil template itself 🙂 and I agree with David, I think we should rename the template file `structured-swift5.stencil` (and of course update that line in the documentation accordingly then)

Sources/SwiftGenKit/Parsers/Files/FilesParser.swift
mgray88: Are you saying all I should need is `files.append(try File(path: path, relativeTo: parent))`? That doesn't work with the default filter of `.*`. The path passed in to `searchAndParse` then matches the folder and the recursion never happens

AliSoftware: Unrelated to how we implement recursion, I'm wondering if we should make either the filter or a parser option allow the user to opt-in or out of recursion?

E.g. when using glob patterns in shell (or ruby) you can use `foo/**/*.txt` to mean "every txt file inside the foo folder however deep in the hierarchy it is" while `foo/*.txt` will only match txt files that are directly under the foo folder. (`man fnmatch` for more info)

Obviously we can't use that here because we use `NSRegularExpressions` for our filters, not glob/fnmatch. So that won't be a solution for us. But opens the question is providing a way to have similar choice of behavior would be useful (maybe via a parser option)? wdyt?

Tests/SwiftGenKitTests/FilesTests.swift
mgray88: Hmm. That `searchAndParse` is probably what the CLI uses isn't it? This might be an issue. The reason I used that code instead was to make sure the `relativeTo` parent dir was the actual directory passed in. That's kind of important for creating a file tree relative to a specific directory. The `if/else` pass in different parent directories which throw that off

I'm not sure what the best solution would be. Have an option on the parser? Add some custom handling in the CLI code? Suggestions?

Tests/Fixtures/Generated/Files/swift5/mp4s.swift
mgray88: That's a good idea

mgray88:
RE: template names. As it currently stands the `swift5` template is structured, and `flat-swift5` is, well, flat. I think maybe you all are suggesting the same thing?

`swift5` -> `structured-swift5`
`flat-swift5` -> `flat-swift5`

Am I understanding that right?

djbe:
Exactly.

Sources/SwiftGenKit/Parsers/Files/FilesParser.swift
djbe: @mgray88 Hmmm right, that's an issue here because we don't have extensions to match on. We could change the default to `.*\..*`, to ensure we match files with extensions 🤷 (not the best option, I know)

Another option may be to add a flag to the parser protocol to indicate whether a parser only wants files or not. Could also be an option set, so a parser could say `wantFiles | wantDirectories | wantHiddenFiles`. (Ignore the actual option names, just an example). If needed, I can create a separate PR for this, and we can rebase this PR on that.

@AliSoftware not sure when users would want to disable recursion? 99% of users won't ever have to, and that 1% that has loads of files and deep directory structures, can avoid any performance impact by moving the right resources into a subfolder and point SwiftGen at that.

Tests/Fixtures/Generated/Files/swift5/mp4s.swift
mgray88: Implicit return, or explicit with swiftlint disable?

Sources/SwiftGenKit/Parsers/Files/FilesFile.swift
mgray88: Hmm. Then if `parent` was null, you could have an absolute path, no? `self.path` needs to be relative to some root directory, and shouldn't contain the filename

Tests/Fixtures/Generated/Files/swift5/mp4s.swift
djbe: implicit returns are only available in swift 5.x (can't remember which), so explicit with switlint disable.

Sources/SwiftGenKit/Parsers/Files/FilesFile.swift
djbe: Honestly dunno, it was late last night and just quickly wrote something 😅 Ignore the `.parent()` question.

I haven't checked out this branch yet, and played around with it. Maybe this? 
```swift
let relativeParent = parent.flatMap({ path.relative(to: $0) })?.parent() ?? ""
self.path = relativeParent.components.drop { $0 == "." }
```

Sources/SwiftGenKit/Parsers/Files/FilesParser.swift
mgray88: @djbe I wouldn't know where to start with that 😆. I'm sure you could probably implement that much quicker

Related also is the test below not using `searchAndParse`

Sources/SwiftGenKit/Parsers/Files/FilesFile.swift
mgray88: Put in something similar to this. Thanks!

mgray88:
Working on creating the playground, I'm thinking we may actually want to add a `relativeTo` parameter to the parser. Currently it generates the `path` relative to the shallowest path passed in. If that's not the root project directory, `url` and `path` will not be able to refer to the correct directory.

EDIT: Maybe the issues I'm encountering in Playground aren't related to that but it still seems important

Sources/SwiftGenKit/Parsers/Files/FilesParser.swift
mgray88: Just finished adding a `relativeTo` option to the Files parser, which I think might be a better option than going in and modifying the parser protocol (unless you see uses for that elsewhere)

AliSoftware:
For reference, the fonts parser should be a good inspiration as it is very similar in that sense: it lists custom font files found in a directory and generate the paths to those font files (along with info about font families and variants parsed from those TTF files); that path can be relative or absolute depending on your choice in the template, in order to match how you integrate font files in your project

Remember that when you integrate resources in your Xcode project, even if those resources are under dir1/dir2 subdirectories in the Finder relative to your project root, that doesn't really matter at runtime as once embedded into your `.ipa` all the resource files will be in the same, flat, Resources folder and the folder hierarchy you had in the Finder will *by default* not be reproduced in the `.ipa`…
That is… unless you integrate that folder as a "Create Folder References" when you add it to your Xcode project (creating a blue folder icon in Xcode). If you use folder references (blue folder icons) then the folder will be copied as-is in the `.ipa` and that's when the relative path to each resource file becomes useful.

All that to say that I'm not totally sure if that `reltiveTo` parameter will be useful to many people in Xcode projects, but now that you have it (it's a nice bonus addition just in case), I can see the fonts parser being one other parser that could benefit from that too.

mgray88:
> All that to say that I'm not totally sure if that `relativeTo` parameter will be useful to many people in Xcode projects, but now that you have it (it's a nice bonus addition just in case), I can see the fonts parser being one other parser that could benefit from that too.

😆 Yea I realized that's how Xcode projects work, but I wanted to cover all possible cases.

Also added `compact` to help with `relativeTo` if for some reason someone has a deep hierarchy, `compact` reduces the common ancestor nested `enums`

edit: I could also use a better idea for the name of `compact`. It doesn't feel correct, but I couldn't think of anything better

Tests/SwiftGenKitTests/FilesTests.swift
mgray88: Changing this to `searchAndParse`. Even though logically the output from this is not what would be desired, after adding the `relativeTo` option, this would be the designed output

mgray88:
> Remember that when you integrate resources in your Xcode project, even if those resources are under dir1/dir2 subdirectories in the Finder relative to your project root, that doesn't really matter at runtime as once embedded into your `.ipa` all the resource files will be in the same, flat, Resources folder and the folder hierarchy you had in the Finder will _by default_ not be reproduced in the `.ipa`…
> That is… unless you integrate that folder as a "Create Folder References" when you add it to your Xcode project (creating a blue folder icon in Xcode). If you use folder references (blue folder icons) then the folder will be copied as-is in the `.ipa` and that's when the relative path to each resource file becomes useful.

Hmm. While I understood this was the case, I just realized I may have skipped over actually thinking about it. I'm thinking the way the code is written should *mostly* be fine, except when someone has a hierarchy of files that is *not* a folder reference. Maybe `relativeTo` should be redesigned to work as a signifier for folder references? And if it's not set, assume files flattened to a single folder and no need for `subdirectory` parameter?

AliSoftware:
Yeh I think the best course of action is to mimic how it's done for the fonts templates, via the [preservePath](Documentation/templates/fonts/swift5.md) parameter. We should actually use the same parameter name here for files template too, for consistency.

4brunu:
Any news on this? Looking forward to this new parser 😊

mgray88:
@4brunu I've been pretty busy, but I should have some time this weekend. I don't think there's much left besides a little refactoring

4brunu:
@mgray88 thanks for your answer and your awesome work in this new parser 🙂

djbe:
Allright, so we still need to add the `preservePath` template parameter (it's not a parser option), but I got some questions about the current parser options:

I'm not sure I see the point of `relativeTo`? If a user adds a folder reference to `Folder` in Xcode, then they should configure SwiftGen to match that (point to the enclosing folder of `Folder`). Note that `preservePath` is (should be) the mechanism to differentiate between folder references & non-refs.

Also not sold on the `compact` option. I think this option (and relativeTo) can be handled by storing the file structure in the parser differently: for other parsers, we group content by things like "strings by strings table" or "scenes/segues by storyboard". In this parser we're kind of throwing away the "input folder" information, and I think we should keep it? That way, we can group files by these input folders in the output.

When decided on this, I can probably implemented the needed changes, or if you'd like, you're welcome to implement them 😉

AliSoftware:
I agree on both points.

To be clear in my mind I always thought that the preservePath parameter would _replace_ the relativeTo option and that we would indeed not need to relativeTo option anymore. Much like how we do it for fonts.

Also +1 on keeping input folder info and grouping/structuring generated code by folder

djbe:
I've implemented the filter changes I mentioned in one of my previous review comments, it was really needed (I think) because the old code would just go through all the files/children of a folder without applying the filter anymore. Does this matter for basic extension regexes? No. But it might've mattered for more complex cases 🤷 

As discussed I've applied the other changes as well. The end result is that this parser is kind of like a mix of the Fonts and Strings parsers.

Future PRs could add extra accessors/methods depending on the file mime-type, a bit like the XCAssets templates right now.

Documentation/templates/files/flat-swift4.md
```diff
@@ -0,0 +1,52 @@
+## Template Information
+
+| Name      | Description       |
+| --------- | ----------------- |
+| File name | files/flat-swift5.stencil |
```
mgray88:
```suggestion
| File name | files/flat-swift4.stencil |
```

Documentation/templates/files/flat-swift4.md
```diff
@@ -0,0 +1,52 @@
...
+| Configuration example | <pre>files:<br />  inputs: dir/to/search/with/filter<br />  filter: .+\.*$<br />  outputs:<br />    templateName: flat-swift5<br />    output:Files.swift</pre> |
+| Language | Swift 5 |
```
mgray88:
```suggestion
| Language | Swift 4 |
```

Documentation/templates/files/flat-swift4.md
```diff
@@ -0,0 +1,52 @@
+## Template Information
+
+| Name      | Description       |
+| --------- | ----------------- |
+| File name | files/flat-swift5.stencil |
+| Configuration example | <pre>files:<br />  inputs: dir/to/search/with/filter<br />  filter: .+\.*$<br />  outputs:<br />    templateName: flat-swift5<br />    output:Files.swift</pre> |
```
mgray88:
```suggestion
| Configuration example | <pre>files:<br />  inputs: dir/to/search/with/filter<br />  filter: .+\.*$<br />  outputs:<br />    templateName: flat-swift4<br />    output:Files.swift</pre> |
```

mgray88:
@djbe I'm glad you were able to get these changes in, work's been hectic on my end. I think it definitely looks much better. I did notice a few things in the new swift4 docs that still referenced swift5, or other files they were probably copied from

Documentation/templates/files/flat-swift4.md
```diff
@@ -0,0 +1,52 @@
...
+| Language | Swift 5 |
+| Author | Mike Gray |
+
+## When to use it
+
+- When you need to generate *Swift 5* code.
```
mgray88:
```suggestion
- When you need to generate *Swift 4* code.
```

Documentation/templates/files/structured-swift4.md
```diff
@@ -0,0 +1,53 @@
+## Template Information
+
+| Name      | Description       |
+| --------- | ----------------- |
+| File name | files/structured-swift5.stencil |
```
mgray88:
```suggestion
| File name | files/structured-swift4.stencil |
```

Documentation/templates/files/structured-swift4.md
```diff
@@ -0,0 +1,53 @@
...
+| Configuration example | <pre>files:<br />  inputs: dir/to/search/with/filter<br />  filter: .+\.*$<br />  outputs:<br />    templateName: structured-swift5<br />    output:Files.swift</pre> |
```
mgray88:
```suggestion
| Configuration example | <pre>files:<br />  inputs: dir/to/search/with/filter<br />  filter: .+\.*$<br />  outputs:<br />    templateName: structured-swift4<br />    output:Files.swift</pre> |
```

Documentation/templates/files/structured-swift4.md
```diff
@@ -0,0 +1,53 @@
...
+| Language | Swift 5 |
```
mgray88:
```suggestion
| Language | Swift 4 |
```

Documentation/templates/files/structured-swift4.md
```diff
@@ -0,0 +1,53 @@
...
+| Author | Mike Gray |
+
+## When to use it
+
+- When you need to generate *Swift 5* code.
```
mgray88:
```suggestion
- When you need to generate *Swift 4* code.
```

Documentation/templates/files/structured-swift4.md
```diff
@@ -0,0 +1,53 @@
...
+  internal static let testTxt = File(name: "test", ext: "txt", path: "", mimeType: "text/plain")
+  /// subdir/
+  internal enum Subdir {
+    /// subdir/A Video With Spaces.mp4
+    internal static let aVideoWithSpacesMp4 = File(name: "A Video With Spaces", ext: "mp4", path: "subdir", mimeType: "video/mp4")
+  }
+}
+```
+
+[Full generated code](../../../Tests/Fixtures/Generated/Files/swift5/defaults.swift)
```
mgray88:
```suggestion
[Full generated code](../../../Tests/Fixtures/Generated/Files/swift4/defaults.swift)
```

Documentation/templates/files/flat-swift4.md
```diff
@@ -0,0 +1,52 @@
...
+```swift
+internal enum Files {
+  /// test.txt
+  internal static let testTxt = File(name: "test", ext: "txt", path: "", mimeType: "text/plain")
+  /// subdir/A Video With Spaces.mp4
+  internal static let aVideoWithSpacesMp4 = File(name: "A Video With Spaces", ext: "mp4", path: "subdir", mimeType: "video/mp4")
+}
+```
+
+[Full generated code](../../../Tests/Fixtures/Generated/Files/flat-swift5/defaults.swift)
```
mgray88:
```suggestion
[Full generated code](../../../Tests/Fixtures/Generated/Files/flat-swift4/defaults.swift)
```

4brunu:
Any news on this? Looking forward to it 🙂

djbe (approved):

PR #737, opened 2020-06-29
Template `bundle` parameter

djbe:
As discussed in https://github.com/SwiftGen/SwiftGen/pull/716#issuecomment-643852541:
> Add a template parameter (to all templates that support) to override the bundle, which will default to `BundleToken.bundle`. In essence, what #644 should have been.

Applies to:
-  Fonts
- IB
- JSON
- Plist
- Strings
- XCAssets


SwiftGen-Eve:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

---

</p>

Documentation/templates/fonts/swift4.md
```diff
@@ -17,6 +17,7 @@ You can customize some elements of this template by overriding the following par
 
 | Parameter Name | Default Value | Description |
 | -------------- | ------------- | ----------- |
+| `bundle` | `BundleToken.bundle` | Allows you to set from which bundle font files are loaded from. By default, it'll point to the same bundle as where the generated code is. |
```
AliSoftware, line 20: 👍

Documentation/templates/strings/flat-swift4.md
```diff
@@ -18,9 +18,10 @@ You can customize some elements of this template by overriding the following par
...
 | `enumName` | `L10n` | Allows you to change the name of the generated `enum` containing all string tables. |
 | `forceFileNameEnum` | N/A | Setting this parameter will generate an `enum <FileName>` _even if_ only one FileName was provided as input. |
-| `lookupFunction` | N/A | Allows you to set your own custom localization function. The function needs to have as signature: `(key: String, table: String) -> String`. The parameters of your function can have any name (or even no external name), but if it has named parameters, you must provide the complete function signature, including those named parameters – e.g. `yourFunctionName(forKey:table:)`. |
+| `lookupFunction` | N/A | Allows you to set your own custom localization function. The function needs to have as signature: `(key: String, table: String) -> String`. The parameters of your function can have any name (or even no external name), but if it has named parameters, you must provide the complete function signature, including those named parameters – e.g. `yourFunctionName(forKey:table:)`. Note: overrides `bundle` parameter. |
```
AliSoftware: Not sure if "overrides" is the right term here 🤔  Maybe "If you define this parameter, the bundle parameter will be ignored"?

Tests/Fixtures/Generated/Fonts/swift4/compilation-configuration.yml
```diff
@@ -3,3 +3,7 @@ common:
   swift_versions:
     - 4
     - 4.2
+file_specific:
+  defaults-customBundle.swift:
+    files:
+      - Common-CustomBundle.swift
```
AliSoftware, line 9: Side note: interesting, this file was hidden (due to `linguist-generated` but it feels it would be valuable if it weren't, as online for generated outputs and stencils that one (1) won't change that often so we definitively want to look deeper when it does, and (2) isn't really generated, unlike outputs and stencils… 🤔

Tests/Fixtures/Generated/IB-iOS/scenes-swift4/compilation-configuration.yml
```diff
@@ -11,6 +11,10 @@ file_specific:
   all.swift:
     files:
       - IB-iOS-Definitions.swift
+  all-customBundle.swift:
+    files:
+      - IB-iOS-Definitions.swift
+      - Common-CustomBundle.swift
```
AliSoftware, line 17: (for example I almost missed this one because I was scrolling past the various other `Generated/*/<template>/<variant>.swift` files in that diff and it was in the middle of it)

Tests/TemplatesTests/FontsTests.swift
```diff
@@ -21,6 +21,10 @@ final class FontsTests: XCTestCase {
         context: context,
         suffix: ""
       ),
+      (
+        context: try StencilContext.enrich(context: context, parameters: ["bundle=ResourcesBundle.bundle"]),
```
AliSoftware, line 25: 💭  Oh, for some reason I thought I remembered that the parameters were supposed to be passed using `:` for key/value separator, was it `=` all this time? (I mean, since now the config file is the primary way to pass parameters and that this one uses YAML and thus `:`, maybe that's where I thought that from, not sure why we picked `=` in the first place back then

rakelib/CompilationConfiguration.rb
```diff
@@ -45,6 +45,12 @@ def self.remove_swiftdoc()
       FileUtils.rm_rf(file)
     end
   end
+
+  def self.remove_swiftsourceinfo()
+    Dir.glob("#{MODULE_OUTPUT_PATH}/*/*.swiftsourceinfo").each do |file|
```
AliSoftware, line 50: Is that a new type of file since Xcode 11 or something? (Wondering if maybe we could avoid generating it in the first place by passing some special arg when we run `xcodebuild`? tho if there is, maybe there's a similar one for swiftdoc files too 🤔 )

AliSoftware (commented):
Looking good!

Just one nit about documentation ~and an open question about swiftsourceinfo files~.

---

I also think it would be a nice bonus to add a doc in `Documentation/Articles` about how to implement the common "fallback mechanism" for devs who needs one in their project but don't want to go full-custom-templates-everywhoere for that. And then link to that article somewhere in each template's docs.

What I mean by that idea is something like this:
```swift
class BundleLookup {
  let bundles: [Bundle]

  func url(forResource name: String, withExtension ext: String) -> URL? {
    // find the first bundle which returns a non-nil resource URL
    bundles.lazy.compactMap {
      $0.url(forResource: name, withExtension: ext)
    }.first
  }
  // … you can apply the same logic for other Bundle methods for which you want that lookup/fallback mechanism

  // also provide some constructors for common scenarios
  static let mainThenRsrc = BundleLookup(bundles: [.main, ResourcesBundle.bundle])
}
```
Then passing `BundleLookup.mainThenRsrc` for the `bundle` parameter in SwiftGen's config

That could be considered an "advanced trick" (thus why it would be in `Documentation/Articles`) but would be quite a useful one imho especially given that that's the sort of trick I had to implement in the past in multiple projects (especially white-label ones)

rakelib/CompilationConfiguration.rb
AliSoftware: Ok seems it's always generated no matter what: https://forums.swift.org/t/proposal-emitting-source-information-file-during-compilation/28794 so might as well keep the deletion code in here indeed.

djbe:
@AliSoftware I think that that `LookupBundle` example would better fit the `lookupFunction` PR, where you can go full custom stuff if you want. The intent of the `bundle` override in this PR is for most podspec users (for example) that need to point to some resource bundle.
- Simply want to change bundle from `BundleToken.bundle` to some other `Bundle`: use `bundle` parameter
- Fully custom loading mechanism, that for example checks multiple bundles for loading a resource: use `lookupFunction` parameter

Of course, nothing stops users from doing advanced tricks like above, where they know  which function is used to load a resource from a bundle, and abuse that information 🤷 But I'd avoid actually **documenting** that fact. Point them to `lookupFunction` instead.

Note that I agree that an article explaining `lookupFunction` would be helpful. We can even explain the 2 points I mentioned here in that article.

AliSoftware:
Right. My initial thoughts were that the problem with `lookupFunction` is that it doesn't work for all templates; especially for xcassets.

But looking at that template, using a `LookupBundle` won't solve it either after all, because in those templates we don't call a method on the `bundle`, instead we use the `bundle` as a parameter to UIKit/AppKit method calls – like `UIColor(named: asset.name, in: bundle, compatibleWith: nil)` – so that trick wouldn't even work after all… 🤷

djbe:
Jep. Either we add a bunch of `lookupFunction` parameters to `xcassets` at some point, or we leave it up to the users for custom templates. And even then, I'm not 100% sure `lookupFunction` would work everywhere in `xcassets`...

templates/fonts/swift5.stencil
```diff
@@ -70,8 +70,8 @@
   }
 
   fileprivate var url: URL? {
-    let bundle = BundleToken.bundle
-    return bundle.url(forResource: path, withExtension: nil)
+    // swiftlint:disable:next implicit_return
+    return {{param.bundle|default:"BundleToken.bundle"}}.url(forResource: path, withExtension: nil)
```
AliSoftware, line 74: I'm guessing you didn't use implicit_return in order for the template to be compatible with Swift 5.0 (since they were added in 5.1); was wondering if there were a risk of other places where we might have used implicit returns making our generated code not compatible with 5.0?

I'm not entirely sure to remember how our `compilation-configuration.yml` files work but if we have `swift_versions: [5]` in it, I think it just uses a `SWIFT_VERSION` of `5` in Xcode… which only uses the latest Swift 5.x sub-version shipped with that Xcode SDK, so currently 5.2, and doesn't test with Swift 5.0? If that's the case, that means we could easily have missed other places and 5.1+-only features 🤔

AliSoftware (approved):
Looking good!

templates/fonts/swift5.stencil
djbe: Well, yeah, exactly. Thing is, we can't test it, unless we add something on Circle to run it on older Xcode versions, but then... Probably bunch of other issues.

Luckily, SwiftLint helps us out here. It wants us to remove the `return` because we can in Swift 5.1+, but **we** can't, so it we disable the rule here. If we at some point fix the code, then SwiftLint will warn us about a useless disable.

AliSoftware: Ah, good point. Perfect then, that's reassuring 🙂

PR #738, opened 2020-06-29
Template `lookupFunction` parameter

djbe:
As discussed in https://github.com/SwiftGen/SwiftGen/pull/716#issuecomment-643852541:
> Adapt the `lookupFunction` parameter to other templates that support it: fonts, IB scenes, json, and plist.
> For now, we won't adapt `lookupFunction` to the assets templates, as we'd need multiple lookup functions (one for each type).

Applies to:
-  Fonts
- IB
- JSON
- Plist

PR #739, opened 2020-06-29
Small doc & template tweaks #trivial

djbe:
Made these initially for #737 and #738, but they might as well be merged for 6.2.1.
- Template docs: sort parameters alphabetically for easier lookup
- Template docs: clarify the strings `lookupFunction` signature explicitely, mainly useful later in #738 where lookup functions have other return types.
- IB: don't generate bundletoken for segue templates, it isn't used
- Strings: don't generate bundletoken if `lookupFunction` is provided, as it isn't used.


SwiftGen-Eve:

---

</p>

Documentation/templates/strings/structured-swift4.md
```diff
@@ -25,10 +25,10 @@ You can customize some elements of this template by overriding the following par
 | Parameter Name | Default Value | Description |
 | -------------- | ------------- | ----------- |
 | `enumName` | `L10n` | Allows you to change the name of the generated `enum` containing all string tables. |
+| `forceFileNameEnum` | N/A | Setting this parameter will generate an `enum <FileName>` _even if_ only one FileName was provided as input. |
+| `lookupFunction` | N/A | Allows you to set your own custom localization function. The function needs to have as signature: `(key: String, table: String) -> String`. If the function has named parameters, you must provide the complete function signature, including those named parameters – e.g. `yourFunctionName(forKey:table:)`. |
```
AliSoftware: A bit confusing imho that you both say "signature is `(key: String, table: String)`" and also "if function has named parameters. It's not clear if the parameters need to explicitly have the exact `key:` and `table:` parameter names or not, as those two sentences seems a bit contradictory, one suggesting they params must be named that way and the other suggesting that the function can decide to not have named params…

```suggestion
| `lookupFunction` | N/A | Allows you to set your own custom localization function. The function expects the key and table as parameters (in that order) and needs to have as signature: `(String, String) -> String`. If the function has named parameters, you must provide the complete function signature, including those named parameters – e.g. `yourFunctionName(forKey:table:)`. |
```

AliSoftware (commented):
Apart from the wording about `lookupFunction`, the rest looks good to me 👍 

I think it could even deserve a CHANGELOG entry for the `BundleToken` removed from IB templates

Documentation/templates/strings/structured-swift4.md
djbe: Reason I changed the wording like that is to give it a clear meaning. They're both `String`s, but the first one is the localization key, the 2nd one is the table key. It also unifies the documentation for other `lookupFunction`s (in #738), where we can have signatures like `(name: String, family: String, path: String) -> URL?` for the fonts templates.

djbe:
Note that BundleToken was removed from the **segue** templates, not the scene templates. It wasn't used **at all** in those templates, and wasn't accessible from outside that file either (`private`). So it was just a bit of useless code that nobody could use.

Documentation/templates/strings/structured-swift4.md
AliSoftware: Right. Then maybe mention that named labels for parameters are optional?

```suggestion
| `lookupFunction` | N/A | Allows you to set your own custom localization function. The function needs to have as signature: `(key: String, table: String) -> String`. The parameters of your function can have any name (or even no external name), but if it has named parameters, you must provide the complete function signature, including those named parameters – e.g. `yourFunctionName(forKey:table:)`. |
```

AliSoftware:
Ah right, forgot that they were private so there's no risk anyone was using them. #trivial it is then 😉

AliSoftware (approved):

PR #763, opened 2020-08-12
Use SPM `Bundle.module` if available

arnauddorgans:
The built-in templates will now correctly handle Swift Package Manager resources, using `Bundle.module` if it's available. As before, you can override the used `Bundle` using the `bundle` or `lookupFunction` template parameters. 

https://github.com/apple/swift-evolution/blob/master/proposals/0271-package-manager-resources.md#runtime-access-to-resources-bundle

---


CHANGELOG.md
```diff
@@ -10,7 +10,9 @@ _None_
 
 ### New Features
 
-_None_
+* All templates with bundle have now a new `spmBundle` parameter which allows usage of `Bundle.module` instead of `Bundle(for: BundleToken.self)` for SPM packages.
```
  <tbody>
    <tr>
      <td>:warning:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/SwiftGen/SwiftGen/blob/e07695b6e8f8cff2bc7769bba89d931152600121/CHANGELOG.md#L13"></span>Line describing your entry should end with a period and 2 spaces.</td>
    </tr>
  </tbody>
</table>

SwiftGen-Eve:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

---

</p>

djbe:
Do we "need" a parameter for this? Can't we just always generate that code (as long as users aren't providing a custom bundle/lookupFunction)?

CHANGELOG.md
```diff
@@ -10,7 +10,9 @@ _None_
 
 ### New Features
 
-_None_
+* Add SwiftPM `Bundle.module` compatibility to all templates.
```
```suggestion
* The built-in templates will now correctly handle Swift Package Manager resources, using `Bundle.module` if it's available. As before, you can override the used `Bundle` using the `bundle` or `lookupFunction` template parameters.  
```

Not 100% sure this is the right terminology in the 1st sentence 🤷

djbe (changes requested):
Cool, thx for the quick update! I think we can slightly better explain the change in the changelog.

Ideally we should test this, but not a clue how to actually do it... We'd have to define a package for each test with the right resources, compile it and test it? 😱

CHANGELOG.md
arnauddorgans: This is better thanks 👍 

We could test each source with a SwiftPM 5.3 bundle but, doing this, I guess it will be resumed to test if SwiftPM is working and not `SwiftGen`? 🤔

djbe: Hmmm what we could do is test the output once without that flag defined, and once with SWIFT_PACKAGE defined, to ensure both versions compile. I think that involves a slight refactor to the test yaml files though 🤔 Stuff for another PR for sure.

arnauddorgans: OK but `Bundle.module` is generated only when `Package.swift` has defined `resources`.
So even if `SWIFT_PACKAGE` is defined we may have a compiler error when `Package.swift` has no `resources` (it make sense because the bundle can not exist without resources).

So to test it we will have to create a package with defined resources first to generate the `Bundle.module` boilerplate code.

djbe: No need for that actually, we can perfectly fake it like we do in other tests, just include a bit of code that defines a static `let` on `Bundle` called `module`. That part is easy, and we support it already.

The bit that'll be a bit more involved, is allowing a file to be compiled multiple times with different configurations. A compilation configuration like this one:
https://github.com/SwiftGen/SwiftGen/blob/stable/Tests/Fixtures/Generated/IB-iOS/scenes-swift5/compilation-configuration.yml
Will need to be changed somehow to allow variations per file (the keys in `file_specific`).

Like I said, issue for a follow-up ticket/PR.

djbe (approved):

PR #768, opened 2020-08-18
Colors: support parsing colors in ARGB notation

### LINKED ISSUE #562: Colors: Support for Android's color.xml format
I notice that #15 declare support for Android colors.xml, but when i use Android's xml from our team, i catch a error: "error: Invalid hex color "@color/theme_cyan" found for key "colorPrimary" (colors.xml)", because the xml use this format "<color name="colorPrimary">@color/theme_cyan</color>", maybe it's new for Android Studio support, can SwiftGen deal with this situation? i just see parse xml's value all use `Colors.parse(hex: value, key: name, path: path)`, or it's plan for support?

kevinstier:
This PR adds support for the default Android color format (ARGB) to the XML color parser. Fixes #562


SwiftGen-Eve:

---

</p>

djbe:
I think we should go with a more generic option name, such as `colorFormat` (or other name, naming is hard). That option would then accept something like `rgba` (default) or `argb`.

It would keep our options open for other formats, such as non-hex formats, wide colors, etc...

kevinstier:
@djbe I've added the support for `colorFormat`. Can you review my changes?

djbe (changes requested):
Nice, thanks for the update.
The only thing left is documentation, always a bit of a chore.

In `Documentation/Parsers/colors.md`: Add  a table in the "customization" to document the option, with all available values. You can base yourself on the strings parser, which also defines a parser option.
I don't think there are any other documentation/example files that need any updates.

djbe (approved):

PR #770, opened 2020-08-23
Added command to generate input and output xcfilelist

### LINKED ISSUE #441: [New Feature] Generate xcfilelist for Script Build Phases
## Context

In [WWDC'18 Session #408](https://developer.apple.com/videos/play/wwdc2018/408/) ("Building Faster in Xcode"), they present a new feature of Xcode 10 where you can provide `.xcfilelist` files to specify the list of input and output files for the build phase, which will allow Xcode to better determine the dependency graph between targets, files having to be rebuilt, etc.

SwiftGen Build Phase in Xcode 10, with File Lists sections

> Note: Input & Output files for build phases were already present in previous versions of Xcode to determine when to run the build phases and optimise the build dependency resolution; what's new in Xcode 10 is the ability to put the list of those input/output files in `xcfilelist` — that can then be generated/updated by external tools without having to modify the xcodeproj for that

## Feature Description

It would be cool to make a new `swiftgen xcfilelist` command that would read the config file and generate an `swiftgen-input-files.xcfilelist` and `swiftgen-output-files.xcfilelist` according to the input and output files listed/used by the `swiftgen.yml`

Obviously the users would still have to:

1. Configure their Script Build Phase invoking `"$(PODS_ROOT)/SwiftGen/bin/swiftgen` to also use `swiftgen-input-files.xcfilelist` and `swiftgen-output-files.xcfilelist` as Input/Output File Lists reps. — But that would be only necessary once, when you configure the Script Build Phase during your installation of SwiftGen into your project
2. Run `swiftgen xcfilelist` manually every time they update their `swiftgen.yml` file — because (see 13'50 in the WWDC video) Xcode won't allowed us to modify `xcfilelist` _during_ the build so we wouldn't be able to make `swiftgen` (aka `swiftgen config run`) generate those `xcfilelist` in addition to generating the other outputs.

> That second point is to investigate further though, because it could still be interesting that `swiftgen` would in fact — like what we do for other generated files — only re-generate-and-override the `xcfilelist` file if the new content is different than the old one, and not touching the `xcfilelist` file if nothing has changed — which could lead to Xcode being happy when you didn't change your config file… and probably failing with an (hopefully explicit) error when your xcfilelists are out-of-sync and have been re-generated, which could actually be a good thing…?

## Suggested CLI

```sh
# Explicit path to files to generate
$ swiftgen xcfilelist --inputs Resources/swiftgen-inputs.xcfilelist --outputs Resources/swiftgen-output.xcfilelist

# Default paths when not specified anywhere: defaults to --inputs ./swiftgen-input-files.xcfilelist --outputs ./swiftgen-output-files.xcfilelist
$ swiftgen xcfilelist 

# Ability to also list the paths in swiftgen.yml to avoid repeating them?
$ swiftgen xcfilelist # uses values from xcfilelists.inputs and xcfilelists.outputs keys from config file
```

> The last mode is debatable, because if the user indeed does have to run `swiftgen xcfilelist` manually — so assuming we don't re-generate the xcfilelists every time we run swiftgen but only if the content has changed, and we are fine with Xcode failing if the file lists have changed, etc — then having the an `xcfilelists` entry in the `swiftgen.yml` to list those paths would likely make most people assume that, because it's in the config file, it will be executed every time, same as all other commands listed in that same config file… and thus if we do that I foresee a lot of false-positive issues/bug reports coming…

## Example of generated files

```sh
# swiftgen-input-files.xcfilelist

# xcassets
$(SRCROOT)/MyApp/Resources/Images.xcassets
$(SRCROOT)/MyApp/Resources/Logos.xcassets

# fonts
$(SRCROOT)/MyApp/Resources/Fonts/MyFont.ttf
$(SRCROOT)/MyApp/Resources/Fonts/MyOtherFont.ttf
```

```sh
# swiftgen-output-files.xcfilelist

# xcassets
$(SRCROOT)/MyApp/Generated/Images-generated.swift

# fonts
$(SRCROOT)/MyApp/Generated/Fonts-generated.swift
```

CraigSiemens:

---

Added `xcfilelist` subcommand to generated input and output file lists to be used with Xcode's Run Script build phase.  Fixes #441


SwiftGen-Eve:

---

</p>

Sources/SwiftGen/Commander/XCFileListCommands.swift
```diff
@@ -0,0 +1,94 @@
...
+      "config",
+      default: "swiftgen.yml",
+      flag: "c",
+      description: "Path to the configuration file to use",
+      validator: checkPath(type: "config file") { $0.isFile }
+    )
+
+    static let inputs = Option<Path>(
+      "inputs",
+      default: "./swiftgen-input.xcfilelist",
```
djbe: I wonder if we should even provide defaults. Pretty sure everyone will have different paths for these. May be better to instead leave these empty, and throw an error if not provided?

Sources/SwiftGen/Commander/XCFileListCommands.swift
```diff
@@ -0,0 +1,94 @@
...
+      entry.makeRelativeTo(inputDir: config.inputDir, outputDir: config.outputDir)
+      pathsByCommand[command.name, default: []].formUnion(paths(entry))
+    }
+
+    return pathsByCommand
+  }
+
+  private static func xcFileList(for pathsByCommand: [String: Set<Path>]) -> String {
+    pathsByCommand
+      .sorted { $0.key < $1.key }
```
djbe: I'd actually move the sort to **after** the map, so that paths are sorted in the file list?

Sources/SwiftGen/Commander/XCFileListCommands.swift
```diff
@@ -0,0 +1,94 @@
...
+      let config = try Config(file: configPath)
+
+      try writeInputsXCFileList(for: config, to: inputPath)
+      try writeOutputsXCFileList(for: config, to: outputPath)
+    }
+  }
+
+  private static func writeInputsXCFileList(for config: Config, to path: Path) throws {
+    let inputPaths = pathsByCommand(for: config) { entry -> [Path] in
+      entry.inputs.map { (config.sourcePath + $0).absolute() }
```
djbe: This will not work, as parsers can recurse into folders and match paths with regexes (in contrast with output paths, which are well defined).

Maybe add a mechanism to each parser to get the list of paths that they are going to load? Might be sufficient to add this to the `Parser` protocol (haven't checked).

If we add this mechanism to `Parser`, we may want to do the same for `output` paths, so that the path calculation logic doesn't have to exist here.

Sources/SwiftGen/Commander/XCFileListCommands.swift
```diff
@@ -0,0 +1,94 @@
...
+    let content = xcFileList(for: outputPaths)
+    try path.write(content)
+  }
+
+  private static func pathsByCommand(for config: Config, paths: (ConfigEntry) -> [Path]) -> [String: Set<Path>] {
+    var pathsByCommand: [String: Set<Path>] = [:]
+
+    for (command, var entry) in config.commands {
+      entry.makeRelativeTo(inputDir: config.inputDir, outputDir: config.outputDir)
+      pathsByCommand[command.name, default: []].formUnion(paths(entry))
```
djbe: Rename that `paths` to something more meaningful to its actual purpose. `extractPaths`? Dunno.

Sources/SwiftGen/main.swift
```diff
@@ -47,6 +47,8 @@ let main = Group {
     }
   }
 
+  $0.addCommand("xcfilelist", "generates xcfilelists based on the configuration file ", XCFileListCLI.generate)
```
djbe:
```suggestion
  $0.addCommand("xcfilelist", "generates xcfilelists based on the configuration file", XCFileListCLI.generate)
```

Not sure if this should be renamed to something like `generate-xcfilelists` to clarify it's purpose. Or a command + subcommand such as `xcfilelist generate`? To better match the structure of the other groups.

djbe (changes requested):
Love that this is getting implemented 🎉 

Some general feedback first:
- Decide on the name / structure of the command
- The path building logic may not belong here (move to `Parser`)
- Needs documentation (readme, other places?). **Don't** start on it yet before we finish the rest of the PR.

Sources/SwiftGen/Commander/XCFileListCommands.swift
CraigSiemens: This is sorting by the command name first and below there is another sort by the path.

djbe: Right but I'm thinking of the end result: a file with a list of paths, which to the user will look like an unsorted list of paths. If we should sort anything, it should be the whole list of paths.

Sources/SwiftGen/Commander/XCFileListCommands.swift
CraigSiemens: This seems like a bigger change and I might not have time to do it until later in the week. 

I did take a quick look at it and `Parser` doesn't feel like the right place for it (but I. am new to the codebase). I was thinking maybe `ConfigEntry` would make sense since it already contains the `inputs` and `filter`. Maybe there could be a `filteredInputPaths` variable that returns an array of matching paths. How does that sound?

djbe: So, I've checked the codebase. The logic for searching & parsing is already in the `Parser` protocol:
https://github.com/SwiftGen/SwiftGen/blob/stable/Sources/SwiftGenKit/Parsers/Parser.swift#L57-L70

We can copy (or extract) the logic for recursing & matching a filter, into a separate property (or function) so we can use it here.

Sources/SwiftGen/main.swift
```diff
@@ -47,6 +47,10 @@ let main = Group {
     }
   }
 
+  $0.group("xcfilelist", "manage xcfilelists") {
```
djbe: Did a quick check of the current list of groups/commands (should have done this before 🤦 ), maybe it'd be better under the `config` group? Like `swiftgen config generate-xcfilelists`?

Sources/SwiftGen/main.swift
```diff
@@ -47,6 +47,10 @@ let main = Group {
...
+    $0.addCommand("generate", "generates xcfilelists based on the configuration file", XCFileListCLI.generate)
```
djbe:
```suggestion
    $0.addCommand("generate", "generates xcfilelists based on the configuration file, for use in an Xcode build step that executes `swiftgen config run`", XCFileListCLI.generate)
```

Sources/SwiftGen/Commander/XCFileListCommands.swift
```diff
@@ -0,0 +1,106 @@
...
+      if let inputPath = inputPath {
+        try writeInputsXCFileList(for: config, to: inputPath)
+      }
+
+      if let outputPath = outputPath {
+        try writeOutputsXCFileList(for: config, to: outputPath)
+      }
+
+      if inputPath == nil, outputPath == nil {
+        logMessage(.error, "An input or output path must be provided.")
```
djbe:
```suggestion
        logMessage(.error, "You must provide the path of an input or output xcfilelist (or both).")
```
Maybe something like this? Just to avoid confusion because "input" and "output" can mean a lot of things.

djbe (changes requested):

Sources/SwiftGen/Commander/XCFileListCommands.swift
CraigSiemens: The output file is arranged with the paths grouped by command name. It was based on the suggestion in the original issue.
```
# fonts
$(SRCROOT)/MyApp/Generated/Fonts-generated.swift

# xcassets
$(SRCROOT)/MyApp/Generated/Images-generated.swift
```

Sources/SwiftGenKit/Parsers/Parser.swift
```diff
@@ -55,16 +55,33 @@ public extension Parser {
...
+    for (path, parentDir) in try Self.subpaths(in: path, matching: filter) {
+      try parse(path: path, relativeTo: parentDir)
+    }
+  }
+
+  /// Recursively search through the given paths, returning any files or folders that matches the given filter.
+  ///
+  /// - Parameter paths: The paths to search recursively through.
+  /// - Parameter filter: The filter to apply to each path.
+  static func subpaths(in paths: [Path], matching filter: Filter) throws -> [(path: Path, parentDir: Path)] {
```
djbe, line 68: Don't forget to document the return type (same below)

Sources/SwiftGen/Commander/XCFileListCommands.swift
djbe: Completely missed that bit 🤦 apologies!

djbe:
I know the original issue #441 proposed an output format of:
```
# fonts
...

# xcassets
...
```

But do we know for sure that `xcfilelist`s support comments? I've tried searching for documentation on this, but couldn't find any.

Sources/SwiftGen/Commander/ConfigCommands.swift
```diff
@@ -80,11 +80,42 @@ enum ConfigCLI {
...
   static let doc = command {
     let docURL = gitHubDocURL(version: Version.swiftgen, path: "ConfigFile.md")
     logMessage(.info, "Open documentation at: \(docURL)")
     NSWorkspace.shared.open(docURL)
   }
+
+  // MARK: Generate XCFileLists
+
+  // swiftlint:disable:next line_length
+  static let generateXCFileListsDescription = "generates xcfilelists based on the configuration file, for use in an Xcode build step that executes `swiftgen config run`"
```
djbe: I'd move this into `main.swift` like the rest.

Sources/SwiftGen/main.swift
```diff
@@ -32,6 +32,7 @@ let main = Group {
     $0.addCommand("run", "run commands listed in the configuration file", ConfigCLI.run)
     $0.addCommand("init", "create an initial configuration file", ConfigCLI.create)
     $0.addCommand("doc", "open the documentation for the configuration file on GitHub", ConfigCLI.doc)
+    $0.addCommand("generate-xcfilelists", ConfigCLI.generateXCFileListsDescription, ConfigCLI.generateXCFileLists)
```
djbe: Put the description here directly (for consistency). You can use multiline strings to avoid line length warnings, like so:
```swift
"""
Lorem ipsum dolor sit amet, consectetur adipiscing elit, \
sed do eiusmod tempor incididunt ut labore 
"""
```
The `\` avoids a newline.

CHANGELOG.md
```diff
@@ -10,7 +10,9 @@ _None_
 
 ### New Features
 
-_None_
+* Added `xcfilelist` subcommand to generated input and output file lists to be used with Xcode's Run Script build phase.
```
djbe: This needs to be updated with the new command name, and I'd add the explanation from the description, that mentions it's meant for phases that execute `swiftgen config run`.

Something like:
```suggestion
* Added `config generate-xcfilelist` subcommand to generated input and output `xcfilelists` based on your configuration file. This should be used in an Xcode build step that executes `swiftgen config run`.  
```

CraigSiemens:
I cant find any official docs about comments in `xcfilelist`, but using `#` for comments is defined in the `xclangspec` bundled in Xcode.

`Xcode.app/Contents/SharedFrameworks/SourceModel.framework/Versions/A/Resources/LanguageSpecifications/xcfilelist.xclangspec`

Sources/SwiftGen/Config/Config+XCFileList.swift
```diff
@@ -0,0 +1,63 @@
...
+    for (command, var entry) in commands {
+      entry.makeRelativeTo(inputDir: inputDir, outputDir: outputDir)
+      let paths = try pathsForEntry(command, entry)
+      pathsByCommand[command.name, default: []].formUnion(paths)
+    }
+
+    return pathsByCommand
+  }
+
+  private func xcFileList(for pathsByCommand: [String: Set<Path>]) -> String {
```
djbe, line 58: I'd add a small doc comment here about the output, eg.:
```suggestion
  /// Generates `xcfilelist` with all the paths, grouped per command into sections, like:
  ///
  ///     # fonts
  ///     …
  ///
  ///     # xcassets
  ///     …
  private func xcFileList(for pathsByCommand: [String: Set<Path>]) -> String {
```

djbe (changes requested):
Some small nitpicks

CHANGELOG.md
```diff
@@ -10,7 +10,9 @@ _None_
 
 ### New Features
 
-_None_
+* Added `config generate-xcfilelist` subcommand to generated input and output `xcfilelists` based on your configuration file. This should be used in an Xcode build step that executes `swiftgen config run`.
```
  <tbody>
    <tr>
      <td>:warning:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/SwiftGen/SwiftGen/blob/dc8c7910f181388fa82098adb2e08765e00e778a/CHANGELOG.md#L13"></span>Line describing your entry should end with a period and 2 spaces.</td>
    </tr>
  </tbody>
</table>

djbe (approved):

PR #774, opened 2020-09-14
Strings: fix multi-line translations in comments

mrackwitz:
Hey there, I saw some discussion in earlier issues around this, but I'm not quite sure what's the last take on multi-line strings.

## Problem 

What I do know is, that using `structured-swift5` generated invalid Swift code, bc the additional lines of my multi-line translation made it into the comment, the newlines were not stripped out, so the resulting Swift code is invalid and fails to compile.
This change fixed it.

## Limitations

~This change probably needs to be also applied to a few other templates, if it turns out this is indeed the desired fix. Also there should be probably tests covering this. I have not enough context and can't dig deeper into this right now, especially not knowing what's the latest take on multi-line translation string support, but I wanted to share what fixed it for me at least.~ 
Done!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,9 @@ _None_
 * Generate `xcfilelist`: Adds the template file path to the inputs `xcfilelist` (for custom templates).  
   [Craig Siemens](https://github.com/CraigSiemens)
   [#815](https://github.com/SwiftGen/SwiftGen/pull/815)
+* Strings: built-in templates now have better handling of multi-line translations.  
+  [@mrackwitz](https://github.com/mrackwitz)
+  [#774](https://github.com/SwiftGen/SwiftGen/pull/774)
 
 ### Internal Changes
 
--- a/Documentation/templates/strings/objc.md
+++ b/Documentation/templates/strings/objc.md
@@ -27,13 +27,13 @@ You can customize some elements of this template by overriding the following par
 
 ```swift
 @interface Localizable : NSObject
-// alert__message --> "Some alert body there"
+/// alert__message --> "Some alert body there"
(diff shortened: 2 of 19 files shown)


SwiftGen-Eve:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Error">
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please include a CHANGELOG entry to credit your work.  <br />
You can find it at <a href="https://github.com/mrackwitz/SwiftGen/blob/patch-1/CHANGELOG.md">CHANGELOG.md</a>.</td>
    </tr>
  </tbody>
</table>

---

📝 We use the following format for CHANGELOG entries:
```
* Fix multi-line translations in comments in structured-swift5.  
  [#774](https://github.com/SwiftGen/SwiftGen/pull/774)
  [@mrackwitz](https://github.com/mrackwitz)
```
:bulb: Don't forget to end the line describing your changes by a period and two spaces.


</p>

djbe:
Hi @mrackwitz,

Thank you for the PR. I'm wondering if this is necessary though? I did a quick test with a multiline string like this:
```strings
"multiple-lines" = "test
foo
bar";
```

And the output for both flat & structured was correct (valid):
```swift
internal enum L10n {
  /// test\nfoo\nbar
  internal static let multipleLines = L10n.tr("Localizable", "multiple-lines")
}
```

djbe (approved):
While I couldn't reproduce the issue and our tests (which do have multiline cases) never gave any problems, I do recognize that this change just generates nicer comments 😄

PR #779, opened 2020-09-26
Plist: fix handling of Bool properties

### LINKED ISSUE #775: Plist: generates incorrect Bool values
### Environment
- Xcode 12.0 GM
- SwiftGen 6.2.1 (via CocoaPods)

### Repro steps
- Create a plist `Example.plist`
- In the root dictionary add a Bool key `IsTestEnabled` with value `false` (which Xcode corrects to **0** and it cannot be bypassed)
- In swiftgen.yml:
```yaml
plist:
  - inputs: [PATH]/Example.plist
    outputs:
      - templateName: inline-swift5
        output: ExamplePlistKeys.swift
        params:
          enumName: "PlistKeys"
```

### Result
Compiler can't accept `0` as a Boolean value.
```swift
internal enum PlistKeys {
  internal enum Example {
    internal static let isTestEnabled: Bool = 0
  }
```

### Expected result
Boolean values are converted from `0` and `1` to `false` and `true`, respectively.

fortmarek:
Resolves https://github.com/SwiftGen/SwiftGen/issues/775

---

---

The problem is well-described in the added issue. The fix is simple - convert `Int` to `Bool` 

I am not sure how to catch this in the tests, so if there is away, I'd appreciate some quick pointer, so we can ensure there won't be any regression regarding this in the future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@ _None_
 * Update the Swift version in `.swift-version` so that the right version is used when building manually (using `swiftenv`).  
   [@cfiken](https://github.com/cfiken)
   [#764](https://github.com/SwiftGen/SwiftGen/issues/764)
+* Plist: fix Bool being assigned Integer when using `inline` template.  
+  [@fortmarek](https://github.com/fortmarek)
+  [#779](https://github.com/SwiftGen/SwiftGen/pull/779)
 
 ### Internal Changes
 
@@ -73,9 +76,6 @@ _None_
 * Strings: fix incorrect interpretation of format placeholders when there were missing positional parameters (e.g. `"%2$@"` without a `%1$…` defined).  
   [@AliSoftware](https://github.com/AliSoftware)
   [#634](https://github.com/SwiftGen/SwiftGen/pull/634)
-* Plist: fix Bool being assigned Integer when using `inline` template
-  [@fortmarek](https://github.com/fortmarek)
-  [#779](https://github.com/SwiftGen/SwiftGen/pull/779)
 
 ## 6.2.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,7 +75,7 @@ _None_
 ### Bug Fixes
 
 * Strings: fix incorrect interpretation of format placeholders when there were missing positional parameters (e.g. `"%2$@"` without a `%1$…` defined).  
-  [@AliSoftware](https://github.com/AliSoftware)
+  [Olivier Halligon](https://github.com/AliSoftware)
   [#634](https://github.com/SwiftGen/SwiftGen/pull/634)
 
 ## 6.2.1
@@ -133,25 +133,25 @@ Read the [SwiftGen 6.2 Migration Guide](Documentation/MigrationGuide.md#migratin
   [David Jennes](https://github.com/djbe)
   [#614](https://github.com/SwiftGen/SwiftGen/pull/614)
 * The use of `swiftgen <parser>` (e.g. `swiftgen strings`, `swiftgen xcassets`, …) command line for running individual parsers is now deprecated in favor of `swiftgen run <parser>`. See "New Features" below.  
-  [@AliSoftware](https://github.com/AliSoftware)
+  [Olivier Halligon](https://github.com/AliSoftware)
   [#705](https://github.com/SwiftGen/SwiftGen/pull/705)
 * The subcommand `swiftgen templates` has been renamed `swiftgen template` (singular); the plural form of the command has been deprecated and will be removed in next major version.  
-  [@AliSoftware](https://github.com/AliSoftware)
+  [Olivier Halligon](https://github.com/AliSoftware)
   [#697](https://github.com/SwiftGen/SwiftGen/pull/697)
 * The ability for SwiftGen to search custom named templates in `~/Library/Application Support` has been deprecated and will be removed in SwiftGen 7.0. This little known feature made SwiftGen dependent on the machine it was running on. Use `templatePath` to reference custom templates by path instead.  
-  [@AliSoftware](https://github.com/AliSoftware)
+  [Olivier Halligon](https://github.com/AliSoftware)
   [#717](https://github.com/SwiftGen/SwiftGen/pull/717)
 
 ### New Features
 
 * Invoking individual parsers from the command line is now done via `swiftgen run <parser>`.  We still highly recommend to use a configuration file for flexibility and performance reasons in your projects, and only use `swiftgen run <parser>` for things like quick iterations when writing your own custom templates.  
-  [@AliSoftware](https://github.com/AliSoftware)
+  [Olivier Halligon](https://github.com/AliSoftware)
   [#705](https://github.com/SwiftGen/SwiftGen/pull/705)
 * You can now easily create a new config file using `swiftgen config init`. This will create an example and commented config file and open it to let you edit it to your needs.  _Note that the generated config file is static content which doesn't take the user's project into account (though that might change in the future)_.  
-  [@AliSoftware](https://github.com/AliSoftware)
+  [Olivier Halligon](https://github.com/AliSoftware)
   [#694](https://github.com/SwiftGen/SwiftGen/pull/694)
 * You can now use `swiftgen template doc [parser] [templateName]` on the command line to quickly open the documentation for templates on GitHub directly from your terminal.  
-  [@AliSoftware](https://github.com/AliSoftware)
+  [Olivier Halligon](https://github.com/AliSoftware)
   [#697](https://github.com/SwiftGen/SwiftGen/pull/697)
 * Each parser now accepts an `options` dictionary, with which you can set internal parser settings to change its behaviour. See the parser's specific documentation for available options.  
   [David Jennes](https://github.com/djbe)
@@ -193,13 +193,13 @@ Read the [SwiftGen 6.2 Migration Guide](Documentation/MigrationGuide.md#migratin
 * Most templates now accept a parameter to force having the file name used as namespace (`enum <FileName>`) in generated code _even if_ there's only one single input file.  
   [Viktoras Laukevičius](https://github.com/viktorasl)
   [#669](https://github.com/SwiftGen/SwiftGen/issues/669)
-  [@AliSoftware](https://github.com/AliSoftware)
+  [Olivier Halligon](https://github.com/AliSoftware)
   [#693](https://github.com/SwiftGen/SwiftGen/pull/693)
 
 ### Bug Fixes
 
 * SwiftGen now properly shows a better help message and the command usage when running an incomplete command, instead of complaining about a config file.  
-  [@AliSoftware](https://github.com/AliSoftware)
+  [Olivier Halligon](https://github.com/AliSoftware)
   [#706](https://github.com/SwiftGen/SwiftGen/pull/706)
 * XCAssets: improved the performance for color assets by caching the resolved colors.  
   [David Jennes](https://github.com/djbe)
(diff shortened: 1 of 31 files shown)


SwiftGen-Eve:

---

</p>

CHANGELOG.md
```diff
@@ -73,6 +73,9 @@ _None_
 * Strings: fix incorrect interpretation of format placeholders when there were missing positional parameters (e.g. `"%2$@"` without a `%1$…` defined).  
   [@AliSoftware](https://github.com/AliSoftware)
   [#634](https://github.com/SwiftGen/SwiftGen/pull/634)
+* Plist: fix Bool being assigned Integer when using `inline` template
```
  <tbody>
    <tr>
      <td>:warning:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/SwiftGen/SwiftGen/blob/7f25db7e56d2f6963484bdf33115e11a3eb9f3a2/CHANGELOG.md#L76"></span>Line describing your entry should end with a period and 2 spaces.</td>
    </tr>
  </tbody>
</table>

templates/plist/inline-swift4.stencil
```diff
@@ -57,6 +57,8 @@ import Foundation
     {% empty %}
       :
     {% endfor %}]
+  {% elif metadata.type == "Bool" %}
+    Bool(truncating: {{ value }})
```
AliSoftware: That should work but wouldn't it be nicer to make the template solve the 0/1 test and generate `true` or `false` directly?

AliSoftware:
About the testing: you need to add a couple of Bool keys (one with value `true` the other with `false`) in one of the fixtures we use for the Plist Tests, and then regenerate the test outputs (Xcode Scheme "Generate Outputs"). That should cover the missing case by validating that the generated code indeed produces `true/false`

fortmarek:
So, funnily enough, when using `Generate Outputs` scheme, the values from `.plist` are not translated into ones and zeros, but into, what one would rather expect, true and false. I can't seem to find what the real difference between parsing using `swiftgen` scheme and parsing via `Generate Outputs` scheme is - but it is rather weird that the test scheme produces a different result from how it then behaves "in the real world".

AliSoftware:
That's strange; did you double-check that the plist used as fixture didn't have true/false but indeed 0/1 values in its XML?

If it does have 0/1 values in the XML but behaves really differently when you do a ⌘U vs when you enable the right arguments in the scheme and do a ⌘R then that might be a deeper problem (I think @djbe encountered a similar difference between test context and normal run context for something else but can't remember what it was)

fortmarek:
> did you double-check that the plist used as fixture didn't have true/false but indeed 0/1 values in its XML?

The true/false values are specified with `<true></true>` or `</true>`, 0 and 1 is not used anywhere.

The weird thing is that when I run `SwiftgenKit - Generate Contexts` scheme it parses the Bool values to `true` and `false`. That's why the tests for Plists have been passing.

But when I use the `swiftgen` Scheme and run it *on the very same* Plist, I get 0s and 1s. So, there's nothing wrong with the plist itself, but there is just a different behavior between test scheme and the `swiftgen` scheme.

AliSoftware (commented):
> The true/false values are specified with `<true></true>` or `</true>`, 0 and 1 is not used anywhere.
> …
> So, there's nothing wrong with the plist itself

Ah, so the original issue #775 is misleading then?

> The weird thing is that when I run SwiftgenKit - Generate Contexts scheme it parses the Bool values to `true` and `false`.
> But when I use the swiftgen Scheme and run it on the very same Plist, I get 0s and 1s.

Ok so that definitively seems like an issue between run and tests like we had in the past then 🤔  @djbe do you remember what the other one (when we encountered a similar issue with another feature) was about? Maybe it's related…

templates/plist/inline-swift4.stencil
fortmarek: Yep, agreed -> I think it'd be better if the 0s and 1s would not get into the context at all (as it happens in the tests currently). If that's not possible, we can check for `value` and return `true` or `false` directly, as you mentioned

fortmarek:
> Ah, so the original issue #775 is misleading then?

The part `which Xcode corrects to 0 and it cannot be bypassed` is, indeed, misleading, I think because Xcode does not change it to zero anywhere (it's about how we process the output when reading `.plist`)

AliSoftware:
Ok I think I know why the test and the actual run are different.

SwiftGen runs in multiple stages:

1. In the first step, we parse the input file(s) – the `.plist` – to transform it into a "Context" (this term comes from the templating engines vocabulary, basically it's some JSON-compatible structure)
2. In the second step, we combine the template the user passes with that "Context" to generate the output

The key difference between ⌘U and ⌘R is that during the *unit* tests, we test phases 1 and 2 separately (that's why they are *unit* tests), and to do that we store the intermediate results in YAML. Those YAML files we have in the repo (see `Tests/Fixtures/StencilContexts/`) serve a dual purpose during our unit tests: (1) as the expected output for phase 1 (we compare the actual output of phase 1 with those YAML to ensure they match) and (2) as input fixtures for phase 2, so that we can decouple the 2 phases – instead of having to always test them in tandem via an integration test.

That's why the problem doesn't appear during the tests:

 - When we run "SwiftGenKit - Generate Outputs" to generate those YAML intermediate files, the value `1` (well, more precisely as an `NSNumber(value: true)`) is parsed from the `.plist` but is serialised into the `YAML` as the value `true`. Then when we run the tests (⌘U) for phase 2, it uses that YAML as input to generate the output swift files, and a proper `true` value is generated. 
 - But when we run `swiftgen` (⌘R) it *doesn't* use the intermediate serialisation into YAML, instead it keeps the context in-memory, so it goes directly from `NSDictionary` to Stencil, and that's probably where the `NSNumber(value: true)` gets interpreted as a `1` instead of `true`

---

We might need to fix this directly in Stencil itself to ensure that a `NSCFBoolean` in the stencil input context is properly interpreted as a boolean. But since that's not the first time we see this kind of difference between ⌘R and ⌘U, maybe the more resilient solution would be to make the parsed context go into a YAML round-trip (not necessary serialising into a file, just use YAMS to convert to a YAML stream and back in-memory) during ⌘R so that it matches more closely what we have in ⌘U (and fixes the stencil boolean interpretation issue at the same time)

AliSoftware:
@fortmarek I've pushed a commit to do the round-trip to and from YAML when generating the `stencilContext` for Plist and this seems to fix it.

Not sure if we want a broader fix (apply the same logic to all the parsers), since the issue here seems to be with the way `Foundation` and `NSDictionary` deals with `NSNumber` and Cocoa types, not Swift native types 🤔  but most of other parsers don't parse their input with `NSDictionary(contentsOf: …)` so might not be a problem for those.

AliSoftware (approved):

djbe:
I'm curious what Yams does here that fixes the context... We may not want to depend on such a side-effect, which our tests don't even catch.

AliSoftware:
That's exactly why I've approved… but not merged this PR yet, wanted to get your take on it.
I agree that it feels a bit like a workaround to use this YAML round-trip. The other alternatives would be:

1. Fix Stencil to be able to interpret `NSCFBoolean` as `Bool` and not `Int`
2. Fix the thing that is parsed by `NSDictionary(contentsOf: file)` in a different way than using a round-trip to `YAMS`
3. Use a dedicated Property-List parser other than `NSDictionary`. I wonder if `PropertyListSerialization` and/or `Codable`+`PlistParser` have the same underlying issue that `NSDictionary(contentsOf:)` has. If they don't, using them instead could be a nicer solution, but we need to check that first.

---

iiuc, the key problem here is that we're using a parser that is in ObjC/Cocoa-land and thus uses `NS/CF` types (like `NSNumber`) rather than Swift native types like `Bool`.

But I fear that just doing (1) to fix Stencil for this will only fix one specific thing but not protect us from any similar misinterpretation of other NS/CF types and Cocoa implementation details like we saw in the past. Besides, we can't implement some generic fix that would walk the context tree and fix it in Stencil given that Stencil is supposed to be able to also work with dynamic objects in contexts (like what we use for the JSON/Plist `metadata`), so implementing this as a Stencil level would risk bypassing the laziness of those cases…

So my take is we try the `Codable` route (3) first, and if it doesn't work, we could implement some normalisation step (2) that would tree-walk the context and normalise every type (like `NSCFNumber`) we encounter into a more standard type (like `Bool` or `Int`), maybe copy/pasting some of the code from YAMS if that's something that they already have implemented there. But I feel that this normalisation pass (2) would very likely be error-prone, and would be hard to test by nature (because our tests serialise in YAML anyway)

djbe:
Allright, so the effect of this round-trip is that it converts NSNumber instances to actual Swift scalar types (`Bool`, `Int`, ...). 

During serialization, this bit is triggered (recursively for dictionaries/arrays), which we added during one of the previous issues:
https://github.com/SwiftGen/SwiftGen/blob/stable/Sources/SwiftGenKit/Utils/YamsSerialization.swift#L55-L67
And during de-serialization the Yams parser nicely parses it as actual Swift types.

djbe:
Not sure how the `Codable` solution would work, because Plists can have anything at their root (I think), could be dictionary, array, simple boolean, ... (same with JSON).

AliSoftware:
> Allright, so the effect of this round-trip is that it converts NSNumber instances to actual Swift scalar types (`Bool`, `Int`, ...).
> 
> During serialization, this bit is triggered (recursively for dictionaries/arrays), which we added during one of the previous issues:
> https://github.com/SwiftGen/SwiftGen/blob/stable/Sources/SwiftGenKit/Utils/YamsSerialization.swift#L55-L67
> And during de-serialization the Yams parser nicely parses it as actual Swift types.

Yeah, that corresponds to the idea I had for solution 2 – normalise the `NSDictionary` in order for it to only contain Swift native types instead of `NS/CF` types.

Ok so since we seem to already have that kind of patch anyway, maybe just applying that same `.represented()` normalisation  both for the test stages and during normal run would solve this issue more nicely

I'd still want to try the `Codable` approach at some point for Plist to see if – being Swift `Codable` and all – it generate native types for that or not

AliSoftware:
> Not sure how the `Codable` solution would work, because Plists can have anything at their root (I think), could be dictionary, array, simple boolean, ... (same with JSON).

Ah, good point. I think it's still doable (I have some gists around to do similar stuff of parsing arbitrary JSON) but might require some custom code similar to [Ian's AnyCodable](https://gist.github.com/IanKeen/df3bed29e2613ead94aa3b16967b9d14)

AliSoftware:
Note that even if this still feels like a workaround, the benefit of doing the round-trip to YAMS is that at least it guarantees us that the ⌘R will behave the same as ⌘U. I agree that ideally what we want is the opposite – implement what we want to do in ⌘R and make sure ⌘U behaves the same as real-life, rather than ⌘R be driven by the behavior of ⌘U – but at least this round trip guarantees us that even if there is some glitches during the tests that would mask some issues (like the Boolean interpretation here) because of this YAML intermediate step, we would at least be consistent in behavior everywhere and be more confident about our tests matching reality.

AliSoftware:
@djbe Just tried the solution with `AnyCodable` and it seems to work! (Just had to add a case for the `Date` type, which the gist didn't include). See the last commit I just pushed.

In fact, that seems to fix another bug we didn't spot before: the `User Ambiguous Integer` key in the fixture Plists are `<integer>1</integer>` but so far were generating `userAmbiguousInteger: Bool = true` instead of `: Int = 1` (and their associated metadata were `Bool` instead of `Int`)

---

If that's the solution we decide to use and you're satisfied with using this (rather than `.represented()`), I'm thinking we might actually want to apply the same for the JSON parser, using `JSONDecoder` + `AnyCodable` there too. (We might first want to check that it has good performance though, to avoid reverting the perf improvements that @marcelofabri worked on about JSON parsing)

Sources/SwiftGenKit/Parsers/AnyCodable.swift
```diff
@@ -0,0 +1,129 @@
+//
+// SwiftGen
+// Copyright © 2020 SwiftGen
+// MIT Licence
+//
+
+import Foundation
```
AliSoftware, line 7:
```suggestion
// Original credits to Ian Keen (https://gist.github.com/IanKeen/df3bed29e2613ead94aa3b16967b9d14)

import Foundation
```

Sources/SwiftGenKit/Parsers/AnyCodable.swift
```diff
@@ -0,0 +1,129 @@
+//
+// SwiftGen
+// Copyright © 2020 SwiftGen
+// MIT Licence
+//
+
```
AliSoftware, line 6:
```suggestion

// Original credits to Ian Keen (https://gist.github.com/IanKeen/df3bed29e2613ead94aa3b16967b9d14)

```

AliSoftware: @IanKeen Is it ok with you if we credit your gist this way while including it in our MIT-licensed code?

djbe (approved):
Looks good!

Once we've got the OK from Ian and this is merged, I'll open a PR to apply the same changes to the JSON parser. I'll try to run it on some big JSON files, to see if there's any major difference.

AliSoftware:
Note: we might want to update the CHANGELOG copy a bit for that fix now that it's not just an `elif` for booleans but more globally a change in how the files are parsed (and that it also fixed the `userAmbiguousInteger` case too)

fortmarek:
Thanks for fixing it @AliSoftware - sorry, I was not able to partake in the discussion yesterday, but the conclusion seems sound to me 👍 Let me know if you need any help with this PR

Sources/SwiftGenKit/Parsers/AnyCodable.swift
IanKeen: Of course! 🤘

AliSoftware: @IanKeen feel free to double-check my changes to your gist [here](https://github.com/SwiftGen/SwiftGen/pull/779/commits/bd3644d5bdd4c934de1404bbacbed9d44e6a7bba) – a second pair of eyes is always welcome, especially from the Codable expert in the room 😛 

(Another diff from our code and your gist is that I've added `Date` as a type supported by `AnyCodable` – since Plist are able to hold date instances after all)

PR #783, opened 2020-10-02
JSON/YAML: correctly handle `Bool` values in `inline` templates

### LINKED ISSUE #781: JSON: incorrectly handles some values
Similar to #775, the JSON parser incorrectly parses some values. This JSON file:
```json
{
  "count": 3,
  "feature": true,
  "other": false,
  "zero": 0
}
```

Becomes:
```swift
internal enum JSONFiles {
  internal static let count: Int = 5
  internal static let feature: Bool = 1
  internal static let other: Bool = 0
  internal static let zero: Bool = 0
}
```

So, what's wrong here is:
- The value of the 2 booleans is incorrectly generated
- The type of the `zero` property is `Bool` instead of `Int`. From more testing it incorrectly parses 0 and 1 integers, and arrays that could be confused as one type or the other. All because of NSNumber shenanigans.

PR #783 fixes the first issue. The second one seems to be an issue with `JSONSerialization`.

---------

Applying the `AnyCodable` fix from #779 generates this Swift code:
```swift
internal enum JSONFiles {
  internal static let count: Date = 5
  internal static let feature: Bool = true
  internal static let other: Bool = false
  internal static let zero: Date = 2001-01-01 00:00:00 +0000
}
```

Clearly the `Date` check is a bit too eager here. Temporarily removing `Date` parsing gives:
```swift
internal enum JSONFiles {
  internal static let count: Int = 5
  internal static let feature: Bool = true
  internal static let other: Bool = false
  internal static let zero: Int = 0
}
```

So couple of things to note:
- Need a variant of `AnyCodable` without `Date` parsing. Maybe subclassing `AnyCodable` with `AnyCodableIncludingDate`?
- The performance is really bad... For a 3.5MB file, `JSONSerialization` takes 0.028s, whereas `AnyCodable` takes 2.551s, about **91 times slower**.
- Maybe a JSON parser option to switch between `AnyCodable` and `JSONSerialization` parsers? The default should be `AnyCodable` (slower but more correct), and it's up to the end-user to write templates that handle bizarre/broken data?

djbe:
Partially addresses #781, references #779.


SwiftGen-Eve:

---

</p>

PR #786, opened 2020-10-05
JSON: fix 0/1 integer parsing

### LINKED ISSUE #781: JSON: incorrectly handles some values

djbe:
Fixes #781 

With JSONSerialization, 0 and 1 are parsed as `NSNumber`s. The issue is that these values are considered valid boolean values, so the `is Bool` or `is [Bool]` tests succeed.

This PR addresses the issue using runtime introspection.

## Implementation

There is no "simple" check to see what's actually inside an `NSNumber` instance. We first have to check if it's a boolean, and if not, get the internal type of the number. We then map these to the info we actually want.

The fix for `Array`s is a bit more involved, because `[NSNumber]` is tool free bridged to `[Int]`, `[Bool]`, etc... And the same applies to the inverse (a `[Bool]` is bridged to `[NSNumber]`). When we just check if something is `[NSNumber]` (and sometimes bridge it), we lose the easy check to see if all the elements in an array have the same type. We'd have to go through each element and check their type.

Without this extra array inspection (see commit b12aaa3da413d68d4b397a70f142afacd1d68806), things break. In the old method, `[0.1, 2, true]` is considered a `[Double]` array (see `mixed2` in the tests). This generates wrong Swift code, which causes compilation to fail:
```swift
static let mixed2: [Double] = [0.1, 2, true]
```

## Note

This issue **does not** appear in the YAML or Plist parsers. I've added a bunch of tests for all 3 parsers to verify some scenarios.

### DIFF
--- a/Sources/SwiftGenKit/Utils/Metadata.swift
+++ b/Sources/SwiftGenKit/Utils/Metadata.swift
@@ -36,33 +36,55 @@ enum Metadata {
   /// - Parameter data: The value to describe
   /// - Returns: Dictionary with type information about the value (for Stencil context)
   static func generate(for data: Any) -> [String: Any] {
-    switch data {
-    case is String:
+    let dataType = type(of: data)
+
+    if data is String {
       return [Key.type: ValueType.string]
-    case is Bool:
+    } else if dataType == Bool.self {
       return [Key.type: ValueType.bool]
-    case is Int:
+    } else if dataType == Int.self {
       return [Key.type: ValueType.int]
-    case is Double:
+    } else if dataType == Double.self {
       return [Key.type: ValueType.double]
-    case is Date:
+    } else if dataType == Date.self {
       return [Key.type: ValueType.date]
-    case is Data:
+    } else if dataType == Data.self {
       return [Key.type: ValueType.data]
-    case let data as [Any]:
+    } else if let data = data as? NSNumber {
+      return generate(for: data)
+    } else if let data = data as? [Any] {
       return [
         Key.type: ValueType.array,
         Key.element: Metadata.describe(arrayElement: data)
       ]
-    case let data as [String: Any]:
+    } else if let data = data as? [String: Any] {
       return [
         Key.type: ValueType.dictionary,
         Key.properties: Metadata.describe(dictionary: data)
       ]
-    case is NSNull, nil:
+    } else if dataType == NSNull.self || Mirror(reflecting: data).displayStyle == .optional {
       return [Key.type: ValueType.optional]
-    default:
+    } else {
       return [Key.type: ValueType.any]
+    }
+  }
+
+  static func generate(for number: NSNumber) -> [String: Any] {
+    if CFGetTypeID(number) == CFBooleanGetTypeID() {
+      return [Key.type: ValueType.bool]
+    } else {
+      switch CFNumberGetType(number) {
+      case .sInt8Type, .sInt16Type, .sInt32Type, .sInt64Type, .charType, .intType, .longType, .longLongType:
+        return [Key.type: ValueType.int]
+      case .float32Type, .float64Type, .floatType, .doubleType:
+        return [Key.type: ValueType.double]
+      case .nsIntegerType:
+        return [Key.type: ValueType.int]
+      case .cgFloatType:
+        return [Key.type: ValueType.double]
+      default:
+        return [Key.type: ValueType.any]
+      }
     }
   }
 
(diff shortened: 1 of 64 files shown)


SwiftGen-Eve:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

---

</p>

AliSoftware:
What about using Codable (and our AnyCodable helper recently introduced to fix a similar issue in Plist) instead?

djbe:
@AliSoftware see #781 where I did some tests with AnyCodable, the performance isn't that good unfortunately.

Sources/SwiftGenKit/Utils/Metadata.swift
```diff
@@ -36,32 +36,35 @@ enum Metadata {
   /// - Parameter data: The value to describe
   /// - Returns: Dictionary with type information about the value (for Stencil context)
   static func generate(for data: Any) -> [String: Any] {
-    switch data {
-    case is String:
+    let dataType = type(of: data)
+
+    if data is String {
```
AliSoftware, line 44: I think it could be worth a code comment to explain (iiuc) that this code uses `data is X` when we want toll-free-bridging to apply but `dataType == X.self` when we want to avoid TFB like for Int/Bool/… which could be misinterpreted and mis-bridged when coming in as NSNumber.

AliSoftware (approved):
I feel like we're adding more and more special cases as we encounter differences between JSONSerialization/Codable/YAMS interpreting things differently each; that also doesn't inspire confidence regarding unit tests (which always go through the intermediate state of YAML representation while real run doesn't)…

I really wish we could find a satisfying unified solution for those behavior diffs between test vs run one day…


Apart from that, the fix looks ok to me and should do (until we hopefully solve the bigger problem one day?)

djbe:
There are actually quite a few things going here (and all the other issues):
- Wild world of unstructured input data, anything goes (unfortunately?)
- Using `JSONSerialization` for performance reasons, which uses ObjC types internally
- Our metadata runtime inspection
- Swift implicit toll-free bridging, which we can't turn off even if we wanted to (for some cases)

I can always check if there are other (faster) JSON parsers directly to Swift types, although I'm not 100% sure it will fix everything. I think the other parsers are working correctly, it's only `JSONSerialization` + ObjC types that are throwing a wrench in things.

Note: let's also not forget that `Stencil` sometimes does weird things too, especially when outputting/writing the content of a variable.

AliSoftware:
Yeah which is why even if I think the ideal solution would be to find a _fast_ parser that would parse directly into _Swift_ types, I'm not sure such parser combining the two (Swift types + top performance) would be easy to find. So for now keeping NSJSONserialization + having those fixes to handle those "wild world of unstructured data" indeed, should be enough 👍

Issue #487, opened 2018-09-06
Tests: Compile output with a specific deployment target


djbe:
The NSDataAsset asset type PR (#445) added code that depends on a type that's only available from certain iOS/tvOS versions. It did have an `@available` check, but it was in the wrong place (it needed to be on the extension of the new type instead of the method, fixed in #482).

The reason our tests didn't catch this is because, while we do provide an SDK (the latest public available, see [here](https://github.com/SwiftGen/SwiftGen/blob/master/Rakefile#L63)), we do not provide a deployment target.

The main question then is, what deployment target do we chose for our templates? (per platform)
And we should document these choices in our template docs.

Note: this should not affect the `swiftgen` podspec, as the user can always choose to use their own templates for older platform versions.

AliSoftware:
Other question is, which deployment target do we test our targets with? Which is somewhat related to the deployment target we chose for our templates, but what I means is that we'll likely provide a *range* of SDKs + deployment targets supported (like "template tested for iOS 9+" or "template tested on macOS 10.10 thru 10.12"), but do we then run our CI to test only the minimum target supported (iOS9, macOS 10.10), or also to test with the latest target (iOS 12b as of now, and will evolve later), to be informed when it breaks in a future version ?

That would mean a lot of builds (each target being built on multiple SDKs and for each SDK on at least 2 deployment targets, the min and the max) though…

Or maybe just using the *latest* BaseSDK of the platform and just providing a DEPLOYMENT_TARGET setting for the *min version* supported ensures us that it covers everything, and we'd then be good with just 3 builds per template (one per platform iOS, macOS, watchOS)?

djbe:
Implemented in #789.

PR #788, opened 2020-10-07
XCAssets: add support for `symbolset`

djbe:
Fixes #767.

PR #789, opened 2020-10-07
XCAssets: fix some `availability` annotations

djbe:
While working on #788 and checking the docs, I noticed that some availability annotations are incorrect.
- NSDataAsset support on watchOS. Maybe support was probably added after we implemented this feature
- ~For some reason we were using the wrong initializer for colors on watchOS~ The docs are WRONG!
- Image initializers were completely missing availability annotations

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -79,6 +79,12 @@ opt_in_rules:
 conditional_returns_on_newline:
   if_only: true
 
+deployment_target:
+  iOS_deployment_target: 1.0
+  macOS_deployment_target: 10.0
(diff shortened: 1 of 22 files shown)


SwiftGen-Eve:

---

</p>

djbe:
That last commit fixes #487, so we define the deployment target (min. OS version) for each platform.
- For iOS we could have gone with 7.0 instead of 8.0, but then the IB tests would have failed (because of `AVPlayerViewController`).
- For macOS we need are bound to a minimum that supports `NSStoryboard`
- For watchOS and tvOS, we're using the minimum OS that supports Swift

PR #793, opened 2020-10-11
Convert project to SPM

### LINKED ISSUE #723: Switch to compilation via SPM
Seeing the issues with the 6.2 release on Homebrew, we should think about switching to compilation via SPM. We already support it partially: 
- We test building on CI (not testing yet though)
- We support distribution via Mint (which is SPM based)

We can easily switch building methods from `xcodebuild` to `swift build`, it would even simplify our Rakefile a bit (no more `install_name_tool` shenanigans). 

One of the things that we **have** to consider is maintaining compatibility for existing users. With that, I mostly mean what path users use to invoke SwiftGen. This only really matters for CocoaPods/zip users. The other installation methods install the binary at an internal path that doesn't matter to the user.

# Existing structure

This is a summary of the existing file structure, and the value of `templatesRelativePath` used when referencing to templates.

## CocoaPods (and zip)

- bin/swiftgen (MUST REMAIN)
- lib/...
- templates/...

`templatesRelativePath` = Info.plist:TemplatePath = ../templates

## Homebrew

- bin/swiftgen
- lib/...
- share/swiftgen/fixtures/...
- share/swiftgen/templates/...

`templatesRelativePath` = Info.plist:TemplatePath = #{pkgshare}/templates = ../share/swiftgen/templates

## Mint

- swiftgen
- ...-Info.plist
- templates/...

`templatesRelativePath` = bundle.path(forResource: "templates") = templates
**Note**: TemplatePath isn't filled in the info.plist

# Proposed structure

Essentially, we'd prefer compiling everything the same way. We could try to port the Mint/SPM structure to CocoaPods, as long as we add a symlink to not break usage by existing users.

Note that I've removed the `plist` files as used in the Mint build. I think it's best if we move away from that mechanism before tackling this issue.

## CocoaPods (and zip)

- swiftgen
- bin/swiftgen (symlink to ../swiftgen)
- templates/...

## Mint

- swiftgen
- templates/...

## Homebrew

- bin/swiftgen
- share/swiftgen/fixtures/...
- share/swiftgen/templates/...

# SPM support

While we don't support it yet because we need resources support, SPM recently accepted some evolution proposals that should enable us to support it at some point. Specifically [SE-0271](https://github.com/apple/swift-evolution/blob/master/proposals/0271-package-manager-resources.md) (resources support).

From a quick check of the proposal, I can't see any actual description of paths (yet). If anyone is more familiar with the proposal, please add your thoughts to this issue. Main question is what the directory structure is, and how much it differs from the proposed structures above. Another question is how much this will affect Mint support.

djbe:
Fixes #723.

This PR does a few things:
- Converts the whole project to use SPM
- Getting rid of the `xcodeproj`/`xcworkspace` in the process
- Getting rid of Cocoapods in the process
- Uses SPM to build the release (statically), so no more messing around with copying frameworks
- Separates out most CLI code into a framework (for testing purposes)
- Use commit hooks (Komondor) to run SwiftLint
- Adapts rake and docs as needed

Couple of things to note:
- Using SPM we have no control over "build phases", hence the switch to commit hooks. Even if new developers don't have the hooks installed, our CI will catch it anyway
- No more noisy git logs because of project/pod changes
- Had to move the fixtures & templates to new locations inside the corresponding "packages" (required by SPM)
- No more plists to manage the version info, it's hardcoded in a generated swift file
- Xcode 12.0 (and 12.1) has a bug where `swift test` fails (in terminal) because it can't find the resources. Xcode 12.2 fixes this. More info [here](https://github.com/apple/swift-package-manager/pull/2817).

Obviously this will require some brew installation changes, haven't tested that yet. The other installation methods work AFAIK.


SwiftGen-Eve:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

---

</p>

djbe:
Would you look at that, danger & rake still work (test locally but you never know):
Screen Shot 2020-10-11 at 22 30 59

sebj:
Is there any update on this pull request @djbe?

This PR seems to fix SwiftGen on M1 Macs (resolving #811 and probably other issues too), which I think might also be related to the `--disable-sandbox` flag that's being passed to `swift build` in the `Rakefile` of this PR (which is also what SwiftLint is doing, which compiles/installs with no issue!)

PR #808, opened 2020-12-15
Adding a SwiftUI `colorView` convenient property.

stuffmc:
By simply calling `.colorView` on a color (we could also simply call `.view` or declare the color with another name (or prefixed `Color.` instead of `Asset.`?). Also, I probably shouldn't have committed that Scheme change, right?


djbe:
Replaced by #979

PR #813, opened 2020-12-21
Strings: Parse comments from strings file

### LINKED ISSUE #563: Strings: Use comments from .strings file as documentation
It doesn’t currently seem possible to use the comment from within the .strings file as the document comment for the `String` constant in the source code. The current template just uses the translation as the documentation comment instead.

Input (.strings):
```
/* The title of the button used to confirm an action in an alert. */
"Alert.Confirmation" = "OK"
```

Actual Output (.swift):
```
/// OK
internal static let confirmation = Strings.tr("Example", "Alert.confirmation")
```

Expected Output (.swift):
```
/// The title of the button used to confirm an action in an alert.
internal static let confirmation = Strings.tr("Example", "Alert.confirmation")
```

Relevant documentation:
https://github.com/SwiftGen/SwiftGen/blob/master/Documentation/SwiftGenKit%20Contexts/Strings.md#output

CraigSiemens:
This PR adds parsing the the comments from the strings file so they can be added in the generated code. #563

The comments are passed to the `context` as `commentLines` since string files allow multiline comments. It's passed as an array because using `|split:"\n"` in the stencil file didn't seem to work.

The resulting comment will look unchanged is there's no comment in the string file, otherwise it's added below the comment with the translation.
```
/// Translation
///
/// A multiline
/// comment
```


SwiftGen-Eve:

---

</p>

CraigSiemens:
@djbe Is there anything I can do to help move along this PR?

CraigSiemens:
Hey @AliSoftware and @djbe, any chance I could get a review on this PR?

djbe:
So going quickly through the changes:

1. No need to add a separate strings test file. Just add some comments to existing ones, same cases like you had (string with/without comment, multiline comment). It reduces the amount of test files (and output) we have to maintain
2. There's no need to split a comment into `commentLines` at the parser level. Template can handle this just fine. See [this PR](#774) for example. So we just need to add the `comment` property to the context.
3. I think that the generated code should just use the comment if available, and fall back to the translation (what have right now). That'd match how `genstrings` generates strings from your codebase for example, and also how export to XLIFF would handle it.
4. Finally, don't forget to update the context documentation & template (generated code) documentations, as well as the readme and playgrounds. If you want, I can do that last step.

djbe (changes requested):
So, besides the merge issue, going quickly through the changes:

1. No need to add a separate strings test file. Just add some comments to existing ones, same cases like you had (string with/without comment, multiline comment). It reduces the amount of test files (and output) we have to maintain
2. There's no need to split a comment into `commentLines` at the parser level. Template can handle this just fine. See [this PR](https://github.com/SwiftGen/SwiftGen/pull/774) for example. So we just need to add a `comment` property to the context.
3. I think that the generated code should just use the `comment` if available, and fall back to the `translation` (what have right now). That'd match how `genstrings` generates strings from your codebase for example, and also how export to XLIFF would handle it.
4. Finally, don't forget to update the context documentation & template (generated code) documentations, as well as the readme and playgrounds. If you want, I can do that last step.

Documentation/SwiftGenKit Contexts/strings.md
```diff
@@ -42,6 +42,7 @@ The output context has the following structure:
          (useful to do recursion when splitting keys against `.` for structured templates)
        - `key`  : `String` — the full translation key, as it appears in the strings file
        - `translation`: `String` — the translation for that key in the strings file
+       - `comment`  : `String` — the full comment, if it is present in the strings file
```
djbe:
```suggestion
       - `comment`: `String` — the full comment, if it is present in the strings file
```

CHANGELOG.md
```diff
@@ -25,6 +25,9 @@ _None_
   [David Jennes](https://github.com/djbe)
   [#562](https://github.com/SwiftGen/SwiftGen/issues/562)
   [#797](https://github.com/SwiftGen/SwiftGen/pull/797)
+* Strings: The strings file parser will try to parse comments for the strings and add them to the stencil context.
```
```suggestion
* Strings: added support for `.strings` files comments. The built-in templates will now use them for comments instead of the translation of a key.  
```

djbe (changes requested):
Allright, some small nitpicks, but uhm... I did some tests, and I don't think the Scanner solution is going to work (at least as is).

For example, it crashes if I feed it a string like:
`"what./*happens*/.here" = "hello world! /* still in string */";`

To prevent having to do loads of logic/whatever to handle these cases, let's just ignore them. What if we:
- read the file into memory
- REMOVE all the keys and values, and replace them with our own markers
- then scan the comments like you're doing, but it'll be simpler, as we'll no longer need to handle other stuff, just simplified "markers"

Now, we'll have to be careful when removing the keys and values, to not match the contents of comments themselves. For example given the key `foo`  and comment `/* describes "foo" in some cases */`, if we're not careful, that comment would become `/* describes MAGIC_MARKER in some cases */`. Our best bet is using regexes matching `"KEY"\s*=\*s"VALUE"\s*;`, appropriately escaping things as needed.

The magic markers themselves could be something uncommon enough like `___SWIFTGEN___STRING___ENTRY___###` with those `#` being the index of the entry in the `entries`. So the whole `"key" = "value";` entry is replaced by such a marker.

Obviously, we'll need to add the necessary test cases to handle these annoying values 🤷

Sources/SwiftGenKit/Parsers/Strings/FileTypeParser/StringsFileParser.swift
```diff
@@ -23,100 +23,96 @@ extension Strings {
...
+
+      // replace entries with markers
+      for (index, entry) in zip(0..., entries) {
+        let key = entry.key.escapedStringsRegex
+        let translation = entry.translation.escapedStringsRegex
+        contents = contents.replacingOccurrences(
+          of: #"(?m)"\#(key)"\s*=\s*"\#(translation)"\s*;"#,
+          with: "\(String.startMarker)\(index)\(String.endMarker)",
+          options: .regularExpression
+        )
```
AliSoftware: The whole logic of the "marker system" will also fail for any `/* "key" = "translation" */` occurrence. For example:

```
"key1" = "value1";

// Same as "key1" = "value1"; but in the context of user not logged in
"key1.anonymous" = "value2";
```

imho, it's better to implement a proper tokenizer – rather than trying to play with this marker idea and risk introducing more corner cases (and potential injection "attacks") as you change the file content like that. Besides, this regex replacement means we'd have 2 passes on the whole text, one to search/replace, then one to parse, rather than tokenise directly.

AliSoftware (commented):
I've toyed with Parser combinators this weekend, and came up with the following playground code:

<details>
<summary><tt>StreamParser.swift</tt> (the base code to define generic Parser type)</summary>

```swift
import Foundation

struct Parser<Atom> {
  let parse: (Substring) -> (atom: Atom, rest: Substring)?
}

// MARK: - Parser Operators

extension Parser {
  func parse(_ string: String) -> (atom: Atom, rest: Substring)? {
    return self.parse(string[...])
  }

  static func oneOf<A>(_ list: [Parser<A>]) -> Parser<A> {
    Parser<A> { stream in
      list.lazy.compactMap { parser in parser.parse(stream) }.first
    }
  }

  func repeated(_ range: ClosedRange<Int> = 0...Int.max) -> Parser<[Atom]> {
    Parser<[Atom]> { stream in
      var rest = stream
      var atoms: [Atom] = []
      while range ~= atoms.count+1, let res = self.parse(rest) {
        atoms.append(res.atom)
        rest = res.rest
      }
      return atoms.isEmpty ? nil : (atom: atoms, rest: rest)
    }
  }

//  var maybe: Parser<Atom?> {
//    Parser<Atom?> { stream in
//      let res = self.parse(stream)
//      return (atom: res?.atom, rest: res?.rest ?? stream)
//    }
//  }
//
//  func then<NextAtom>(_ next: Parser<NextAtom>) -> Parser<(Atom,NextAtom)> {
//    Parser<(Atom,NextAtom)> { stream in
//      guard let resA = self.parse(stream) else { return nil }
//      guard let resB = next.parse(resA.rest) else { return nil }
//      return (atom: (resA.atom, resB.atom), rest: resB.rest)
//    }
//  }

  func map<NewAtom>(_ transform: @escaping (Atom) -> NewAtom) -> Parser<NewAtom> {
    Parser<NewAtom> { stream in
      guard let res = self.parse(stream) else { return nil }
      return (atom: transform(res.atom), rest: res.rest)
    }
  }
}

// MARK: - Some atomic parsers

extension Parser {
  static func character(_ char: Character) -> Parser<Character> {
    Parser<Character> { stream in
      guard let firstChar = stream.first, firstChar == char else { return nil }
      return (atom: firstChar, rest: stream.dropFirst())
    }
  }

  static var spacesOrTabs: Parser<[Character]> {
    let spaceOrTab: Parser<Character> = .oneOf([.character(" "), .character("\t")])
    return spaceOrTab.repeated()
  }

  static func literal(_ string: String) -> Parser<Substring> {
    Parser<Substring> { stream in
      guard stream.hasPrefix(string) else { return nil }
      let postIndex = stream.index(stream.startIndex, offsetBy: string.count)
      return (atom: stream[..<postIndex], rest: stream[postIndex...])
    }
  }

  static func wrappedString(openingBracket: String, closingBracket: String) -> Parser<Substring> {
    Parser<Substring> { stream in
      guard stream.hasPrefix(openingBracket) else { return nil }
      let startIndexAfterBracket = stream.index(stream.startIndex, offsetBy: openingBracket.count)
      var index = startIndexAfterBracket
      var inEscapeSequence = false
      while index < stream.endIndex {
        defer { index = stream.index(after: index) }
        if inEscapeSequence {
          inEscapeSequence = false
        } else if stream[index] == "\\" {
          inEscapeSequence = true
        } else if stream[index...].hasPrefix(closingBracket) {
          let afterClose = stream.index(index, offsetBy: closingBracket.count)
          return (atom: stream[startIndexAfterBracket..<index], rest: stream[afterClose...])
        }
      }
      return nil
    }
  }

  static var quotedString: Parser<Substring> {
    wrappedString(openingBracket: "\"", closingBracket: "\"")
  }
  static var inlineCommentString: Parser<Substring> {
    .wrappedString(openingBracket: "/*", closingBracket: "*/")
  }
}
```
</details>

##### <tt>StringsFileTokenParser.swift</tt>

```swift
// ////////////////////
// MARK: - Tokens-aware parsers

enum Token {
  case inlineComment(Substring) /* like these */
  case eolComment(Substring) // like there
  case keyValuePair(key: Substring, value: Substring, inlineComments: [Substring])
  case whitespace(Int)
  case newLine
}

extension Parser {
  static var inlineComment: Parser<Token> {
    return inlineCommentString.map { str in .inlineComment(str) }
  }

  static var eolComment: Parser<Token> {
    let p: Parser<Substring> = .wrappedString(openingBracket: "//", closingBracket: "\n")
    return p.map { str in .eolComment(str) }
  }

  static var keyValuePair: Parser<Token> {
    return Parser<Token> { stream in
      var comments: [Substring] = []

      func eatSpacesAndComments(_ stream: inout Substring) {
        stream = spacesOrTabs.parse(stream)?.rest ?? stream
        if let com = inlineCommentString.parse(stream) {
          comments.append(com.atom)
          stream = com.rest
        }
        stream = spacesOrTabs.parse(stream)?.rest ?? stream
      }

      guard var key = quotedString.parse(stream) else { return nil }
      eatSpacesAndComments(&key.rest)
      guard var equal = character("=").parse(key.rest) else { return nil }
      eatSpacesAndComments(&equal.rest)
      guard var value = quotedString.parse(equal.rest) else { return nil }
      eatSpacesAndComments(&value.rest)
      guard let semicolon = character(";").parse(value.rest) else { return nil }

      let atom: Token = .keyValuePair(key: key.atom, value: value.atom, inlineComments: comments)
      return (atom: atom, rest: semicolon.rest)
    }
  }

  static var whitespace: Parser<Token> {
    spacesOrTabs.map { c in .whitespace(c.count) }
  }

  static var newLine: Parser<Token> {
    character("\n").map { _ in .newLine }
  }

  static var localizableStringsFile: Parser<[Token]> {
    let singleToken: Parser<Token> = .oneOf([
      .inlineComment, .eolComment, .keyValuePair, .whitespace, .newLine
    ])
    return singleToken.repeated()
  }
}
```

<details>
<summary><tt>StreamParserDemo.swift</tt></summary>

```swift
// ////////////////////
// MARK: - Demo

extension Token: CustomDebugStringConvertible {
  var debugDescription: String {
    switch self {
    case .inlineComment(let string): return ".inlineComment(\(string))"
    case .eolComment(let string): return ".eolComment(\(string))"
    case .keyValuePair(let key, let value, let comments): return ".pair {\(key)} = {\(value)} - \(comments)"
    case .whitespace(let count): return ".ws(\(count))"
    case .newLine: return ".nl"
    }
  }
}

func demo() {
  let fileURL = Bundle.main.url(forResource: "Localizable", withExtension: "strings")
  let string = try! String(contentsOf: fileURL!)
  let parser = Parser<[Token]>.localizableStringsFile

  guard let res = parser.parse(string) else {
    return print("Invalid start of file") // throw some error in real code
  }
  guard res.rest.startIndex == string.endIndex else {
    let offset = string.distance(from: string.startIndex, to: res.rest.startIndex)
    return print("Invalid character at offset \(offset): \(res.rest)") // throw some error in real code
  }

  let tokens = res.atom
  print(tokens.map(\.debugDescription).joined(separator: "\n")) // nice line-by-line debug print
}
demo()
```
</details>

Once the file is tokenised into a list of comment, key-pairs and newlines, it's then easier to manipulate and convert into a list of entries with their associated comments.

@djbe I've tested the above parser with a couple of convoluted examples and so far it works (but of course it's easy to miss corner test cases when you're the one implementing the code… so feel free to try)

Won't have much more time today to translate that code into a commit on this PR, so if this seems like a good idea and right path to go, I'll let you use it as inspiration and adapt to the codebase.

Sources/SwiftGenKit/Parsers/Strings/FileTypeParser/StringsFileParser.swift
```diff
@@ -23,100 +23,96 @@ extension Strings {
...
+  /// Generates a regex for matching a key or value in a strings file
+  ///
+  /// Note that keys & values need to escape all `"\n\r\`. On top of that, newlines need to be prefixed with a `\`.
+  var escapedStringsRegex: String {
+    NSRegularExpression.escapedPattern(
+      for: self.replacingOccurrences(of: "\"", with: #"\""#)
+        .replacingOccurrences(of: "\r", with: #"\\#r"#)
+        .replacingOccurrences(of: "\n", with: #"\\#n"#)
+    )
+  }
```
CraigSiemens: String files support both newlines and escaped newlines. So this may not correctly match the contents of the string file if it's using non-escaped newlines.

Based on `parseQuotedPlistString` in
https://opensource.apple.com/source/CF/CF-550/CFPropertyList.c.auto.html

For example, the following keys will be parsed to have identical values.

```
let string = """
"key1" = "value
value";
"key2" = "value\nvalue";
"""

let decoder = PropertyListDecoder()
let dict = try decoder.decode([String: String].self, from: Data(string.utf8))

dict["key1"] == dict["key2"] // true
```

CraigSiemens (commented):
> For example, it crashes if I feed it a string like:

Could you share what the crash was? I cant reproduce it, all I see is it can fail to parse the comment on the following string entry with your example.

A simpler solution could be to have the scanner scan the comment, then the key, and then the value. That way it would avoid incorrectly reading the content of the value as a comment after the end of both the key and value.

The only checks would be needing to rename `scanKey()` to something like `scanQuotedString()`, then after successfully scanning the key, calling 

```
scanner.scanUpTo(.quote, into: nil)
_ = scanner.scanQuotedString()
```

CraigSiemens:
Hey @djbe and @AliSoftware,

I made some updates to this branch a little while and was hoping you could take a look again.

I changed the parsing to handle keys and values that look like they contain comments. I think it's better to avoid rebuilding all the string parsing and still rely on `PlistDecoder` to parse out the keys and values since there's probably way more edge cases we may not be aware of. Parsing out the comments feels like more of a "nice to have" thing, so I figured it shouldn't mess with pairings the strings and just try to add comments to this if it can find them. Worse case scenario, it would just treat a string like it didn't have a comment.

CraigSiemens:
Hey @djbe and @AliSoftware,

I've merged develop back into the PR to try and keep it up to date. I'd love if you could take a look at it again and let me know what you think.

The Danger check seems to be failing due to missing environment variable, I don't think it's anything to do with this branch.

djbe:
Hey @CraigSiemens,

I fully agree that the plist parser should be the "source of truth", it's what Apple uses, and handles all weird cases as it should. Parsing the comments is an "extra", just to enrich the existing information.

But the current parser in this PR is too simple, especially with all the many many **many** edge cases that are possible. I'd hoped avoiding that with the markers parser I proposed, but it was limited as well. There's a reason regexes shouldn't be used for parsing ~HTML~ this. We need to use a tokenizer like @AliSoftware mentioned, Apple probably does something like that as well.

So:
1. Use plist parser like we're already doing
2. Parse strings file again using tokenizer
3. Try to assign comments from tokenized items (step 2) to corresponding keys from step 1

CraigSiemens:
> But the current parser in this PR is too simple, especially with all the many many **many** edge cases that are possible. I'd hoped avoiding that with the markers parser I proposed, but it was limited as well. There's a reason regexes shouldn't be used for parsing ~HTML~ this.

In this current implementation I've moved away from using regex and it's scanner based now. I looked at the open source Plist decoder to try and ensure this works correctly for the string file format.

`parseOldStylePropertyListOrStringsFile` at https://opensource.apple.com/source/CF/CF-550/CFPropertyList.c.auto.html

Would you be able to share some of the edge cases where this wouldn't work properly? You may be more aware of what they are than I am. From what I can tell with the examples currently in the repo, this handles all the cases present. I'd be more than happy to improve this to handle more cases.

CraigSiemens:
Hey @djbe and @AliSoftware.

Any chance you could take a look at my last comment and help answer some questions for this? We'd still really like to have this in SwiftGen and since it's purely additive, it seems to be pretty low risk. We've been using my fork on our app with no issues.

djbe (approved):
I think this has reached a (soon to be 🤞) mergeable status.

I've rebased it on develop and merged some commits together for clarity's sake. Added one test case mentioned by AliSoftware that wasn't in there yet, and simplified the template implementation a bit.

Unless anyone got any more input/remarks, I'll merge this after the checks go green.

PR #836, opened 2021-05-15
Remove internal, default, access control level usage

cruisediary:

---

**Access Control Level**
https://docs.swift.org/swift-book/LanguageGuide/AccessControl.html

>All entities in your code (with a few specific exceptions, as described later in this chapter) have a default access level of internal if you don’t specify an explicit access level yourself. As a result, in many cases you don’t need to specify an explicit access level in your code.


djbe:
Hey @cruisediary,

Thank you for the PR! In this case, it might have been better to first open an issue to discuss this. Reason is, I'm not sure we should merge this:

One of the requirements (IMO) for generated code, should be (and this ranking is by no means complete):
1. Correctness (when dealing with Foundation/UIKit/...)
2. Performance
3. Well written code (a bit subjective 🤷)
4. Clarity (of code, purpose, etc...)
5. ...
6. Brevity?

This PR, which addresses brevity, does so at the expense of code clarity. While like you mentioned, the Swift default is `internal` if nothing is specified, it's something that in my opinion should be clear at all times without even thinking about it. If we'd leave it out, new Swift developers who are not that familiar with modules and access control, might expect declarations without access modifiers to be "public". By explicitly annotating declarations, they'll at least see that they are "internal", and start researching what impact that has on access from outside a module (for example).

All this to say that, while I greatly appreciate the effort for this PR, IMO we shouldn't merge this. If another core team member wants to chime in with a different opinion, we can always discuss this of course 😉

cruisediary:
Thank you for the your great comment! @djbe 

I agree that it would have been nice if I had first created an issue and asked for an opinion before PR 😄 and
It's really nice to consider new Swift developers who are not familiar with it as an open source provider.

To explain the background about this PR, I wanted to maintain consistency in access control level usage.
Already some codes in SwiftGen allow to use default acl Implicitly 🙄(by checking with SwiftLint)
>Done linting! Found 637 violations, 1 serious in 145 files.

I propose two ways to improve it.
- Use default access control level Implicitly like this pull request
- Use access control level Explicitly using [SwiftLint](https://github.com/realm/SwiftLint/pull/1966) already support it. We can use 'explicit_acl' rule for it

djbe:
You shouldn't be getting any lint violations, we disable all linting in generated files using `// swiftling:disable all` at the top of each file.

djbe:
Closing this due to inactivity. If there's still an issue, please open an *issue* (not a PR) to discuss this further.

As I mentioned before, we have certain defaults for clarity purposes. Issues you may have regarding swiftlint errors should not be happening (as we disable all linting). If not, as I said above, open an issue for it.

PR #846, opened 2021-06-17
Add  `--quiet` flag to suppress most log messages

### LINKED ISSUE #823: Support silent mode
It looks like there's currently no way to suppress the `File written: /path/to/file.swift` output. It would be nice to have this supported, and only print out log like this in the verbose mode.

Andre113:
Fixes #823

Adding a new `CommandLogLevel` enum to add the option for silent log level. The only exception will be an error message if the config file fails to open, like this one:

> 
> swiftgen: error: File swiftgen.yml not found.
> swiftgen: error: It seems like there was an error running SwiftGen.
> 
> - Verify that your configuration file exists at the correct path, or create a new one using:
> > swiftgen config init
> 
> - Verify that your configuration file is valid by running:
> > swiftgen config lint
> 
> - If you have any other questions or issues, we have extensive documentation and an issue tracker on GitHub:
> > https://github.com/SwiftGen/SwiftGen

OId scripts passing the `verbose` parameter will still work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ _None_
 
 ### New Features
 
-* Adding support for quiet log level.
+* Added support for `--quiet/-q` flag, to suppress all logs (except errors).  
   [Andre113](https://github.com/Andre113)
   [#562](https://github.com/SwiftGen/SwiftGen/issues/823)
   [#797](https://github.com/SwiftGen/SwiftGen/pull/846)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,7 @@ _None_
   [David Jennes](https://github.com/djbe)
   [#562](https://github.com/SwiftGen/SwiftGen/issues/562)
   [#797](https://github.com/SwiftGen/SwiftGen/pull/797)
+<<<<<<< HEAD
 * Support M1 and Intel devices (universal binary).  
   [David Jennes](https://github.com/djbe)
   [#805](https://github.com/SwiftGen/SwiftGen/issues/805)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,8 @@ _None_
 
 * Added support for `--quiet/-q` flag, to suppress all logs (except errors).  
   [Andre113](https://github.com/Andre113)
-  [#562](https://github.com/SwiftGen/SwiftGen/issues/823)
-  [#797](https://github.com/SwiftGen/SwiftGen/pull/846)
+  [#823](https://github.com/SwiftGen/SwiftGen/issues/823)
+  [#846](https://github.com/SwiftGen/SwiftGen/pull/846)
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,10 @@ _None_
 
 ### New Features
 
-_None_
+* Adding support for quiet log level.
+  [Andre113](https://github.com/Andre113)
+  [#562](https://github.com/SwiftGen/SwiftGen/issues/823)
+  [#797](https://github.com/SwiftGen/SwiftGen/pull/846)
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,7 +62,6 @@ _None_
   [David Jennes](https://github.com/djbe)
   [#562](https://github.com/SwiftGen/SwiftGen/issues/562)
   [#797](https://github.com/SwiftGen/SwiftGen/pull/797)
-<<<<<<< HEAD
 * Support M1 and Intel devices (universal binary).  
   [David Jennes](https://github.com/djbe)
   [#805](https://github.com/SwiftGen/SwiftGen/issues/805)
--- a/Documentation/ConfigFile.md
+++ b/Documentation/ConfigFile.md
@@ -132,13 +132,26 @@ If you need more control when using a configuration file, you can use some advan
   swiftgen config init --config tools/swiftgen/swiftgen-config.yml
   ```
   
-* Enable the verbose mode, which will print every command being executed when executing it, using the `--verbose` flag. This allows your to:
-  * control what is being run, by logging what happens during execution
-  * know the equivalent command to type if you were to run each swiftgen command manually instead of using the config file — which can be useful if you need to debug or tweak a particular command in isolation for example
+* Change the default log level. There are three options available. If no log level is specified, default will be used.
 
-  ```sh
-  swiftgen config run --verbose
-  ```
+  * Verbose mode, which will print every command being executed when executing it, using the `--verbose` flag. This allows your to:
+    * control what is being run, by logging what happens during execution
+    * know the equivalent command to type if you were to run each swiftgen command manually instead of using the config file — which can be useful if you need to debug or tweak a particular command in isolation for example
+
+      ```sh
+      swiftgen config run --logLevel verbose
+      ```
+  * Silent mode, which will silent every command being executed when executing it:
+
+      ```sh
+      swiftgen config run --logLevel silent
+      ```
+
+  * Default mode, which will print some of the commands being executed when executing it:
+
+    ```sh
+    swiftgen config run --logLevel default
+    ```
 
 ## Linting the configuration file
 
--- a/Documentation/ConfigFile.md
+++ b/Documentation/ConfigFile.md
@@ -141,7 +141,7 @@ If you need more control when using a configuration file, you can use some advan
       ```sh
       swiftgen config run --verbose
       ```
-  * Quiet mode (`--quiet` or `-q` flag), which will silent every command being executed when executing it (except for errors):
+  * Quiet mode (`--quiet` or `-q` flag), which will silence all logs (except for errors):
 
       ```sh
       swiftgen config run --quiet
--- a/Documentation/ConfigFile.md
+++ b/Documentation/ConfigFile.md
@@ -134,14 +134,14 @@ If you need more control when using a configuration file, you can use some advan
   
 * Change the default log level to quiet or verbose. If no log level is specified, default will be used.
 
-  * Verbose mode, which will print every command being executed when executing it, using the `--verbose` flag. This allows your to:
+  * Verbose mode (`--verbose` or `-v` flag), which will print every command being executed when executing it. This allows your to:
     * control what is being run, by logging what happens during execution
     * know the equivalent command to type if you were to run each swiftgen command manually instead of using the config file — which can be useful if you need to debug or tweak a particular command in isolation for example
 
       ```sh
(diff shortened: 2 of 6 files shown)


Sources/SwiftGenCLI/Logs.swift
```diff
@@ -8,6 +8,12 @@ import Foundation
 
 // MARK: Printing on stderr
 
+public enum CommandLogLevel: String {
+  case silent, `default`, verbose
```
djbe: I'd rename `silent` to `quiet`

Tests/SwiftGenTests/ConfigRunTests.swift
```diff
@@ -27,7 +27,7 @@ final class ConfigRunTests: XCTestCase {
     do {
       let config = try Config(file: configFile, logger: logger.log)
       try configFile.parent().chdir {
-        try config.runCommands(verbose: true, logger: logger.log)
+        try config.runCommands(logLevel: CommandLogLevel.verbose, logger: logger.log)
```
djbe: You can just write `.verbose` here.

Sources/SwiftGen/Commander/ConfigCommands.swift
```diff
@@ -39,20 +39,27 @@ enum ConfigCLI {
...
+    Option("logLevel", default: CommandLogLevel.default.rawValue, description: "Log level"),
     Flag("verbose", default: false, flag: "v", description: "Print each command being executed")
-  ) { file, verbose in
+  ) { file, logLevel, legacyVerbose in
     do {
       try ErrorPrettifier.execute {
         let config = try Config(file: file)
-
-        if verbose {
+
```
djbe: So something like 
```swift
switch (quiet, verbose) {
case (true, _): commandLogLevel = .quiet
case (_, true): commandLogLevel = .verbose
default: commandLogLevel = .default
}
```

Documentation/ConfigFile.md
```diff
@@ -132,13 +132,26 @@ If you need more control when using a configuration file, you can use some advan
   swiftgen config init --config tools/swiftgen/swiftgen-config.yml
   ```
   
-* Enable the verbose mode, which will print every command being executed when executing it, using the `--verbose` flag. This allows your to:
```
djbe, line 135: Obviously tweak docs regarding the `--quiet` flag

djbe (changes requested):
Good idea, although I'd keep the verbose flag, and instead just add a `-q/--quiet` flag, like all other CLI tools use.

Keep the loglevel enum though, that's a good change, but in the command create it based on the combination of verbose&quiet flags.

Andre113:
> Good idea, although I'd keep the verbose flag, and instead just add a `-q/--quiet` flag, like all other CLI tools use.
> 
> Keep the loglevel enum though, that's a good change, but in the command create it based on the combination of verbose&quiet flags.

I decided to use an enum because it would be easier to add new log levels in the future (eg: `debug`, `critical`, etc). If we choose to use flags, we would have to create a new one for each of them.

```
static let run = command(
  CLIOption.configFile(),
  Flag("quiet", default: false, flag: "q", description: "Print nothing"),
  Flag("verbose", default: false, flag: "v", description: "Print each command being executed"),
  Flag("debug", default: false, flag: "d", description: "Print some commands"),
  Flag("critical", default: false, flag: "c", description: "Print critical errors only")
)
```

Also, I would like to avoid giving the user the option to use both flags at the same time, like this:
`swiftgen config run -q -v`

I like the idea of changing from `silent` to `quiet thought`.

djbe:
If at some point we need more levels that **users** need to be able to control, then we can consider an enum input/argument, and deprecate the old flags if we want to. For now, our logging needs aren't that extensive, and I don't see that changing anytime soon.

> Also, I would like to avoid giving the user the option to use both flags at the same time, like this:

I'm honestly not sure what that would do... `-v` increases logging, but `-q` cancels all (non-error) logging. Quiet overrides verbose, as it usually does.

Andre113:
> If at some point we need more levels that **users** need to be able to control, then we can consider an enum input/argument, and deprecate the old flags if we want to. For now, our logging needs aren't that extensive, and I don't see that changing anytime soon.
> 
> > Also, I would like to avoid giving the user the option to use both flags at the same time, like this:
> 
> I'm honestly not sure what that would do... `-v` increases logging, but `-q` cancels all (non-error) logging. Quiet overrides verbose, as it usually does.

Ok, I made the changes. We have `--quiet` now and it will override `--verbose` if both are set by mistake.

Sources/SwiftGenCLI/Logs.swift
Andre113: Done.

Tests/SwiftGenTests/ConfigRunTests.swift
Andre113: Sure. Done.

Sources/SwiftGen/Commander/ConfigCommands.swift
Andre113: Done.

Documentation/ConfigFile.md
Andre113: Made the changes

Sources/SwiftGen/Commander/ConfigCommands.swift
```diff
@@ -39,20 +39,28 @@ enum ConfigCLI {
...
-
-        if verbose {
-          logMessage(.info, "Executing configuration file \(file)")
+        
+        switch (quiet, verbose) {
+        case (true, _):
+            commandLogLevel = .quiet
+        case (_, true):
+            commandLogLevel = .verbose
+            logMessage(.info, "Executing configuration file \(file)")
```
djbe: Move this out (after) of the switch, it won't be displayed when the level is set accordingly.

Documentation/ConfigFile.md
```diff
@@ -132,13 +132,20 @@ If you need more control when using a configuration file, you can use some advan
...
-  * control what is being run, by logging what happens during execution
-  * know the equivalent command to type if you were to run each swiftgen command manually instead of using the config file — which can be useful if you need to debug or tweak a particular command in isolation for example
+* Change the default log level to quiet or verbose. If no log level is specified, default will be used.
 
-  ```sh
-  swiftgen config run --verbose
-  ```
+  * Verbose mode, which will print every command being executed when executing it, using the `--verbose` flag. This allows your to:
```
djbe:
```suggestion
  * Verbose mode (`--verbose` or `-v` flag), which will print every command being executed when executing it. This allows your to:
```

Documentation/ConfigFile.md
```diff
@@ -132,13 +132,20 @@ If you need more control when using a configuration file, you can use some advan
...
+    * control what is being run, by logging what happens during execution
+    * know the equivalent command to type if you were to run each swiftgen command manually instead of using the config file — which can be useful if you need to debug or tweak a particular command in isolation for example
+
+      ```sh
+      swiftgen config run --verbose
+      ```
+  * Quiet mode, which will silent every command being executed when executing it (except for errors):
```
djbe:
```suggestion
  * Quiet mode (`--quiet` or `-q` flag), which will silent every command being executed when executing it (except for errors):
```

Sources/SwiftGenCLI/Logs.swift
```diff
@@ -41,12 +47,15 @@ enum ANSIColor: UInt8, CustomStringConvertible {
 
 // Based on https://github.com/realm/SwiftLint/blob/0.39.2/Source/SwiftLintFramework/Extensions/QueuedPrint.swift
 
+
```
djbe: Why this extra whitespace?

djbe (changes requested):
Small nitpicks (and the linter warnings), after that GTM I think.

Note: you can run the linter locally using `bundle exec rake lint:code` and `bundle exec rake lint:tests`. Or if you want to have these run as commit hooks, run `swift run komondor install` once inside the repo.

Sources/SwiftGenCLI/Logs.swift
Andre113: Removed

Documentation/ConfigFile.md
Andre113: Done

Documentation/ConfigFile.md
Andre113: Done

Sources/SwiftGen/Commander/ConfigCommands.swift
Andre113: Oops. Done.

CHANGELOG.md
```diff
@@ -29,6 +29,10 @@ _None_
   [David Jennes](https://github.com/djbe)
   [#805](https://github.com/SwiftGen/SwiftGen/issues/805)
   [#880](https://github.com/SwiftGen/SwiftGen/pull/880)
+* Adding support for quiet log level.
```
djbe: With the release of 6.5 and changelog reset, pretty sure that once you update your branch this will be in the wrong place.

djbe (approved):
Should be good to go once you update your branch

CHANGELOG.md
djbe: Bad news, the changelog is going to reset after release of 6.5.1 again 😅 Had to quickly release a fix for Mint installation flows

Andre113: Oops
I just updated it. Will update again after 6.5.1 so.

Andre113:
Done, just updated the changelog after 2.5.1 reset. Made something wrong when rebasing, so I had to do my lint fixes again.

CHANGELOG.md
```diff
@@ -10,7 +10,10 @@ _None_
 
 ### New Features
 
-_None_
+* Adding support for quiet log level.
```
```suggestion
* Added support for `--quiet/-q` flag, to suppress all logs (except errors).  
```

Documentation/ConfigFile.md
```diff
@@ -132,13 +132,20 @@ If you need more control when using a configuration file, you can use some advan
...
-  swiftgen config run --verbose
-  ```
+  * Verbose mode (`--verbose` or `-v` flag), which will print every command being executed when executing it. This allows your to:
+    * control what is being run, by logging what happens during execution
+    * know the equivalent command to type if you were to run each swiftgen command manually instead of using the config file — which can be useful if you need to debug or tweak a particular command in isolation for example
+
+      ```sh
+      swiftgen config run --verbose
+      ```
+  * Quiet mode (`--quiet` or `-q` flag), which will silent every command being executed when executing it (except for errors):
```
djbe:
```suggestion
  * Quiet mode (`--quiet` or `-q` flag), which will silence all logs (except for errors):
```

djbe (changes requested):

Andre113:
Updated the documentation with your suggestions.

djbe (approved):

PR #885, opened 2021-10-04
Workaround for Mint that does not yet support the SPM resource bundle

### LINKED ISSUE #883: Mint: crash with 6.5 and bundled templates
I just installed SwiftGen 6.5.0 via Mint and now I get the following error when running Swiftgen:

```
mint run SwiftGen/SwiftGen@6.5.0 config run --config scripts/swiftgen.yml
SwiftGenCLI/resource_bundle_accessor.swift:11: Fatal error: could not load resource bundle: from /Users/braste/.mint/packages/github.com_SwiftGen_SwiftGen/build/6.5.0/SwiftGen_SwiftGenCLI.bundle or /private/var/folders/rl/9ykwdfps5sq6gb386g34dfmd2d7tbj/T/mint/github.com_SwiftGen_SwiftGen/.build/arm64-apple-macosx/release/SwiftGen_SwiftGenCLI.bundle
[1]    92911 trace trap  mint run SwiftGen/SwiftGen@6.5.0 config run --config scripts/swiftgen.yml
```

EDIT: It works fine when I replace SwiftGen with mthole/SwiftGen@xc13-fix

tid-kijyun:
Fixes #883.
It seems that Mint(v0.17.0) does not yet support the resource bundles supported by SPM (since 5.3).
To work around this problem, write the path to the bundle in `Package.resources` and tell Mint to copy it.

`Package.resources` should be removed after Mint adds support for SPM resource bundles.
However, it should work the same way as overwriting the same bundle, so I don't expect any additional problems for the time being.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,10 @@ _None_
 
 ### Bug Fixes
 
-_None_
+* Workaround for Mint that does not yet support the SPM resource bundle.
+  [@tid-kijyun](https://github.com/tid-kijyun)
+  [#883](https://github.com/SwiftGen/SwiftGen/issues/883)
+  [#885](https://github.com/SwiftGen/SwiftGen/pull/885)
 
 ### Internal Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ _None_
 
 ### Bug Fixes
 
-* Workaround for Mint that does not yet support the SPM resource bundle.
+* Workaround for Mint that does not yet support the SPM resource bundle.  
   [@tid-kijyun](https://github.com/tid-kijyun)
   [#883](https://github.com/SwiftGen/SwiftGen/issues/883)
   [#885](https://github.com/SwiftGen/SwiftGen/pull/885)
--- a/Package.resources
+++ b/Package.resources
@@ -1,0 +1,2 @@
+.build/release/SwiftGen_SwiftGenCLI.bundle
+


CHANGELOG.md
```diff
@@ -44,6 +44,10 @@ _None_
 * Strings: built-in templates now have better handling of multi-line translations.  
   [@mrackwitz](https://github.com/mrackwitz)
   [#774](https://github.com/SwiftGen/SwiftGen/pull/774)
+* Workaround for Mint that does not yet support the SPM resource bundle.
```
djbe: This is in the wrong place.

djbe (changes requested):
`mint run tid-kijyun/SwiftGen@develop config run --config ...` seems to work (compared to 6.5.0)

Could you update your branch and put the changelog entry in the right place? Then I can merge this and release a bug fix (or change the 6.5 tag, not sure yet)

CHANGELOG.md
tid-kijyun: Is the correct place for this an "Internal Changes"?

djbe: No it's in the right "section", but your branch is behind on ours. 6.5 has been tagged and the changelog reset, put your changelog entry there.

tid-kijyun: Sorry about that. I fixed it.

CHANGELOG.md
```diff
@@ -14,7 +14,10 @@ _None_
 
 ### Bug Fixes
 
-_None_
+* Workaround for Mint that does not yet support the SPM resource bundle.
```
djbe: Two spaces after the `.`. It's unfortunate that the danger bot can't run on PR's from other repositories, ~or we'd have noticed this sooner.~ the previous commit was correct.

djbe (changes requested):

djbe (approved):

PR #890, opened 2021-10-18
Strings: reverted to case-sensitive sorting for keys

### LINKED ISSUE #891: Strings: unstable output for keys that differ just by case
If .strings contains two keys that differ just by case, the order of entries in the generated .swift file is not stable from invocation to invocation.

This is quite edge case, given that (as far as I understand) SwiftGen does not support multiple entries that differ just by case, but it still happens in our case, as we have to entries that differ just by case of keys *and* number of parameters, like below.

Localizable.strings:
```
"A" = "Foo";
"a" = "Bar %s";
```

The other rationale for this change is that the current case-insensitive sorting does not make that much sense in case camelCase is used for the keys: capital letters in camelCase typically introduce a grouping/another "word", but lowercase letters just introduce word variation. From that perspective I believe that word variations should better belong to one group.

Demo of the issue: https://github.com/grigorye/SwiftGen-Case-Insensitive

grigorye:
This should fix https://github.com/SwiftGen/SwiftGen/issues/891, as well as slightly affect the performance due to no need to lowercase keys on every comparison during the sorting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,11 @@ _None_
   [David Jennes](https://github.com/djbe)
   [#950](https://github.com/SwiftGen/SwiftGen/issues/950)
   [#959](https://github.com/SwiftGen/SwiftGen/pull/959)
+* All parsers: fixed unstable output for keys that differed just by case.  
+  [grigorye](https://github.com/grigorye)
+  [David Jennes](https://github.com/djbe)
+  [#891](https://github.com/SwiftGen/SwiftGen/issues/891)
+  [#890](https://github.com/SwiftGen/SwiftGen/pull/890)
 
 ### Internal Changes
 
--- a/Sources/SwiftGenKit/Stencil/AssetsCatalogParser+Context.swift
+++ b/Sources/SwiftGenKit/Stencil/AssetsCatalogParser+Context.swift
@@ -14,7 +14,7 @@ import Foundation
 extension AssetsCatalog.Parser {
   public func stencilContext() -> [String: Any] {
     let catalogs = self.catalogs
-      .sorted { lhs, rhs in lhs.name < rhs.name }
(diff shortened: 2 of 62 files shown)


djbe:
Hey @grigorye I've updated the PR to use `$0.key.caseInsensitiveCompare($1.key) == .orderedAscending` again as discussed in #891, and also broadly applying it to the other parsers.

Mind giving it a whirl?

jasdev:
~~Any chance this can land? 🙏🏽 it’s unfortunately a papercut I hit on every build after an asset changed case in our repo. —~~ see below.

Screen Shot 2022-08-28 at 10 57 11 AM

djbe:
@jasdev Have you tried the changes in this branch?

Do note that `git` also has issues with case renames, so you may still have issues. The example you provided is kinda normal? There'll always be a change to the generated code because `grain` != `Grain`. Obviously the sorting could be better (this PR), but the changes will still be there.

jasdev:
@djbe: Thanks for the reply and ahhh I see, sorry for the noise — this seems like a git issue after another look at it. What had happened was, that asset’s case changed, we committed that re-gen. a while back, and for some reason every subsequent build on my machine caused SwiftGen to undo the case change. I ended up just wiping the repo. and doing a re-clone and now I don’t get that generated case undo.

djbe:
Yeah this is a very well known issue with `git`, encountered it myself quite a few times. The thing with those renames is that they have to happen with `git mv` (see [here](https://sebastiandedeyne.com/how-to-do-a-case-sensitive-file-rename-in-git-on-macos/)). And even then, I think other people working on the same repo with existing checkouts, it'll still be the old case 🤷 

Anyway, this has little to do with SwiftGen 😅 @grigorye's original issue was with 2 keys with the same value (differing only in case). Honestly not a realistic situation, curious if anyone has a good test case for this. And if so, please try out these changes.

PR #940, opened 2022-05-18
Make generated CoreData model class final when model isn't abstract

grdsdev:
Change Core Data Swift 5 template for adding support to set a class as `final` when the model isn't abstract.

### DIFF
--- a/Sources/SwiftGenCLI/templates/coredata/swift5.stencil
+++ b/Sources/SwiftGenCLI/templates/coredata/swift5.stencil
@@ -20,6 +20,7 @@ import {{ import }}
 {% for name, entity in model.entities %}
 {% set superclass %}{{ model.entities[entity.superEntity].className|default:"NSManagedObject" }}{% endset %}
 {% set entityClassName %}{{ entity.className|default:"NSManagedObject" }}{% endset %}
+{% set hasChildren %}{% for aName,anEntity in model.entities where anEntity.superEntity == name %}1{% endfor %}{% endset %}
 // MARK: - {{ entity.name }}
 
 {% if not entity.shouldGenerateCode %}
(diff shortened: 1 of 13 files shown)


grdsdev:
What you guys think about this? @djbe and @AliSoftware

Sources/SwiftGenCLI/templates/coredata/swift5.stencil
```diff
@@ -20,6 +20,7 @@ import {{ import }}
 {% for name, entity in model.entities %}
 {% set superclass %}{{ model.entities[entity.superEntity].className|default:"NSManagedObject" }}{% endset %}
 {% set entityClassName %}{{ entity.className|default:"NSManagedObject" }}{% endset %}
+{% set final %}{% if entity.isAbstract %}{% else %}final {% endif %}{% endset %}
```
djbe: Avoid setting a property if you're only going to use it once, especially for for simple checks.

Sources/SwiftGenCLI/templates/coredata/swift5.stencil
```diff
@@ -32,7 +33,7 @@ import {{ import }}
 {% if param.generateObjcName %}
 @objc({{ entityClassName }})
 {% endif %}
-{{ accessModifier }} class {{ entityClassName }}: {{ superclass }} {
+{{ accessModifier }} {{ final }}class {{ entityClassName }}: {{ superclass }} {
```
djbe:
```suggestion
{{ accessModifier }} {% if not entity.isAbstract %}final {% endif %}class {{ entityClassName }}: {{ superclass }} {
```

djbe (changes requested):
Good idea. I've added some comments on your changes.

Also note that this change needs to also be made for the other templates (`swift4` in this case), and generate the output as well for those.

grdsdev:
Hey, @djbe I noticed that you already made the changes. Is there anything that still needs to be handled?

djbe:
I was waiting for the tests to finish, and they curiously enough revealed an error. In our tests we have:
- A `MainEntity`, which is *not* abstract
- A `ChildEntity`, which has `MainEntity` as a parent

You can create this situation in the model editor. I have no idea how useful it is, or valid, but it's possible. Hence, we have to handle it gracefully.

djbe:
There, that should solve it. Had to resolve to a trick in Stencil to check if something exists in an array 😅

djbe (approved):

PR #941, opened 2022-05-21
Add stringLocale parameter fixes a bug where plurals were not displayed correctly when using bundle or lookupFunction

HonQii:
…ed correctly when using bundle or lookupFunction

---


djbe:
This is in general a bad practice. Sure it may work for you as long as you provide a valid locale for your codebase, but it will NOT work for things like:
- Interface builder
- System permissino dialogs (that use infoplist.strings)
- Whole bunch of other stuff

Also note that this approach is no longer needed, since users can override the language per-app in iOS 13+. Apple even has a documentation page for this: https://developer.apple.com/news/?id=u2cfuj88

For those reasons I can't see us merging this. You can keep using a custom template (which you probably already are).

PR #964, opened 2022-07-22
Strings: implement fallback translation in generated code

### LINKED ISSUE #937: Strings: Fallback to development language when key is missing
Creating a new issue since #381 was closed. - I believe I have addressed the concerns of previous reviewers and provided more clarity on the source of the issue.

swiftgen --version = **6.5.1**

Happy to create the PR, once there is some consensus. 

## The Problem

Xcode does not handle fallback for partially translated .strings files, only when the file is missing. To state the issue more clearly, ill take a paragraph from this article that explains it well.

[Article link](https://samwize.com/2018/01/23/localization-fall-back-to-base-language/)

CleanShot 2022-05-09 at 14 21 57@2x

> **Fallback 3: Unsupported Phrase**
Let’s take the example where a user prefers zh-Hans.
>
>What happens when iOS try to get the NSLocalizedString of `“another-title”`?
Remember, zh-Hans is not fully translated. It does not have `“another-title”`. What do you think will happen?
>
>1) Fallback to development language en, or
>2) return `“another-title”`
>
>Many developers think it is (1). Unfortunately, it is (2), with that ugly key name!

## Solution

Here is an updated version of the code by @jonandersen in 381, that handles any development language as the base/fallback.

```
func localizeStringOrFallback(_ table: String, _ key: String, _ args: CVarArg...) -> String {
    let bundle = BundleToken.bundle
    let format = NSLocalizedString(key, tableName: table, bundle: bundle, comment: "")
    let string = String(format: format, locale: Locale.current, arguments: args)

    // If string key does not find a match within the Locale.current file. Look up the fallback string.
    // The edge case of this is that if a key and value are expected to match: "Foo" = "Foo", it will trigger the fallback logic.
    if string == key,
       let developmentRegion = bundle.infoDictionary?["CFBundleDevelopmentRegion"] as? String,
       let path = bundle.path(forResource: developmentRegion, ofType: "lproj"),
       let bundle = Bundle(path: path)
    {
        let format = NSLocalizedString(key, bundle: bundle, comment: "")
        return String(format: format, locale: Locale(identifier: developmentRegion), arguments: args)
    } else {
        return string
    }
}
```

I have tested this solution, and created a custom stencil implementing this logic. However, this complicates maintenance of the repo when SwiftGen releases updated stencils for new Swift versions.

I also think it would be a useful solution to others. Where showing fallback text is likely a better "bug" than displaying a key.

## Known issue with this approach
One known issue with this is that if a key and value are expected to match: "Foo" = "Foo", it will trigger the fallback logic.

Therefore if you had:
```
// en.proj
"Foo" = "A string that does not match its key";

// fr.proj
"Foo" = "Foo";      <------ if there was a use-case to repeat the key.
```
In this case, the displayed string would be "A string that does not match its key".

## Stencil Code

```
extension {{enumName}} {
    return localizeStringOrFallback(table, key, args)
  }
}

extension {{enumName}} {
  func localizeStringOrFallback(_ table: String, _ key: String, _ args: CVarArg...) -> String {
    let bundle = BundleToken.bundle
    let format = NSLocalizedString(key, tableName: table, bundle: bundle, comment: "")
    let string = String(format: format, locale: Locale.current, arguments: args)

    // If string key does not find a match within the Locale.current file. Look up the fallback string.
    // The edge case of this is that if a key and value are expected to match: "Foo" = "Foo", it will trigger the fallback logic.
    if string == key,
       let developmentRegion = bundle.infoDictionary?["CFBundleDevelopmentRegion"] as? String,
       let path = bundle.path(forResource: developmentRegion, ofType: "lproj"),
       let bundle = Bundle(path: path)
    {
        let format = NSLocalizedString(key, bundle: bundle, comment: "")
        return String(format: format, locale: Locale(identifier: developmentRegion), arguments: args)
    } else {
        return string
    }
  }
}
```

## Alternative solution

Don't change anything, and expect dev's to insert an entry for all locales, even if not translated. i.e. Insert the fallback string manually.

This is error prone, ive seen it in apps ive worked on. where keys might be added to `en` but not `en-US`. and then `en-US` users are seeing keys displayed in the UI as text.

djbe:
As rightly documented by @chrisjrex in #937:
- We use `localizedString(forKey:value:table:)` to load translations
- If a translation does not exist for a given key, the key will be returned instead
- **unless** a value is provided!

Good thing we parse the translation then. We can simply pass that along to our `tr` function that will send it further down the line.  This avoids the need of custom loading of bundles and strings files, and other fragile workarounds.

Note that this is a teeny-tiny breaking change for anyone using custom `lookupFunction`s. Probably not many people using this, and if there are any, it's easily fixed by adding a parameter to their custom function.


CHANGELOG.md
```diff
@@ -6,7 +6,9 @@
 
 ### Breaking Changes
 
-_None_
+* Strings: due to the bugfix for fallback translations, custom `lookupFunction`s will need to accept 3 arguments (table, key and value), up from 2 arguments (table and key).
```
  <tbody>
    <tr>
      <td>:warning:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/SwiftGen/SwiftGen/blob/230aa4f3c16364cb2a8fdb33569c9b0810f49805/CHANGELOG.md#L9"></span>Line describing your entry should end with a period and 2 spaces.</td>
    </tr>
  </tbody>
</table>

SwiftGen-Eve:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

---

</p>

PR #967, opened 2022-07-25
Fonts: fix issue with access in sandbox

### LINKED ISSUE #952: When running swiftgen from a command plugin the fonts are not detected because of missing Universal Type Identifier
I looked into the code and the problem seams to be the checks in the `FontParser`

The warning I get is

```
swiftgen: warning: Unable to determine the Universal Type Identifier for file lib/ios/Fonts/Roboto-Medium.ttf
```

Which is declared in the config file as

```
## FONTS

fonts:
  inputs: 
    - lib/ios/Fonts
  outputs:
    templateName: swift5
    params: 
      publicAccess: true
    output: lib/ios/Fonts.swift
```

When running from command line this works. But I use a command plugin to loop over the targets. I'm waiting for #926 to actually use it in a buildtool plugin. But for now I just commented out the code in the fontparser checking for the mentioned identifier, and this works.

But this seams nog ok to just remove the checks. But do the checks have value if you parse the file after and check for the correct extension?

```swift

public enum Fonts {
  public final class Parser: SwiftGenKit.Parser {
    var entries: [String: Set<Font>] = [:]
    private let options: ParserOptionValues

    public init(options: [String: Any] = [:], warningHandler: Parser.MessageHandler? = nil) throws {
      self.options = try ParserOptionValues(options: options, available: Parser.allOptions)
    }

    public static let defaultFilter = filterRegex(forExtensions: ["otf", "ttc", "ttf"])

    public func parse(path: Path, relativeTo parent: Path) throws {
        // TODO: unclear why when runnin in a plugin this check fails, also not clear why check is needed
//      guard let values = try? path.url.resourceValues(forKeys: [.typeIdentifierKey]),
//        let uti = values.typeIdentifier else {
//        warningHandler?("Unable to determine the Universal Type Identifier for file \(path)", #file, #line)
//        return
//      }
//      guard UTTypeConformsTo(uti as CFString, kUTTypeFont) else {
//        warningHandler?("File is not a known font type: \(path)", #file, #line)
//        return
//      }

      let fonts = CTFont.parse(file: path, relativeTo: parent)
      fonts.forEach { addFont($0) }
    }

    private func addFont(_ font: Font) {
      let familyName = font.familyName
      entries[familyName, default: []].insert(font)
    }
  }
}
```

That is what I did in my fork for now. Is there a better solution?

djbe:
Try to fix #952.

Getting resource keys of a file can fail in sandbox environments. Until this is fixed somehow, we're better of skipping this check (if the first step fails).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,10 @@ _None_
   [David Jennes](https://github.com/djbe)
   [#727](https://github.com/SwiftGen/SwiftGen/issues/727)
   [#965](https://github.com/SwiftGen/SwiftGen/pull/965)
+* Fonts: fix file-type check in sandboxed environments.  
+  [David Jennes](https://github.com/djbe)
+  [#952](https://github.com/SwiftGen/SwiftGen/issues/952)
+  [#967](https://github.com/SwiftGen/SwiftGen/pull/967)
 
 ### Internal Changes
 
--- a/Sources/SwiftGenKit/Parsers/Fonts/FontsParser.swift
+++ b/Sources/SwiftGenKit/Parsers/Fonts/FontsParser.swift
@@ -22,12 +22,7 @@ public enum Fonts {
     public static let defaultFilter = filterRegex(forExtensions: ["otf", "ttc", "ttf"])
 
     public func parse(path: Path, relativeTo parent: Path) throws {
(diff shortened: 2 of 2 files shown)


SwiftGen-Eve:

---

</p>

PR #988, opened 2022-08-01
Strings: escape translation strings

### LINKED ISSUE #985: Unprintable ASCII character found in Swift Source Code
Environment:
- SwiftGen 6.6.1
- Xcode 13.4
- Targeting iOS 15+

With the latest Swiftgen version, `\t` contained inside a localised string are incorrectly converted into a non ASCII character.

For instance:
```strings
"a_string_with_a_tab" = "A string with a tab\t";
```
Will result in the `strings.generated.swift` file as so:
```swift
internal static let aStringWithATab = L10n.tr("Localizable", "a_string_with_a_tab", fallback: "A string with a tab	")
```
Instead of:
```swift
internal static let aStringWithATab = L10n.tr("Localizable", "a_string_with_a_tab", fallback: "A string with a tab\t")
```

Here is the result with Xcode 13.4:
Screenshot 2022-08-01 at 11 02 33

### LINKED ISSUE #986: Build fails when using #" char sequence within the string value.
**Environment**: 

- SwiftGen 6.6.1
- Xcode 13.4.1
- Targeting iOS 15+

We use HTML tags in our string values so that the app can parse them and convert them to actual text styles. From the SwiftGen version 6.6.0 on, we're facing an issue with the new `fallback` generated parameter in the generated string file when using the HTML tag `font color="#12345"`. For instance, for the string `"text_with_color" = "<font color="#333333">Styled text</font>";` the generated string looks like this:

`internal static let failed = Strings.tr("Localization", "text_with_color", fallback: #"<font color="#333333">Styled text</font>"#)`

**Problem:** 

`fallback` parameter uses Swift's raw strings with the delimiter `#` which is colliding with the `"#` from the string, causing the project to fail at compile time.

**Possible solution**

Add more hash symbols as delimiter. The generated string below would compile perfectly. 

`internal static let failed = Strings.tr("Localization", "text_with_color", fallback: ##"<font color="#333333">Styled text</font>"##)`

Thank you.

djbe:
- If a string contains the sequence `"#`, it'll break our use of raw strings
- If a translation contains a tab `\t`, Xcode will complain that the source code contains invalid characters

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,11 @@ _None_
 
 ### Internal Changes
 
-_None_
+* Strings: correctly handle strings containing `\t` (tabs) and `"#`.  
+  [David Jennes](https://github.com/djbe)
+  [#985](https://github.com/SwiftGen/SwiftGen/issues/985)
+  [#986](https://github.com/SwiftGen/SwiftGen/issues/986)
+  [#988](https://github.com/SwiftGen/SwiftGen/pull/988)
 
 ## 6.6.1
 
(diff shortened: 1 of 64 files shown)


SwiftGen-Eve:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

---

</p>

PR #994, opened 2022-08-08
Bumped Yams to 5.0.1 #trivial

### LINKED ISSUE #992: Internal: correctly set minimum `swift-argument-parser` version
I'm seeing a broken build with the 6.6.1 release. This is in an SPM project with constraints on `swift-argument-parser` <= 1.10 (latest swiftlint release).

It looks like swiftgen's [using](https://github.com/SwiftGen/SwiftGen/blob/1db1ca768cdda4a2a1aef785f55167a1812fa9a0/Sources/SwiftGen/Commands/Run.swift#L61) ArgumentParser's `visibility` features, which got introduced in ArgumentParser's 1.1.0 release: https://github.com/apple/swift-argument-parser/releases/tag/1.1.0

Bumping the dependency up would highlight the issue during package resolving vs. at build time.

bdrelling:
Just wondering if we could bump Yams to `5.0.1`? If not, could we update the range to `"4.0.6"..<"6.0.0"`?

The only change in Yams 5.x is that they raised the minimum Swift version.

### DIFF
--- a/.github/workflows/archive.yml
+++ b/.github/workflows/archive.yml
@@ -18,14 +18,7 @@ jobs:
         uses: ruby/setup-ruby@v1
         with:
           bundler-cache: true
-      - 
-        name: Cache SPM
-        uses: actions/cache@v3
-        with:
-          path: .build
-          key: ${{ runner.os }}-spm-${{ hashFiles('**/Package.resolved') }}
-          restore-keys: |
-            ${{ runner.os }}-spm-
+      # Note: there's no SPM cache step here, to ensure "clean" compilation works
       -
         name: Verify installation
         run: bundle exec rake cli:install[./standalone,./standalone/frameworks,./standalone/stencils]
--- a/Package.resolved
+++ b/Package.resolved
@@ -77,8 +77,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/SwiftGen/StencilSwiftKit.git",
       "state" : {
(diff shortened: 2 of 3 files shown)


djbe:
For some reason I remember this creating a conflict with Komondor, which indirectly depends on Yams 4.x

bdrelling:
@djbe got it, thanks for the explanation. Could you or someone else verify? Also, if that is the case, then why would `"4.0.0"..<"6.0.0"` not work? In testing locally, it resolved as expected when I had a hard requirement on a `4.x` version, though I'd love an extra set of eyes in case something was cached locally or whatever.

Personally, my conflict is coming from another package that depends on Yams 5.x ([periphery](https://github.com/peripheryapp/periphery)), so I can't create a multi-plugin package using both unless I restrict the version of this package. I'm just doing this in the meantime, but thought I'd see if this package could be bumped.

If the range snippet I added above doesn't work, I'm curious how SPM would recommend a downstream consumer of multiple packages with these conflicts would go about handling it, unless the answer is just "explicit version locking"?

And FWIW we should be safe to make the same bump in Komondor. I'm happy to open up a PR over in that project as well, but wouldn't mind some verification here first that I'm not crazy.

Thank you!

djbe:
Why… is it failing to find Stencil now? It works totally fine locally? 😕

bdrelling:
🤷🏻‍♂️

djbe:
Weird thing is that `test:spm` is working, so SPM & full compilation works 😕 It's only compiling for release that seems broken.

PR #998, opened 2022-08-10
Strings: fix escape sequences for translations (swift 5)

### LINKED ISSUE #986: Build fails when using #" char sequence within the string value.

djbe:
Further fix for #986.

Problem with swift raw strings, is that every escape sequence needs to be expanded with the raw delimiter. So in a  `#"…"#` raw string, `\n` must become `\#n`.

With our delimiters already set to `###`, and an unknown number of escape sequences, we're better of removing raw strings and using normal ones again :shrug:.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,11 +18,12 @@ _None_
 
 ### Internal Changes
 
-* Strings: correctly handle strings containing `\t` (tabs) and `"#`.  
+* Strings: correctly handle translations containing `\t` (tabs) and other escape sequences.  
   [David Jennes](https://github.com/djbe)
   [#985](https://github.com/SwiftGen/SwiftGen/issues/985)
   [#986](https://github.com/SwiftGen/SwiftGen/issues/986)
   [#988](https://github.com/SwiftGen/SwiftGen/pull/988)
+  [#998](https://github.com/SwiftGen/SwiftGen/pull/998)
 * Strings: greatly improve the performance of the new comments parser.  
   [David Jennes](https://github.com/djbe)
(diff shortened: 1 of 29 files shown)


SwiftGen-Eve:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

---

</p>

djbe:
@teejayhh @narlei mind giving this a try? To make sure it works for whatever extra cases you guys have.

narlei:
> @teejayhh @narlei mind giving this a try? To make sure it works for whatever extra cases you guys have.

It worked perfectly now!

Issue #1007, opened 2022-08-25
Run command does not support `--quiet` or `-verbose`


jszumski:
Using `swiftgen run` without a config file does not allow you to adjust the log level.  Related to https://github.com/SwiftGen/SwiftGen/issues/823.

PR #1008, opened 2022-08-25
Added support for `--quiet` and `--verbose` flags to `swiftgen run`

jszumski:
Fixes https://github.com/SwiftGen/SwiftGen/issues/1007 by extending https://github.com/SwiftGen/SwiftGen/pull/846 to support using `--quiet` and `--verbose` with the `run` command.  If both flags are provided, quiet will override verbose.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,10 @@ _None_
 
 ### New Features
 
-_None_
+* CLI: Added support for `--quiet` and `--verbose` flags to `swiftgen run`.  
+  [John Szumski](https://github.com/jszumski)
+  [#1007](https://github.com/SwiftGen/SwiftGen/issues/1007)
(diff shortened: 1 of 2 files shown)


luispadron (approved):
Would love to see this merged, requiring a config for this setting is a bit odd.

djbe (changes requested):
Look at the `config run` command, there's already a `logLevel` flag that does this. There's no need for 2 separate flags. Reason for this is that these flags are **exclusive** (can't have both).

jszumski:
> Look at the `config run` command, there's already a `logLevel` flag that does this. There's no need for 2 separate flags. Reason for this is that these flags are **exclusive** (can't have both).

Updated, I didn't realize `CommandLogLevel` would translate directly into multiple flags 💯

djbe (approved):

PR #1038, opened 2023-01-30
Made L10n.tr not use bare strings as formats.

grigorye:

---

This addresses #1039, by introducing a specialized variant of `tr` for argument-less calls used with wrappers generated for bare strings. Beware that this is a breaking change for the cases where `%%` had to be used before in bare strings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@
 
 ### Breaking Changes
 
-_None_
+* Strings: due to the bugfix for bare strings, for structured-swift5, `%%` in bare strings will be rendered as-such, instead of single `%`.  
+  [Grigory Entin](https://github.com/grigorye)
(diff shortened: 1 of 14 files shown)


djbe:
I should add this to some sort of frequently asked questions. Do a search through the issues/PRs, and you'll find multiple discussions about this. 

The big problem with not formatting strings for those with no parameters, is that it creates a double standard for encoding percentages:
- For strings with parameters, if you to "print" the string `100%`, you'll need to provide the string `100%%`.
- For strings without parameters, you'll need to provided the string `100%`.

We do not want our users to have to deal with these inconsistencies. There's unfortunately no good solution for this 🤷 What we usually recommend is providing your own `lookupFunction` (see [the docs](https://github.com/SwiftGen/SwiftGen/blob/stable/Documentation/templates/strings/structured-swift5.md)), and do a check for no/some parameters. Or a custom template.

PR #1051, opened 2023-03-09
Support complicationset folders

Patrick-Kladek:

---


Patrick-Kladek:
I'm not sure if this is the right approach but it works for my use case and passes tests. If there is a better solution please point me in the right direction.

Sources/TestUtils/Fixtures/Resources/XCAssets/Other.xcassets/Complication.complicationset/Modular.imageset/Modular Small Simple 38mm.png
djbe, line 1: Do we need to commit all these images? For our other image assets, we simply added 1 variation, which is enough.

Sources/SwiftGenKit/Parsers/AssetsCatalog/Catalog.swift
```diff
@@ -26,7 +26,10 @@ extension AssetsCatalog.Catalog {
   /// This method recursively parses a directory structure, processing each folder (files are ignored).
   ///
   static func process(folder: Path, withPrefix prefix: String = "") -> [AssetsCatalogEntry] {
-    (try? folder.children().sorted(by: <).compactMap { path in
+    let children = try? folder.children()
+    let sorted = children?.sorted(by: <)
+
+    return (sorted?.compactMap { path in
```
djbe, line 32: This doesn't change anything?

djbe (changes requested):
I'm not very familiar with WatchOS complications, but from some quick googling:
- In your project (info.plist), you point it at the right "group" in your asset catalog
- I couldn't find any code samples where users load a complication asset programatically?

So I'm not sure if there's any point in adding support for this type of asset? 

Note: also found that all docs mark complications as deprecated, and everything now should be done with WidgetKit. Does that framework also use complicationset assets?

Issue #1079, opened 2023-09-27
Warnings about Universal Type Identifier which can't be defined when using plugin


kenji21:
As the warnings are still present, I dig to determine where they are coming from:

https://developer.apple.com/documentation/security/app_sandbox/discovering_and_diagnosing_app_sandbox_violations

To get the detail error:

```
swiftgen: warning: Unable to determine the Universal Type Identifier for file /Users/kenji/.....path-to-Regular.ttf : Error Domain=NSCocoaErrorDomain Code=256 "The file “Regular.ttf” couldn’t be opened." UserInfo={NSURL=file:///Users/kenji/.....path-to-Regular.ttf, NSFilePath=/Users/kenji/.....path-to-Regular.ttf, NSUnderlyingError=0x12d631cd0 {Error Domain=NSOSStatusErrorDomain Code=-10813 "kLSDataUnavailableErr: E.g. no kind string"}}
```

using /usr/bin/sandbox-exec with all the options, only replacing plugin command to: `/Users/kenji/github/SwiftGen/.build/debug/swiftgen` 

and after patching (and `swift build`):
![Screenshot 10](https://github.com/SwiftGen/SwiftGen/assets/1105089/8276f40e-e40b-4ed0-b3c3-ebf0a5f24543)

_Originally posted by @kenji21 in https://github.com/SwiftGen/SwiftGen/issues/952#issuecomment-1736771314_

PR #1080, opened 2023-09-27
Fonts: check file signature before emitting warning in sandboxed env (plugin)

kenji21:
Fix for https://github.com/SwiftGen/SwiftGen/issues/1079 and https://github.com/SwiftGen/SwiftGen/issues/952

To prevent warning being displayed in Xcode when using the plugin, I've added a check for the signature of font files (the first bytes) before emitting the warning

Screenshot 5

### DIFF
--- a/Sources/SwiftGenKit/Parsers/Fonts/FontsParser.swift
+++ b/Sources/SwiftGenKit/Parsers/Fonts/FontsParser.swift
@@ -35,18 +35,44 @@ public enum Fonts {
 
 // MARK: - Helpers
 
+private extension Data {
+    var isTrueTypeFont: Bool {
+        let trueTypeFontHeaderBytes: [UInt8] = [0x00, 0x01, 0x00, 0x00, 0x00] // https://en.wikipedia.org/wiki/List_of_file_signatures
+        var values = [UInt8](repeating: 0, count: trueTypeFontHeaderBytes.count)
+        self.copyBytes(to: &values, count: trueTypeFontHeaderBytes.count)
+        return values == trueTypeFontHeaderBytes
+    }
+
+    var isOpenTypeFont: Bool {
+        let openTypeFontHeaderBytes: [UInt8] = [0x4F, 0x54, 0x54, 0x4F] // https://en.wikipedia.org/wiki/List_of_file_signatures
+        var values = [UInt8](repeating: 0, count: openTypeFontHeaderBytes.count)
+        self.copyBytes(to: &values, count: openTypeFontHeaderBytes.count)
+        return values == openTypeFontHeaderBytes
+    }
+}
+
 private extension Fonts.Parser {
   /// Try to quickly check if the given file is an actual font file.
   ///
   /// - Note: This can seemingly fail in sandboxed environments.
   func checkIsFont(path: Path) -> Bool {
-    guard let values = try? path.url.resourceValues(forKeys: [.typeIdentifierKey]),
-      let uti = values.typeIdentifier else {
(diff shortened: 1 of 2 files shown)


alexandrevienne (approved):

mrackwitz (approved):
Seeing the same issue, I was happy to see this being addressed by a PR.
This looks good to me.

Romaric5M (approved):

RemiRodriguesEH (approved):

Sources/SwiftGenKit/Parsers/Fonts/FontsParser.swift
```diff
@@ -35,18 +35,41 @@ public enum Fonts {
...
   ///
   /// - Note: This can seemingly fail in sandboxed environments.
   func checkIsFont(path: Path) -> Bool {
-    guard let values = try? path.url.resourceValues(forKeys: [.typeIdentifierKey]),
-      let uti = values.typeIdentifier else {
-      warningHandler?("Unable to determine the Universal Type Identifier for file \(path)", #file, #line)
+    do {
+      let values = try path.url.resourceValues(forKeys: [.typeIdentifierKey])
+      guard let uti = values.typeIdentifier else {
+        return true
```
djbe, line 62: This is incorrect. If you can't get the `typeIdentifier`, it should fall back to reading the header bytes like you've done below.

I'd rework this `do-catch` into a simple `if-else`.

Sources/SwiftGenKit/Parsers/Fonts/FontsParser.swift
```diff
@@ -35,18 +35,41 @@ public enum Fonts {
...
+      }
+      return UTTypeConformsTo(uti as CFString, kUTTypeFont)
+    } catch {
+      if let data = try? Data(contentsOf: path.url),
```
djbe: This will load the whole file into memory, which is (or can be) slow. Instead, we should use [FileHandle.read(upToCount:)](https://developer.apple.com/documentation/foundation/filehandle/3516317-read), with a maximum of N header bytes.

Note: best to create a constant for the amount of header bytes, just to make it clear.

Sources/SwiftGenKit/Parsers/Fonts/FontsParser.swift
```diff
@@ -35,18 +35,41 @@ public enum Fonts {
 
 // MARK: - Helpers
 
+private extension Data {
+    var isTrueTypeFont: Bool {
+        var values = [UInt8](repeating: 0, count:5)
+        self.copyBytes(to: &values, count: 5)
+        let firstBytesOfTTF: [UInt8] = [0x00, 0x01, 0x00, 0x00, 0x00] // https://en.wikipedia.org/wiki/List_of_file_signatures
```
djbe: These could be constants, like `trueTypeFontHeaderBytes` and such. And then a simple function on `Path` to check a file type given some header bytes.

Also avoids having a fixed "count" in the line above, as you can get the `.count` from the constant.

djbe (changes requested):
Nice idea, I like it! Some small changes to improve the code & perf.

Sources/SwiftGenKit/Parsers/Fonts/FontsParser.swift
kenji21: > If there’s no available value for a given key in keys, the returned URLResourceValues contains nil for the corresponding item. If this method fails to determine a value’s availability or retrieve its value, the method throws an error.

As [resourceValues(forKeys:)](https://developer.apple.com/documentation/foundation/url/resourcevalues(forkeys:)) throws... I can't rework for a simple if-else

Issue #1099, opened 2024-01-26
Support for visionOS


rcolon-shockoe:
Latest stable version 6.6.2 doesn't support visionOS for generating the image Assets specifically as it doesn't currently have compiler flags for `os(visionOS)`. Are there any updates coming soon to support visionOS either for this repo or for a Stencil script? 
<img width="775" alt="Screenshot 2024-01-26 at 2 50 30 PM" src="https://github.com/SwiftGen/SwiftGen/assets/113468129/1f331ecf-fd9b-4234-80a4-da2c218a9f48">

Thanks!

Liquidsoul:
While we see to work on this stencil issue, note that you can use custom ones. Use the existing ones as a starting point.

lucaArchidiacono:
Might be already solved based on @Liquidsoul 's comment.

But just want to leave a draft, for those who want to quickly have a copy-paste solution:
```
// swiftlint:disable all
// Generated using SwiftGen — https://github.com/SwiftGen/SwiftGen

{% if catalogs %}
{% macro hasValuesBlock assets filter %}
  {%- for asset in assets -%}
    {%- if asset.type == filter -%}
      1
    {%- elif asset.items -%}
      {% call hasValuesBlock asset.items filter %}
    {%- endif -%}
  {%- endfor -%}
{% endmacro %}
{% set enumName %}{{param.enumName|default:"Asset"}}{% endset %}
{% set arResourceGroupType %}{{param.arResourceGroupTypeName|default:"ARResourceGroupAsset"}}{% endset %}
{% set colorType %}{{param.colorTypeName|default:"ColorAsset"}}{% endset %}
{% set dataType %}{{param.dataTypeName|default:"DataAsset"}}{% endset %}
{% set imageType %}{{param.imageTypeName|default:"ImageAsset"}}{% endset %}
{% set symbolType %}{{param.symbolTypeName|default:"SymbolAsset"}}{% endset %}
{% set forceNamespaces %}{{param.forceProvidesNamespaces|default:"false"}}{% endset %}
{% set accessModifier %}{% if param.publicAccess %}public{% else %}internal{% endif %}{% endset %}
{% set hasARResourceGroup %}{% for catalog in catalogs %}{% call hasValuesBlock catalog.assets "arresourcegroup" %}{% endfor %}{% endset %}
{% set hasColor %}{% for catalog in catalogs %}{% call hasValuesBlock catalog.assets "color" %}{% endfor %}{% endset %}
[...]

PR #1101, opened 2024-02-09
Added support for visionOS platform

shadone:
Fixes https://github.com/SwiftGen/SwiftGen/issues/1099

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ _None_
   [John Szumski](https://github.com/jszumski)
   [#1007](https://github.com/SwiftGen/SwiftGen/issues/1007)
   [#1008](https://github.com/SwiftGen/SwiftGen/pull/1008)
+* CLI: Added support for visionOS platform.  
+  [Denis Dzyubenko](https://github.com/shadone)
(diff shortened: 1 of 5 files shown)


Liquidsoul:
Hi @shadone and thank you for your contribution.

Could you please update the fixtures related to your changes in the stencils? This can be done by running the command:
`bundle exec rake generate:output`.

You'll also need to rebase and update the location of your changelog entry.

We may also need to add the new platform to [this file](https://github.com/SwiftGen/SwiftGen/blob/stable/rakelib/compilation_configuration.rb#L26-L31), but I do not if this is a good idea because I tried to use `arm64-apple-visionos1.0` and it seems that you need to install the visionOS SDK as opposed to other SDKs like tvOS and watchOS.  
What I mean is that I do not know if the SDK is installed in the CI and this addition would require anyone who want to run the tests locally to install the visionOS SDK 🤔

AF-cgi:
Any updates?

djbe (changes requested):
As mentioned by @Liquidsoul, this PR needs to generate the updated output files, otherwise the tests will fail.

PR #1107, opened 2024-03-02
fix: rake output:compile for XCAssets templates by removing lazy vars

### LINKED ISSUE #1022: Stored properties cannot be marked potentially unavailable with '@available'
After bumping to version 6.6.2 with Cocoapods, [this piece of code](https://github.com/SwiftGen/SwiftGen/blob/1cf6d7eebd70c2157f69d5a991bc57c1ef182ed1/Sources/SwiftGenCLI/templates/xcassets/swift5.stencil#L215):

```swift
    #if canImport(SwiftUI)
        @available(iOS 13.0, tvOS 13.0, watchOS 6.0, macOS 10.15, *)
        internal private(set) lazy var swiftUIColor: SwiftUI.Color = .init(asset: self)
    #endif
```

brings the following compile error with Xcode 14 (my project supports iOS 12+): 

```
Stored properties cannot be marked potentially unavailable with '@available'
```

After applying the following edit on the generated file:

```diff
         }
     #endif
 
-    #if canImport(SwiftUI)
-        @available(iOS 13.0, tvOS 13.0, watchOS 6.0, macOS 10.15, *)
-        internal private(set) lazy var swiftUIColor: SwiftUI.Color = .init(asset: self)
-    #endif
-
     fileprivate init(name: String) {
         self.name = name
     }
 }
 
+#if canImport(SwiftUI)
+    @available(iOS 13.0, tvOS 13.0, watchOS 6.0, macOS 10.15, *)
+    extension ColorAsset {
+        var swiftUIColor: SwiftUI.Color {
+            .init(asset: self)
+        }
+    }
+#endif
+
 internal extension ColorAsset.Color {
     @available(iOS 11.0, tvOS 11.0, watchOS 4.0, macOS 10.13, *)
     convenience init?(asset: ColorAsset) {
```

I got my project to compile OK.

I would be happy to help with this issue if you give me any guidance, as suggested by the CONTRIBUTING file of the project.

Thanks.

Liquidsoul:

---

The Swift compiler does not allow the use of @available on stored variables anymore. This means that the `lazy var` used in the templates cannot be made available only for certain OSes.
I have fixed the issue by removing the `lazy var` altogether. Compared to the previous implementation, this introduces an overhead on every access to a color. I could not find a way to keep the lazy var without droping old OSes.

Fixes #1022

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,10 @@ _None_
   [#1030](https://github.com/SwiftGen/SwiftGen/issues/1030)
   [#1104](https://github.com/SwiftGen/SwiftGen/issues/1104)
   [#1105](https://github.com/SwiftGen/SwiftGen/pull/1105)
+* XCAssets: Fix rake output:compile for XCAsserts templates by removing lazy vars.  
+  [Liquidsoul](https://github.com/liquidsoul)
+  [#1022](https://github.com/SwiftGen/SwiftGen/issues/1022)
+  [#1107](https://github.com/SwiftGen/SwiftGen/pull/1107)
 
 ### Internal Changes
 
--- a/Sources/SwiftGenCLI/templates/xcassets/swift5.stencil
+++ b/Sources/SwiftGenCLI/templates/xcassets/swift5.stencil
@@ -193,12 +193,12 @@
   #endif
 
   @available(iOS 11.0, tvOS 11.0, watchOS 4.0, macOS 10.13, *)
-  {{accessModifier}} private(set) lazy var color: Color = {
+  {{accessModifier}} var color: Color {
     guard let color = Color(asset: self) else {
       fatalError("Unable to load color asset named \(name).")
     }
     return color
-  }()
+  }
 
   #if os(iOS) || os(tvOS)
   @available(iOS 11.0, tvOS 11.0, *)
@@ -213,9 +213,9 @@
 
   #if canImport(SwiftUI)
   @available(iOS 13.0, tvOS 13.0, watchOS 6.0, macOS 10.15, *)
-  {{accessModifier}} private(set) lazy var swiftUIColor: SwiftUI.Color = {
+  {{accessModifier}} var swiftUIColor: SwiftUI.Color {
     SwiftUI.Color(asset: self)
-  }()
+  }
   #endif
 
   fileprivate init(name: String) {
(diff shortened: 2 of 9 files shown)


Liquidsoul:

---

</p>

ZevEisenberg:
I think `let`s are lazy, right? Or they can be in certain circumstances? Can you make them `let`s?

ZevEisenberg:
I guess `let`s are only lazy in certain circumstances? I’m actually not sure.

Liquidsoul:
> I think `let`s are lazy, right? Or they can be in certain circumstances? Can you make them `let`s?

The point is that I cannot make them "stored" at all with and use the `@available` anymore.
This is something that has been [banned in Xcode 15](https://developer.apple.com/documentation/xcode-release-notes/xcode-15-release-notes).

Screenshot 2024-03-14 at 19 48 03

Even with `let` we have this error:
> error: stored properties cannot be marked potentially unavailable with '@available'

ZevEisenberg:
Ohhh right, I gotcha. I had even read something about that and then didn’t process it properly. I think making it a computed var is fine. All the asset initialization stuff is cached anyway, I believe, so it’s probably fine to defer to Apple’s caching.

djbe:
The `lazy var`'s were introduced because of some performance issues, because the system did NOT cache them, at least back then. The [initial report](https://github.com/SwiftGen/SwiftGen/issues/578) and then the [pull request](https://github.com/SwiftGen/SwiftGen/pull/589) discuss this.

But I'm fine with dropping this cache for now, and we can revisit this issue should it turn up again.

ZevEisenberg:
From that thread:

> It's weird that Apple didn't use some caching mechanism as they did with UIImage's.

💡 I was thinking of images. I didn't realize colors were different. Maybe worth re-testing to see if it's faster now.

Liquidsoul:
> The `lazy var`'s were introduced because of some performance issues, because the system did NOT cache them, at least back then. The [initial report](https://github.com/SwiftGen/SwiftGen/issues/578) and then the [pull request](https://github.com/SwiftGen/SwiftGen/pull/589) discuss this.
> 
> But I'm fine with dropping this cache for now, and we can revisit this issue should it turn up again.

@djbe  My idea when I did this was to:
1. make a short term fix to restore the CI jobs by removing the cache because I did not find a better solution.
2. see to drop the support for previous OSes in templates to restore the cache for at least the `color` property. The `swiftUIColor` would require to drop more versions or to rework the template code.

Would that be OK for you?

djbe:
Yeah dropping OS' is a bit of a difficult question, that I've struggled with before. Right now our templates are Swift "5", but there are many versions of that major version, with many improvements (such as property wrappers). Question is where do we draw the line regarding support, and should we bump SwiftGen a major version to do so. But that's a discussion that shouldn't be done in a closed PR thread 😅

PR #1119, opened 2024-04-29
Make Asset and ImageAsset conform to Sendable

SergeyPetrachkov:
## What's done:

* added Sendable conformance to ImageAsset and Asset types so the generated code is Swift6- and strict-concurrency-checks- ready. Since all types that are used in the equation are `Sendable`, and since `Sendable` is available iOS 8+, macOS 10.10+, watchOS 2.0+ (in other words from the very old versions), it should be safe to just add Sendable conformance to the swift5 stencil without introducing breaking changes. This will allow people to use the generated codes without any unwanted warnings from the compiler.

---

### DIFF
--- a/Sources/SwiftGenCLI/templates/xcassets/swift5.stencil
+++ b/Sources/SwiftGenCLI/templates/xcassets/swift5.stencil
@@ -129,7 +129,7 @@
   {% endfor %}
 {% endmacro %}
 // swiftlint:disable identifier_name line_length nesting type_body_length type_name
-{{accessModifier}} enum {{enumName}} {
+{{accessModifier}} enum {{enumName}}: Sendable {
   {% if catalogs.count > 1 or param.forceFileNameEnum %}
   {% for catalog in catalogs %}
   {{accessModifier}} enum {{catalog.name|swiftIdentifier:"pretty"|escapeReservedKeywords}} {
@@ -275,7 +275,7 @@
 {% endif %}
 {% if hasImage %}
 
-{{accessModifier}} struct {{imageType}} {
+{{accessModifier}} struct {{imageType}}: Sendable {
   {{accessModifier}} fileprivate(set) var name: String
 
   #if os(macOS)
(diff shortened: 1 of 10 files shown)


Sources/SwiftGenCLI/templates/xcassets/swift5.stencil
```diff
@@ -129,7 +129,7 @@
   {% endfor %}
 {% endmacro %}
 // swiftlint:disable identifier_name line_length nesting type_body_length type_name
-{{accessModifier}} enum {{enumName}} {
+{{accessModifier}} enum {{enumName}}: Sendable {
```
lgrzywac, line 132: I assume the template should be compatible with all Swift5 variants. Therefore I suggest the following:

```suggestion
#if swift(>=5.7)
{{accessModifier}} enum {{enumName}}: Sendable {
#else
{{accessModifier}} enum {{enumName}} {
#endif
```

If I'm wrong - just ignore the comment.
If I'm right - please apply the same to struct template below. 

The `Sendable` protocol has been added to Swift 5.7, so it's not available in Swift 5.6 and earlier: [SE-0302 Sendable and @Sendable closures](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0302-concurrent-value-and-concurrent-closures.md#sendable-protocol)

rakesthedon (approved):

oleksiikolomiietssnapp (approved):

emadhegab:
what holding this? we are suffering from that cause we have to add it manually everytime , we really need this ASAP

SergeyPetrachkov:
I don't have writing access to the repo. Someone from the owners should do a review, approve and merge it 🤔

emadhegab:
I think the owner left the project to no one it didn't got updated for 9 months or more. that's really bad. I would consider going to another things like R.swift or something to get this over with .

trimech (approved):

bactn:
Any update this PR?

jayeshkawli:
Thank you for authoring this. We're also facing the similar issue. We will appreciate if this could get merged that will unblock us from the Swift 6 transition work.

PR #634, opened 2019-04-27
Strings: Support plurals / parse .stringsdict files

ffittschen:
As discussed in #184, this PR adds support for parsing `.stringsdict` files and it adds support for (basic) plurals. Since there are some decisions that still need to be made, I marked the PR as WIP.

To be able to parse `.stringsdict` in addition to `.strings` files, I created subparsers similar to the subparsers used for Colors. The `.stringsdict` file is parsed by creating a `StringsDict` type conforming to `Decodable` and then parsing the file using the `PropertyListDecoder`. Since @djbe highlighted in #184 that the `.stringsdict` files can also contain non-plural dictionaries, which are used for variable width strings, I chose to parse those as well to be able to use them at a later point in SwiftGen. 

#### TODO:
- [x] Decide whether to provide only one plural translation in the doc comment of the generated file, or multiple/all
- [x] Fix the stencil so that `vertical_whitespace_opening_braces` doesn't trigger
- [x] Decide how to handle `Localizable.strings` and `Localizable.stringsdict` having the same (table) name.
- [x] Add tests for the `plurals-advanced` and `plurals-same-table` contexts

Closes #184.

PR #1128, opened 2024-06-19
Show translations instead of plural key for entries from stringdict files

PaulTaykalo:

---

Currently, SwiftGen, when parsing `.stringdict` for earch entry generates "translation" in format ```/// Plural format key: "%#@apples@"```, which is a bit confusing.  Since for other files i.e. `.strings`, real translation will be included.  

So this PR changes what will be returned in the `Strings.Entry`. Now, real translation will go to the translation, and additional optional parameter `formatKey` added for the cases, if formatKey needed for the templates (i.e. in comments)

### DIFF
--- a/Sources/SwiftGenKit/Parsers/Strings/FileTypeParser/StringsDictFileParser.swift
+++ b/Sources/SwiftGenKit/Parsers/Strings/FileTypeParser/StringsDictFileParser.swift
@@ -33,9 +33,12 @@ extension Strings {
 
         return try plurals.map { keyValuePair -> Entry in
           let (key, pluralEntry) = keyValuePair
+          // We are taking "Other" rule, since its the only one that always should be there
+          let translation = (pluralEntry.variables.first?.rule.other) ?? ""
           return Entry(
             key: key,
-            translation: "Plural format key: \"\(pluralEntry.formatKey)\"",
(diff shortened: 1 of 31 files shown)


Sources/SwiftGenKit/Parsers/Strings/FileTypeParser/StringsDictFileParser.swift
```diff
@@ -33,9 +33,12 @@ extension Strings {
 
         return try plurals.map { keyValuePair -> Entry in
           let (key, pluralEntry) = keyValuePair
+          // We are taking "Other" rule, since its the only one that always should be there
+          let translation = (pluralEntry.variables.first?.rule.other) ?? ""
           return Entry(
             key: key,
-            translation: "Plural format key: \"\(pluralEntry.formatKey)\"",
+            formatKey: pluralEntry.formatKey,
+            translation: translation,
```
PaulTaykalo, line 41: The actual change is in here

PaulTaykalo:
I'm currently reading the whole conversation why is plularFormat There, But still, thinkg that from the developer's perpsective - having `translation` (any translation) in the Entry is more correct than having "format" in the translation key.
https://github.com/SwiftGen/SwiftGen/pull/634

